PR #6, opened 2019-11-04
Improve SQL error handling

N-o-X:


connection/mysql.go
```diff
@@ -147,6 +172,20 @@ func (dbw *DBWrapper) checkConnection(isTicker bool) bool {
 	}
 }
 
+func (dbw *DBWrapper) isConnectionError(err error) bool {
+	for _, str := range connectionErrors {
+		if strings.Contains(err.Error(), str) {
```
Al2Klimov: What about calling `err.Error()` only once?

N-o-X: Good point. Fixed

PR #40, opened 2019-11-26
Schema: make *history#host_id NOT NULL

### LINKED ISSUE #32: History: host_id should be NOT NULL and always set
This allows us to directly have a reference to a services host in every history entry.

Al2Klimov:
fixes #32


Al2Klimov:
## Blocked by

* [ ] https://github.com/Icinga/icinga2/pull/7657

PR #60, opened 2019-11-29
Handle SIGTERM and SIGINT

### LINKED ISSUE #10: Signal handling & shutdown message

Al2Klimov:
fixes #10

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,10 +62,19 @@ import (
 	"github.com/Icinga/icingadb/prometheus"
 	"github.com/Icinga/icingadb/supervisor"
 	log "github.com/sirupsen/logrus"
+	"os"
+	"os/signal"
 	"sync"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -62,10 +62,19 @@ import (
...
 	"sync"
+	"syscall"
 )
 
 func main() {
+	{
+		ch := make(chan os.Signal, 1)
+		signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT)
+		go handleSignal(ch)
+	}
```
N-o-X: Does this really need to be its own scope? Couldn't you just move all of this into the handleSignal function?

Al2Klimov: 1. We don't need `ch` on the stack.
2. I prefer to capture as less as possible in functions/closures, on principle.

PR #61, opened 2019-11-29
Get rid of not used checksums

### LINKED ISSUE #30: Host/Service: Remove customvar_/group_checksum
Those columns are currently unused and we don't plan to use them in the future.

Al2Klimov:
fixes #30


dnsmichi:
There's likely a reason why they do exist in the schema in the first place. Are you 100% sure that web doesn't need them?

N-o-X:
@dnsmichi they were initially used by IcingaDB to check for changes in child objects, but we don't do that any more. Those child objects have their own checksums or are synced by ID.

Issue #46, opened 2019-11-27
Consider moving away from JSON


Al2Klimov:
... to a faster equivalent.

Al2Klimov:
Alternatives:

* gob (with C/Go static lib or custom implementation in Icinga 2)
* custom binary protocol (Hello ObjectPacker!)
* ...

dnsmichi:
Keep in mind that everything which is a custom implementation introduces bugs and possible breakdowns. I'm not a friend of changing this in Icinga 2 (again), where we move to tested and proven libraries instead. Personally I also don't like the raw Redis implementation in the RedisConnection class, I would never have merged that without OSMC pressure.

jm2c,
Michael

Al2Klimov:
We've implemented 🚀 **a whole custom DSL** 🚀 in v2.0, so I don't see any problems in the custom Redis protocol I/O or in a custom binary protocol (which we already use in our checksums).

If nobody'd make any custom implementations, you wouldn't have any of these:

* OS
* browser/httpd
* mail client/server
* monitoring tool
* ...

dnsmichi:
Since you already have learned how hard these things are to maintain, I can only strongly advise against following this route. Considering the planned release date with next week, this sounds like a total rewrite. Imho this shouldn't be part of the release candidate.

Al2Klimov:
Don't worry, if it won't get done by that date, we'll just not include it.

lippserd:
I think we had a little misunderstanding here because this is something for a future version.

Al2Klimov:
Damn. The alternatives I described won't work.

PR #66, opened 2019-11-29
Use github.com/intel-go/fastjson

Al2Klimov:


Al2Klimov:
May be an alternative to #46.

lippserd:
Do we have a performance improvement here?

Al2Klimov:
According to their docs and our code – yes.

lippserd:
Please provide a benchmark here.

N-o-X:
And please don't work on this before 1.0-rc1.

Al2Klimov:
According to this hijack:

```go
package hijack

import (
	"github.com/intel-go/fastjson"
	"github.com/prometheus/client_golang/prometheus"
	"time"
)

var uspb = prometheus.NewSummary(prometheus.SummaryOpts{Name: "uspb"})
var _ = func() struct{} {
	prometheus.Register(uspb)
	return struct{}{}
}()

func Unmarshal(data []byte, v interface{}) error {
	defer func(start time.Time) {
		uspb.Observe(float64(time.Now().Sub(start)) / float64(time.Microsecond) / float64(len(data)))
	}(time.Now())
	return fastjson.Unmarshal(data, v)
}
```

... fastjson doubles the speed.

lippserd:
`intel-go/fastjson` has been archived. I would not count on that anymore.

PR #77, opened 2019-12-02
XREAD from icinga:stats

Al2Klimov:


Al2Klimov:
The other side: https://github.com/Icinga/icinga2/pull/7678

ha/heartbeat_test.go
```diff
@@ -35,11 +37,17 @@ func TestIcingaHeartbeatListener(t *testing.T) {
...
 	var uj interface{} = nil
-	if err := json.Unmarshal([]byte(icingastate), &uj); err != nil {
-		assert.Nil(t, err)
-	}
-
-	rdb.Rdb.Publish("icinga:stats", icingastate)
+	assert.Nil(t, json.Unmarshal([]byte(app), &uj))
+	assert.Nil(t, json.Unmarshal([]byte(dump), &uj))
+
+	rdb.Rdb.XAdd(&redis.XAddArgs{
```
N-o-X: I think we should not use our own wrappers in this test. I know we did that before, but I don't think we should. Please remove our Publish and XAdd functions.

PR #93, opened 2019-12-11
Schema: Use DEFAULT NULL everywhere & fix table options

N-o-X:
We currently use both `DEFAULT NULL` and `NULL DEFAULT NULL` randomly

### DIFF
--- a/etc/schema/mysql/mysql.schema.sql
+++ b/etc/schema/mysql/mysql.schema.sql
@@ -104,7 +104,7 @@ CREATE TABLE hostgroup_customvar (
   customvar_id binary(20) NOT NULL COMMENT 'customvar.id',
 
   PRIMARY KEY (id)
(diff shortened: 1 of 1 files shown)


etc/schema/mysql/mysql.schema.sql
```diff
@@ -711,7 +711,7 @@ CREATE TABLE customvar (
   value text NOT NULL,
 
   PRIMARY KEY (id)
-) ENGINE=InnoDb ROW_FORMAT=COMPRESSED DEFAULT CHARSET=utf8mb4 COLLATE utf8mb4_bin;
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
```
Al2Klimov: Is this intended?

PR #101, opened 2019-12-12
Schema: s/DEFAULT CHARSET/DEFAULT CHARACTER SET/

Al2Klimov:


Al2Klimov:
https://dev.mysql.com/doc/refman/5.5/en/create-table.html

lippserd:
Let us use **DEFAULT CHARSET**. That's what `SHOW CREATE TABLE ...` is returning on every supported version.

PR #158, opened 2020-04-07
Update my name and email

Crunsher:
This removes my damned dead name (This is not optional) and updates my mail to one GitHub knows (for clout) since the project was uploaded after I left the project.


Al2Klimov:
If you now remove your old Icinga eMails from GitHub, will our contributor stats change?

Issue #147, opened 2020-04-01
IcngaDB fails with: Prepared statement contains too many placeholders - Overdue indicators not chunked


dmitriy-terzeman:
## Description

In high loaded environment i'm facing IcngaDB crash with the following errors:

```
Mar 31 08:56:44 p3i212t4-idb systemd: Started Icinga DB.
Mar 31 08:56:44 p3i212t4-idb icingadb: time="2020-03-31T08:56:44-07:00" level=info msg="Connecting to Redis"
Mar 31 08:56:44 p3i212t4-idb icingadb: time="2020-03-31T08:56:44-07:00" level=info msg="Connecting to MySQL"
Mar 31 08:56:44 p3i212t4-idb icingadb: time="2020-03-31T08:56:44-07:00" level=info msg="Redis connection established"
Mar 31 08:56:44 p3i212t4-idb icingadb: time="2020-03-31T08:56:44-07:00" level=info msg="Starting heartbeat listener"
Mar 31 08:56:44 p3i212t4-idb icingadb: time="2020-03-31T08:56:44-07:00" level=info msg="SQL connection established"
Mar 31 08:56:44 p3i212t4-idb icingadb: time="2020-03-31T08:56:44-07:00" level=info msg="Got initial environment." UUID=e727d600-79fe-4cfb-b636-83e5a7ccd7a8 context=HA environment=da39a3ee5e6b4b0d3255bfef95601890afd80709
Mar 31 08:56:44 p3i212t4-idb icingadb: time="2020-03-31T08:56:44-07:00" level=info msg="Taking over." UUID=e727d600-79fe-4cfb-b636-83e5a7ccd7a8 context=HA environment=da39a3ee5e6b4b0d3255bfef95601890afd80709
Mar 31 08:56:46 p3i212t4-idb icingadb: time="2020-03-31T08:56:46-07:00" level=fatal msg="Error 1390: Prepared statement contains too many placeholders"
Mar 31 08:56:46 p3i212t4-idb systemd: icingadb.service: main process exited, code=exited, status=1/FAILURE
Mar 31 08:56:46 p3i212t4-idb systemd: Unit icingadb.service entered failed state.
Mar 31 08:56:46 p3i212t4-idb systemd: icingadb.service failed.
```

## How to reproduce:
Currently i have lack of info for deeper diagnostics and not able to reproduce it, error occurs randomly on high loaded environment with 100k services per minute.

## Environment:

[...]

PR #165, opened 2020-04-23
MySQL: Make sure to chunk rows correctly

N-o-X:
This fixes a "too many placeholders" exception caused by large nested custom vars.
Custom vars get flattened by calling `getFinalRows()` on custom var rows.
https://github.com/Icinga/icingadb/blob/49891dfa1b0560dcd038241454f27f13c3a0e8d4/connection/mysql.go#L695

This extends the chunk size of 500 rows, provided by our Redis implementation, by an unlimited amount.

Test config:

```
numHosts = 500
for (hid in range(numHosts)) {
        object Host "TestHost - " + hid use (hid) {
                check_command = "dummy"
                vars.perp = {}
                var numPerpVars = 500
                for (i in range(numPerpVars)) {
                        vars.perp["var" + i] = i + "_" + hid
                }
        }
}
```

This PR might also be the solution for #147, but that needs testing.

### DIFF
--- a/connection/mysql.go
+++ b/connection/mysql.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Icinga/icingadb/utils"
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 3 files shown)


connection/mysql_utils.go
```diff
@@ -432,3 +432,24 @@ type Row interface {
...
+			select {
+			case ch <- rows[i:end]:
+			case <-done:
+				return
+			}
+		}
+	}()
+
+	return ch
+}
```
Al2Klimov, line 455: How many rows you expect so you don't compute and return an `[][]Row` sync-ly?

connection/mysql.go
```diff
@@ -702,26 +702,51 @@ func (dbw *DBWrapper) SqlBulkInsert(rows []Row, stmt *BulkInsertStmt) error {
...
+	wgDone := make(chan struct{})
 
-		for _, v := range r.InsertValues() {
-			values[j] = v
-			j++
-		}
-	}
+	for b := range ChunkRows(done, rows, 500) {
+		wg.Add(1)
+		go func(bulk []Row) {
```
Al2Klimov, line 714: Does this parallelism even have an effect? Can you insert >1x into the same table concurrently? Doesn't MySQL lock them?

connection/mysql.go
```diff
@@ -702,26 +702,51 @@ func (dbw *DBWrapper) SqlBulkInsert(rows []Row, stmt *BulkInsertStmt) error {
...
 
-	if err != nil {
+			query := fmt.Sprintf(stmt.Format, strings.Join(placeholders, ", "))
+
+			_, err := dbw.WithRetry(func() (result sql.Result, e error) {
+				return dbw.SqlExec(mysqlObservers.bulkInsert, query, values...)
+			})
+
+			if err != nil {
+				chErr <- err
```
Al2Klimov, line 736: If both of two operations fail, one writes to chErr successfully as you read from it in the select below. The other one will deadlock.

connection/mysql.go
Al2Klimov: CC @lippserd

lippserd: MySQL with InnoDB supports parallel data inserts. It is best used with different tables of course but also works with the very same table. But don't nail me down on how it's done 😆 It could be the case that we face deadlocks here and then. If so, the inserts should be re-executed.

Al2Klimov: OK, then we should keep this behavior at least "just to be sure" not to loose performance.

connection/mysql_utils.go
N-o-X: This can get really large, depending on the size of your custom vars. 
I'm not sure, if this helps a lot, but I also don't think it'll hurt us. What do you think?

connection/mysql.go
N-o-X: Whoops, that's bad... Shouldn't have overlooked that. Thanks!

connection/mysql_utils.go
Al2Klimov: **How** large in the worst case?

N-o-X: Basically infinite, because we don't have limits on how large a custom var can be, right?

Al2Klimov: The point is that you already have a slice `rows[]Row`. The `[][]Row` I suggest would be about `len(rows)/500` long and its `[]Row` items would re-use `rows[i][j]`'s memory.

PR #215, opened 2020-10-07
Add missing indices to *_customvar tables

nilmerg:
Adds the same indices just like they're already part of `host_customvar` and `service_customvar` to all other `*_customvar` tables

### DIFF
--- a/etc/schema/mysql/mysql.schema.sql
+++ b/etc/schema/mysql/mysql.schema.sql
@@ -112,7 +112,10 @@ CREATE TABLE hostgroup_customvar (
   hostgroup_id binary(20) NOT NULL COMMENT 'hostgroup.id',
   customvar_id binary(20) NOT NULL COMMENT 'customvar.id',
 
-  PRIMARY KEY (id)
+  PRIMARY KEY (id),
+
+  INDEX idx_hostgroup_customvar_hostgroup_id (hostgroup_id, customvar_id),
+  INDEX idx_hostgroup_customvar_customvar_id (customvar_id, hostgroup_id)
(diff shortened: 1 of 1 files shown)


Al2Klimov (commented):
Yes, other construction areas (eventually other PRs), but still interesting:

* Why they're not unique?
* Why they're duplicated, i.e. a,b and b,a?

nilmerg:
I just copied the same indices from `service_customar`. No idea why those are not unique. The duplication is due to the join direction I suppose. Though, @lippserd may provide details.

Al2Klimov:
Nothing critical, as I said:

> other construction areas

julianbrost:
Ok, then this is nothing this PR has to block on. If you think this is important enough, please open a separate issue for that.

Al2Klimov (changes requested):
Btw.: rebase the branch.

julianbrost (approved):

Al2Klimov:
Before I approve this, I'd still like to know the point behind the B,A indices in addition to A,B ones.

julianbrost:
One for showing variables by object, the other one for finding all objects with some variable set?

Al2Klimov:
Sure, but is this actually necessary?

julianbrost:
Web supports showing variables for objects and filtering objects by variables, so I have reason to believe so.

lippserd:
> Web supports showing variables for objects and filtering objects by variables, so I have reason to believe so.

Correct. In addition, the MySQL optimizer determines the leading table depending on the queries and table sizes. Which could be one or the other. But there is no rule of thumb here.

lippserd (approved):

PR #219, opened 2020-10-16
mysql/schema: Add name indices to host,service,user,usergroup tables

nilmerg:
`name_ci` is the column used by Icinga DB Web to perform quick searches, so indexing it is required.

Also adds indices for `name` and `display_name` to tables `user` and `usergroup`. Just like it is already the case for tables `host` and `service`.


Al2Klimov (approved):
Why not.

N-o-X (approved):

PR #221, opened 2020-10-21
Support PostgreSQL

### LINKED ISSUE #136: PostgreSQL support
In our production environments, PostgreSQL is much preferred over MySQL/MariaDB due primarily to its much more robust replication support. It would be nice to be able to use PostgreSQL instead of MariaDB for Icinga DB.

Al2Klimov:
fixes #136


Al2Klimov:
@lippserd Please have a look at 9c3982d.

Al2Klimov:
@N-o-X Nice [list](https://github.com/Icinga/icingadb/commit/417990c9a7d6f5979121e22583b6c036106ad447). How did you collect the items?

Al2Klimov:
Stumbled over https://github.com/lib/pq/issues/1006.

@Thomas-Gelf Have you ever seen any customers insisting on TLS between app and DB?

Thomas-Gelf:
> @Thomas-Gelf Have you ever seen any customers insisting on TLS between app and DB?

There are some with such policies in place. Usually the wording is not "the DB connection must use TLS", it's more like "the connection to databases or similar services must be encrypted". Policies vary depending on the kind of data being stored. As Custom Vars in the IcingaDB might carry secrets, we usually fall in the "sensible" class.

While TLS is fine for daemons, it is annoying for the web because of the additional latency - "per click". Workarounds include local DB proxies then talking encrypted to the DB or IPSEC as a transport. Often you can also compile a form and check the "not yet, but soon" option :laughing:

Long story short: yes. And they're probably getting more.

Al2Klimov:
Thanks, good to know. Cool that we don't _have to_ support TLS _directly_.

Thomas-Gelf:
> Thanks, good to know. Cool that we don't _have to_ support TLS _directly_.

IMHO that shouldn't be the conclusion :D In every project where we didn't offer TLS for DB connections we got issues complaining about that. Therefore I'd strongly suggest to offer it, especially for new projects. Eventually we should not only offer it everywhere, but lead by example - and strongly recommend it for every setup and configuration.

Al2Klimov:
As the tests show, it works... theoretically: https://github.com/lib/pq/issues/1007

N-o-X:
> @N-o-X Nice [list](https://github.com/Icinga/icingadb/commit/417990c9a7d6f5979121e22583b6c036106ad447). How did you collect the items?

I think I got that list from @lippserd.

julianbrost:
What's the current state of this? I think it should probably be merged before #232 or we start thinking in more detail about #234 as anything else would probably result in merge conflicts that can probably be better solved in the context of the other tasks.

Al2Klimov:
State: Waiting for feedback from @lippserd

fl0wx:
What's the status on this issue? Really would love to see postgres support in next releae

Al2Klimov:
## TODO

* [ ] .github/ISSUE_TEMPLATE/bug_report.md
* [ ] .github/ISSUE_TEMPLATE/config.yml
* [ ] .github/dependabot.yml
* [ ] .github/workflows/docker.yml
* [ ] .github/workflows/go.yml
* [ ] .github/workflows/integration-tests.yml
* [ ] .gitignore
* [ ] .mailmap
* [ ] AUTHORS
* [ ] CHANGELOG.md
* [ ] LICENSE
* [ ] README.md
* [ ] RELEASE.md
* [ ] cmd/icingadb/main.go
* [ ] config.yml.example
* [ ] doc/01-About.md
* [ ] doc/02-Installation.md
* [ ] doc/03-Configuration.md
* [ ] doc/images/about/icinga-db-in-icinga-context.png
* [ ] go.mod
* [ ] go.sum
* [ ] internal/command/command.go
* [ ] internal/internal.go
* [ ] internal/version.go
* [ ] pkg/backoff/backoff.go
* [ ] pkg/com/bulker.go
* [ ] pkg/com/com.go
* [ ] pkg/com/cond.go
* [ ] pkg/com/counter.go
* [ ] pkg/com/entity_bulker.go
* [ ] pkg/common/sync_subject.go
* [ ] pkg/config/config.go
* [ ] pkg/config/database.go
* [ ] pkg/config/redis.go
* [ ] pkg/contracts/contracts.go
* [ ] pkg/driver/driver.go
* [ ] pkg/flatten/flatten.go
* [ ] pkg/icingadb/db.go
* [ ] pkg/icingadb/delta.go
* [ ] pkg/icingadb/dump_signals.go
* [ ] pkg/icingadb/entitiesbyid.go
* [ ] pkg/icingadb/ha.go
* [ ] pkg/icingadb/history/sync.go
* [ ] pkg/icingadb/objectpacker/objectpacker.go
* [ ] pkg/icingadb/objectpacker/objectpacker_test.go
* [ ] pkg/icingadb/overdue/sync.go
* [ ] pkg/icingadb/runtime_updates.go
* [ ] pkg/icingadb/sync.go
* [ ] pkg/icingadb/v1/checkable.go
* [ ] pkg/icingadb/v1/command.go
* [ ] pkg/icingadb/v1/comment.go
* [ ] pkg/icingadb/v1/customvar.go
* [ ] pkg/icingadb/v1/downtime.go
* [ ] pkg/icingadb/v1/endpoint.go
* [ ] pkg/icingadb/v1/entity.go
* [ ] pkg/icingadb/v1/history/ack.go
* [ ] pkg/icingadb/v1/history/comment.go
* [ ] pkg/icingadb/v1/history/downtime.go
* [ ] pkg/icingadb/v1/history/flapping.go
* [ ] pkg/icingadb/v1/history/meta.go
* [ ] pkg/icingadb/v1/history/notification.go
* [ ] pkg/icingadb/v1/history/state.go
* [ ] pkg/icingadb/v1/host.go
* [ ] pkg/icingadb/v1/icingadb_instance.go
* [ ] pkg/icingadb/v1/meta.go
* [ ] pkg/icingadb/v1/notification.go
* [ ] pkg/icingadb/v1/overdue/host.go
* [ ] pkg/icingadb/v1/overdue/service.go
* [ ] pkg/icingadb/v1/service.go
* [ ] pkg/icingadb/v1/state.go
* [ ] pkg/icingadb/v1/timeperiod.go
* [ ] pkg/icingadb/v1/url.go
* [ ] pkg/icingadb/v1/user.go
* [ ] pkg/icingadb/v1/v1.go
* [ ] pkg/icingaredis/client.go
* [ ] pkg/icingaredis/heartbeat.go
* [ ] pkg/icingaredis/utils.go
* [ ] pkg/icingaredis/v1/icinga_status.go
* [ ] pkg/icingaredis/v1/stats_message.go
* [ ] pkg/retry/retry.go
* [ ] pkg/structify/structify.go
* [ ] pkg/types/acknowledgement_state.go
* [ ] pkg/types/binary.go
* [ ] pkg/types/bool.go
* [ ] pkg/types/comment_type.go
* [ ] pkg/types/float.go
* [ ] pkg/types/int.go
* [ ] pkg/types/notification_states.go
* [ ] pkg/types/notification_type.go
* [ ] pkg/types/notification_types.go
* [ ] pkg/types/state_type.go
* [ ] pkg/types/string.go
* [ ] pkg/types/unix_milli.go
* [ ] pkg/types/uuid.go
* [ ] pkg/utils/utils.go
* [ ] schema/mysql/schema.sql
* [ ] schema/mysql/upgrades/1.0.0-rc2.sql
* [ ] schema/pgsql/schema.sql
* [ ] tests/environment_test.go
* [ ] tests/go.mod
* [ ] tests/go.sum
* [ ] tests/history_test.go
* [ ] tests/instance_test.go
* [ ] tests/internal/utils/slice.go
* [ ] tests/internal/value/notification_states.go
* [ ] tests/internal/value/notification_types.go
* [ ] tests/internal/value/value.go
* [ ] tests/main_test.go
* [ ] tests/object_sync_test.conf
* [ ] tests/object_sync_test.go
* [ ] tests/state_sync_test.go
* [ ] tools/dev_update_schema.sh

Al2Klimov:
@julianbrost You have written `tests/`. Please tick them above if there's nothing MySQL-specific.

julianbrost:
Well, the tests are all MySQL-specific right now as they all do a "give me a MySQL database and connect Icinga DB to it". So there are basically two options:
* Make icinga-testing and the tests database-agnostic and then run tests twice with environment variables specifying whether to use MySQL or PostgreSQL.
* Make sub-tests for MySQL/PostgreSQL for each test-case.

The first one will make the test cases themselves a little more annoying, the latter running these tests.

pkg/icingadb/db.go
```diff
@@ -495,6 +553,21 @@ func (db *DB) Delete(ctx context.Context, entityType contracts.Entity, ids []int
...
+	if db.DriverName() == "icingadb-pgsql" {
+		var i uint64
+		return placeholder.ReplaceAllStringFunc(query, func(string) string {
+			i++
+			return "$" + strconv.FormatUint(i, 10)
+		})
+	}
+
+	return query
+}
```
julianbrost, line 569: Is there no way to have `database/sql` or `sqlx` handle this for us?

pkg/icingadb/db.go
```diff
@@ -296,6 +311,49 @@ func (db *DB) NamedBulkExec(
...
+			break
+		}
+
+		if err := tx.Commit(); err != nil {
+			return errors.Wrap(err, "can't commit transaction")
+		}
+	}
+
+	return err
+}
```
julianbrost, line 349: In what kind of situations does this happen? Ideally we shouldn't get into this situation at all as updating the same row twice is a waste of time anyways.

pkg/icingadb/db.go
```diff
@@ -296,6 +311,49 @@ func (db *DB) NamedBulkExec(
...
+
+func isDuplicateRowInUpsert(err error) bool {
+	var pe *pq.Error
+	return errors.As(err, &pe) && pe.Code == "21000" &&
+		pe.Message == "ON CONFLICT DO UPDATE command cannot affect row a second time"
```
julianbrost, line 354: Does this really need to check the error message here? Also can that message change based on server locale or something like that?

pkg/icingadb/db.go
Al2Klimov: Should there be one? This one converts one kind of placeholders to another, whereas `database/sql` is completely agnostic, the drivers only care for their own kind of placeholders and `sqlx` is about specific query args which are not involved here.

pkg/icingadb/db.go
Al2Klimov: History.

pkg/icingadb/db.go
julianbrost: There is https://pkg.go.dev/github.com/jmoiron/sqlx#DB.Rebind which does pretty much the same thing. There's also https://pkg.go.dev/github.com/jmoiron/sqlx#BindDriver, maybe we should also register the `icingadb-mysql`/`icingadb-pgsql` drivers there (might even be needed to use `DB.Rebind()`)?

pkg/icingadb/db.go
julianbrost: Do you actually need both updates there or could you just omit the first one?

pkg/icingadb/db.go
Al2Klimov: 1. Yes, 21000 is a little ambiguous.
2. I have even trouble with localizing the Postgres shell:

```
postgres@parallels-Parallels-Virtual-Platform:~$ LC_ALL=ru_RU psql
psql (12.2 (Ubuntu 12.2-4))
Type "help" for help.

postgres=# 
```

pkg/icingadb/db.go
Al2Klimov: They could affect different columns, so yes.

julianbrost (changes requested):
* [ ] Requires an update due to #370

Al2Klimov:
* [ ] ... and one due to #236.

lippserd (changes requested):
Aside from https://github.com/lib/pq/commit/6ede22e220e6529d9fde721f92a4ff624a00c63c, could you please check whether it is worth switching to https://github.com/jackc/pgx in terms of API and features?

Al2Klimov:
Worth switching to? IMAO it's a shoot into the foot:

> It is recommended to use the pgx interface if: (...) **No other libraries that require database/sql** are in use.

lippserd:
> Worth switching to? IMAO it's a shoot into the foot:
> 
> > It is recommended to use the pgx interface if: (...) **No other libraries that require database/sql** are in use.

This statement is all about using their own interface (instead of `database/sql`), which we will not use.

Al2Klimov:
I'm afraid the currently used one is even tech. better:

```
2021-10-04T17:17:23.921+0200    FATAL   Cannot encode uint8 into oid 16397 - uint8 must implement Encoder or be converted to a string
```

lippserd:
> I'm afraid the currently used one is even tech. better:
> 
> ```
> 2021-10-04T17:17:23.921+0200    FATAL   Cannot encode uint8 into oid 16397 - uint8 must implement Encoder or be converted to a string
> ```

Ok, so we have our decision.

julianbrost:
I've created a PR for icinga-testing that writes the database type to the config: https://github.com/Icinga/icinga-testing/pull/8

So if you do a `(cd tests && go get github.com/icinga/icinga-testing@2d1f0c7431846a8b26e52773b052e3e4ffce2e95)`, this should make the tests work again (for MySQL only so far).

Al2Klimov:
## Blocked by

* https://github.com/Icinga/icinga-testing/pull/8

pkg/icingadb/db.go
```diff
@@ -342,6 +365,49 @@ func (db *DB) NamedBulkExec(
 	return g.Wait()
 }
 
+// NamedExecContextBulk performs db.NamedExecContext(ctx, query, rows).
+// If that fails due to upserting the same row twice (Postgres only), it splits rows and upserts them in a transaction.
+func (db *DB) NamedExecContextBulk(ctx context.Context, query string, rows []contracts.Entity) error {
```
julianbrost, line 370: Is that function even used anywhere? `git grep NamedExecContextBulk` brings up no other occurrences than these two right here.

pkg/icingadb/db.go
julianbrost: I think it would be nicer if we could never send upserts with duplicate keys to Postgres in the first place (this would also avoid having to perform string matching on error messages). The situation should happen if and only if there are multiple entities in `rows` that return the same value for `rows[i].ID()`. So I think it should be possible to just check this before sending the first query and then splitting the rows into a new query before each duplicate.

julianbrost (commented):
Just a few comments/questions on the schema for now.

schema/pgsql/schema.sql
```diff
@@ -0,0 +1,1904 @@
...
+CREATE DOMAIN bytea20 AS bytea CONSTRAINT exactly_20_bytes_long CHECK ( VALUE IS NULL OR octet_length(VALUE) = 20 );
+CREATE DOMAIN bytea16 AS bytea CONSTRAINT exactly_16_bytes_long CHECK ( VALUE IS NULL OR octet_length(VALUE) = 16 );
+CREATE DOMAIN bytea4 AS bytea CONSTRAINT exactly_4_bytes_long CHECK ( VALUE IS NULL OR octet_length(VALUE) = 4 );
+
+CREATE DOMAIN biguint AS bigint CONSTRAINT positive CHECK ( VALUE IS NULL OR 0 <= VALUE );
+CREATE DOMAIN uint AS bigint CONSTRAINT between_0_and_4294967295 CHECK ( VALUE IS NULL OR VALUE BETWEEN 0 AND 4294967295 );
+CREATE DOMAIN smalluint AS int CONSTRAINT between_0_and_65535 CHECK ( VALUE IS NULL OR VALUE BETWEEN 0 AND 65535 );
+CREATE DOMAIN tinyuint AS smallint CONSTRAINT between_0_and_255 CHECK ( VALUE IS NULL OR VALUE BETWEEN 0 AND 255 );
+
+CREATE TYPE boolenum AS ENUM ( 'n', 'y' );
```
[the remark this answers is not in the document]
lippserd: `boolean` supports `NULL`, so we need to make sure that the `NOT NULL` constraint is added and the default value is set to `false`. The sort order would be correct because `boolean` sorts by `false`, `true`, `null`. But I think the `SELECT` value makes the difference here as it yields `t` or `f`. To be honest, I would just stick to the ENUM.

schema/pgsql/schema.sql
```diff
@@ -0,0 +1,1904 @@
+-- Icinga DB | (c) 2021 Icinga GmbH | GPLv2+
+
+CREATE DOMAIN bytea20 AS bytea CONSTRAINT exactly_20_bytes_long CHECK ( VALUE IS NULL OR octet_length(VALUE) = 20 );
+CREATE DOMAIN bytea16 AS bytea CONSTRAINT exactly_16_bytes_long CHECK ( VALUE IS NULL OR octet_length(VALUE) = 16 );
+CREATE DOMAIN bytea4 AS bytea CONSTRAINT exactly_4_bytes_long CHECK ( VALUE IS NULL OR octet_length(VALUE) = 4 );
+
+CREATE DOMAIN biguint AS bigint CONSTRAINT positive CHECK ( VALUE IS NULL OR 0 <= VALUE );
+CREATE DOMAIN uint AS bigint CONSTRAINT between_0_and_4294967295 CHECK ( VALUE IS NULL OR VALUE BETWEEN 0 AND 4294967295 );
+CREATE DOMAIN smalluint AS int CONSTRAINT between_0_and_65535 CHECK ( VALUE IS NULL OR VALUE BETWEEN 0 AND 65535 );
+CREATE DOMAIN tinyuint AS smallint CONSTRAINT between_0_and_255 CHECK ( VALUE IS NULL OR VALUE BETWEEN 0 AND 255 );
```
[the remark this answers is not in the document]
lippserd: Why should we do that? Because of storage? I don't think you want to keep track of possible separate values for these columns for MySQL and PostgreSQL. It always has to be the same. So yes, we need those types.

schema/pgsql/schema.sql
```diff
@@ -0,0 +1,1904 @@
...
+CREATE TYPE history_type AS ENUM ( 'notification', 'state_change', 'downtime_start', 'downtime_end', 'comment_add', 'comment_remove', 'flapping_start', 'flapping_end', 'ack_set', 'ack_clear' );
+
+CREATE TABLE host (
+  id bytea20 NOT NULL,
+  environment_id bytea20 NOT NULL,
+  name_checksum bytea20 NOT NULL,
+  properties_checksum bytea20 NOT NULL,
+
+  name varchar(255) NOT NULL,
+  name_ci varchar(255) NOT NULL,
```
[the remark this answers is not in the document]
lippserd: That would require code in Web and Icinga DB to handle this column in a special way. I would not do that. PostgreSQL queries aren't that different.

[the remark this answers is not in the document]
julianbrost: But wouldn't the current variant require the PostgreSQL SELECT queries do an extra `LOWER(name_ci)` anyways? Without having looked into this in too much detail, I think the `textci` type might give the behavior most similar to MySQL `COLLATE ..._ci`.

schema/pgsql/schema.sql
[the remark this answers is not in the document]
julianbrost: One would have to look into the individual columns: what is stored in there and what's the best type supported by the database to store this. IMHO this might very well result in different types for MySQL and PostgreSQL.

schema/pgsql/schema.sql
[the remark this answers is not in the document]
lippserd: Just wanted to say that I want to keep both columns. Sure, it would be perfect if we didn't need the `LOWER`. But if so, we can just add it if the column ends with `_ci`.

schema/pgsql/schema.sql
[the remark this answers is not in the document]
lippserd: I do not understand what you are trying to say. The types are there so that there are no possible different values for these types in MySQL and PostgreSQL.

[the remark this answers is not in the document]
julianbrost: I'm trying to say that columns in MySQL are probably `smallint unsigned` because that was the best-fitting type that MySQL had to offer, not because the values we want to store use the full `uint16` range. So the best type PostgreSQL has to offer for the same column might as well be a `smallint` or maybe also a `bit(n)` (maybe that would make sense if we store a bitset) or something totally different.

The domains introduced here look to me like trying to make it look like MySQL at all cost.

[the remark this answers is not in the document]
lippserd: > The domains introduced here look to me like trying to make it look like MySQL at all cost.

Of course, that's the whole point. But it's about the behavior! If a value is wrong it's equally wrong in both databases. But without something like that there might be totally different values accepted. And since MySQL doesn't support custom data types, it's just the way it is.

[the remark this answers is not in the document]
julianbrost: I'm sorry, but then I don't get the point. Let's take the very first column using one of these domains as an example: `host.max_check_attempts`.  Does it make a meaningful difference if this is stored as `int` or as `bigint` constrained to unsigned 32 bit values? Both support all sane values for that config option and neither allows the full range Icinga 2 would accept for this value. Therefore, I'd go with the simpler variant and just use `int` (or possibly even an unconstrained `bigint` to allow everything allowed by Icinga 2, even if that configuration wouldn't make any sense).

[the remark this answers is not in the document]
lippserd: What matters here is not what makes sense for the possible configuration values, but what is set in the MySQL schema. The maximum value there is **4294967295**. This has to be the same in PostgreSQL.

schema/pgsql/schema.sql
```diff
@@ -0,0 +1,1904 @@
...
+  CONSTRAINT pk_hostgroup PRIMARY KEY (id)
+);
+
+ALTER TABLE hostgroup ALTER COLUMN id SET STORAGE PLAIN;
+ALTER TABLE hostgroup ALTER COLUMN environment_id SET STORAGE PLAIN;
+ALTER TABLE hostgroup ALTER COLUMN name_checksum SET STORAGE PLAIN;
+ALTER TABLE hostgroup ALTER COLUMN properties_checksum SET STORAGE PLAIN;
+ALTER TABLE hostgroup ALTER COLUMN zone_id SET STORAGE PLAIN;
+
+CREATE INDEX idx_hostroup_name ON hostgroup(name);
```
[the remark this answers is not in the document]
Al2Klimov: Other construction area :P https://github.com/Icinga/icingadb/search?q=idx_hostroup_name

schema/pgsql/schema.sql
```diff
@@ -0,0 +1,1904 @@
...
+  icon_image_id bytea20 DEFAULT NULL,
+  icon_image_alt varchar(32) NOT NULL,
+
+  zone varchar(255) NOT NULL,
+  zone_id bytea20 DEFAULT NULL,
+
+  command_endpoint varchar(255) NOT NULL,
+  command_endpoint_id bytea20 DEFAULT NULL,
+
+  CONSTRAINT pk_host PRIMARY KEY (id)
```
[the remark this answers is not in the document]
Al2Klimov: To give that a name.

schema/pgsql/schema.sql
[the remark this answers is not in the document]
julianbrost: Why is the allowed range now defined by the MySQL schema and not by Icinga 2? After all, the row represents an Icinga 2 object.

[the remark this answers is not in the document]
lippserd: I can only repeat myself as this has nothing to do with this discussion. Same schemes are a must. No matter what permissible values we have in MySQL. These must be exactly the same in PostgreSQL.

julianbrost (changes requested):
Deletes don't seem to work. Runtime deletions don't seem to do anything, row remains in the database and I couldn't find any related logs, deletion during the initial sync results in the following crash:
```
2021-11-24T15:11:07.313+0100	FATAL	icingadb	number of bindVars less than number arguments
can't build placeholders for "DELETE FROM \"user\" WHERE id IN ($1)"
github.com/icinga/icingadb/pkg/icingadb.(*DB).BulkExec.func1.1.1.1
	/home/jbrost/dev/icingadb/pkg/icingadb/db.go:261
	/home/jbrost/dev/icingadb/pkg/retry/retry.go:43
github.com/icinga/icingadb/pkg/icingadb.(*DB).BulkExec.func1.1.1
	/home/jbrost/dev/icingadb/pkg/icingadb/db.go:256
	/home/jbrost/.local/share/go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57
runtime.goexit
	/usr/lib/go/src/runtime/asm_amd64.s:1581
can't retry
	/home/jbrost/dev/icingadb/pkg/retry/retry.go:62
github.com/icinga/icingadb/pkg/icingadb.(*DB).BulkExec.func1.1.1
	/home/jbrost/dev/icingadb/pkg/icingadb/db.go:256
	/home/jbrost/.local/share/go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57
runtime.goexit
	/usr/lib/go/src/runtime/asm_amd64.s:1581
```

julianbrost:
Also after receiving a deletion runtime update, upserts are no longer handled. But maybe that's also fixed by just fixing deletes. I hope this will fix quite some tests.

julianbrost:
Deletions and runtime updates work now. Any idea why there was no error in the log for the runtime updates?

pkg/com/entity_bulker.go
```diff
@@ -16,14 +64,16 @@ type EntityBulker struct {
 }
 
 // NewEntityBulker returns a new EntityBulker and starts streaming.
-func NewEntityBulker(ctx context.Context, ch <-chan contracts.Entity, count int) *EntityBulker {
+func NewEntityBulker(
+	ctx context.Context, ch <-chan contracts.Entity, count int, splitPolicy BulkChunkSplitPolicy,
```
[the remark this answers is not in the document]
Al2Klimov: Doesn’t the docs clearly say: it’s a state machine?

schema/pgsql/schema.sql
```diff
@@ -0,0 +1,1904 @@
...
+  performance_data text DEFAULT NULL,
+  normalized_performance_data text DEFAULT NULL,
+
+  check_commandline text DEFAULT NULL,
+
+  is_problem boolenum NOT NULL,
+  is_handled boolenum NOT NULL,
+  is_reachable boolenum NOT NULL,
+  is_flapping boolenum NOT NULL,
+  is_overdue boolenum NOT NULL DEFAULT 'n',
```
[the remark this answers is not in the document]
Al2Klimov: The state sync would error.

doc/02-Installation.md
```diff
@@ -201,6 +202,80 @@ After creating the database, you can import the Icinga DB schema using the follo
 mysql -u root -p icingadb </usr/share/icingadb/schema/mysql/schema.sql
 ```
 
+### Configuring PostgreSQL <a id="configuring-icingadb-pgsql"></a>
```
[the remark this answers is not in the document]
Al2Klimov: I have no doubt Postgressers will find their way through the docs.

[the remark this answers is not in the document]
julianbrost: No doubt that there are users that will get it right. But if someone just follows the instruction from top to bottom, they might end up setting up both. So why not group both sections in a new section that starts with an instruction that states these are alternatives and you have to set up just one, probably even with a suggestion which one to choose if you don't know yourself.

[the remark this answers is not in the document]
Al2Klimov: > But if someone just follows the instruction from top to bottom, they might end up setting up both.

OK.. so you think.. once someone has set up MySQL and hit the setup Postgres headline.. they'd not consider it strange to setup both?

schema/pgsql/schema.sql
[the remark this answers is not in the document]
julianbrost: Otherwise you explicitly have to give a value and this happens to be the only one where we don't do this? So if we do add boolean columns in the future, we might have to add them with `DEFAULT 'n'` as well to keep the schema compatible.

If we keep the `boolenum` ([which I'm still not a fan of to be honest](https://github.com/Icinga/icingadb/pull/221#discussion_r755860868)), maybe add the default to the type definition instead to keep all these columns consistent in the long-term?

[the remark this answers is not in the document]
Al2Klimov: CC @lippserd

doc/02-Installation.md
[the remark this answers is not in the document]
julianbrost: You're assuming that everyone installing Icinga DB is familiar with Redis, MySQL, MariaDB, and PostgreSQL. The only hint so far that you need Redis and one of the remaining three is that at the top of the document, there is one bullet point for Redis and one for MySQL/MariaDB/PostgreSQL.

So I think this would a rather simple change for us that improves the documentation so that it could save someone from pointlessly setting up multiple RDMBS.

pkg/com/entity_bulker.go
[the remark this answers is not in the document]
julianbrost: This is just a suggestion to improve code quality by changing the API in a way that makes it harder to use wrong.

doc/02-Installation.md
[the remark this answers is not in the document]
Al2Klimov: Ah, wait! Isn’t https://github.com/Icinga/icingadb/pull/221/files/a357ed9516febb6c99230f2efca21ff03a736bac#diff-a8d6a4bf6bf9e106f20ce651f485c61a9884fe3e30fbd30598e631aa33a3e66bR6-R7 enough?

pkg/com/entity_bulker.go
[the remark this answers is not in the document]
Al2Klimov: I'd say it more general: `s/ wrong//`

Seriously, our API "users" are IT specialists in this case.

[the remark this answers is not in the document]
julianbrost: I don't see how passing `NeverSplit` instead of `NeverSplit{}` and `SplitOnDupId` instead of `NewSplitOnDupId()` is any harder to use.

schema/pgsql/schema.sql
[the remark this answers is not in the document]
julianbrost: > Sure, it would be perfect if we didn't need the `LOWER`. But if so, we can just add it if the column ends with `_ci`.

This would work for this column, but there are quite a few more that have no special naming but still are `COLLATE *_ci` in MySQL and have an index on `LOWER(col)` here.

julianbrost (commented):
Let me try so summarize and classify everything pending in this PR.

Looks fine overall apart from some small things mentioned below. There are tests for it in #418 that pass. Unfortunately, the base commit of this PR is quite old and I based the tests on this PR merged with a more recent version of the master branch and the tests can't simply be rebased onto this PR without having to resolve these. So I'd either keep the tests as the separate PR or this PR would have to be rebased.

Things that must be addressed before merging:
* There need to be corresponding changes to our packaging repos adding `type: mysql` to the default config as this is a required setting now. Looks like those were not created so far.
* a357ed9516febb6c99230f2efca21ff03a736bac should be removed again from the PR as there will be #418 instead.

Open comments, that haven't received a lot of attention so far:
* https://github.com/Icinga/icingadb/pull/221#discussion_r755845684: we should decide how to handle case-insensitive columns, probably also coordinate that with team web as they'll have to use it mostly.
* https://github.com/Icinga/icingadb/pull/221#discussion_r755849053
* https://github.com/Icinga/icingadb/pull/221#discussion_r755854563: I wouldn't intentionally add new things with a typo, just because the same typo exists in MySQL already. Rather fix it now than knowing you'll have to write a schema change you could easily avoid.

Things that I think could still be changed but I see that we have disagreeing opinions on some of these and I won't block this PR of any of these:
* https://github.com/Icinga/icingadb/pull/221#discussion_r755860868 and https://github.com/Icinga/icingadb/pull/221#discussion_r755864505: both look to me like this wouldn't be done this way if we did the PostgreSQL schema first, i.e. I think both aren't the best way to do this in PostgreSQL and making the PostgreSQL-support look like a second-class citizen to me. (And I didn't even mention types like [`uuid`](https://www.postgresql.org/docs/14/datatype-uuid.html) and [`inet`](https://www.postgresql.org/docs/14/datatype-net-types.html#DATATYPE-INET) yet.)
* https://github.com/Icinga/icingadb/pull/221#discussion_r758547937
* https://github.com/Icinga/icingadb/pull/221#discussion_r758562355

schema/pgsql/schema.sql
```diff
@@ -0,0 +1,1904 @@
...
+  version smalluint NOT NULL,
+  timestamp biguint NOT NULL,
+
+  CONSTRAINT pk_icingadb_schema PRIMARY KEY (id)
+);
+
+ALTER SEQUENCE icingadb_schema_id_seq OWNED BY icingadb_schema.id;
+
+INSERT INTO icingadb_schema (version, timestamp)
+  VALUES (2, extract(epoch from now()) * 1000);
```
julianbrost, line 1904: Do we really want to start with version 2 here and try to keep this version number in sync with the MySQL schema? If we'd to, for a change that only affects one schema file (like for example #420), we'd also have to ship an upgrade for the other that does nothing but bump the schema version.

julianbrost (commented):
Oh and of course I've forgot something in that long comment:

schema/pgsql/schema.sql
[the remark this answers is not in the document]
Al2Klimov: I doubt it would be much harder for Web to maintain a list of them. Or to even parse `SHOW CREATE TABLE`.

Al2Klimov:
Wait! Our PostgreSQL-support (in general) is not a second-class citizen? (I have to note this for myself in an instant...)

schema/pgsql/schema.sql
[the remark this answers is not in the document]
lippserd: The `DEFAULT 'n'` should be part of the `CREATE TYPE` specification as we mirror the behavior of MySQL.

doc/02-Installation.md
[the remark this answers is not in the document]
lippserd: I support Julian's thoughts here.

julianbrost:
> Wait! Our PostgreSQL-support (in general) is not a second-class citizen?

Well, that's something I would want to avoid, but... :shrug:

Al2Klimov:
Oh, and.. Postgres uuid? What for?

schema/pgsql/schema.sql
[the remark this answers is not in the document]
lippserd: Why not just fix it here and create a PR that fixes other issues?

julianbrost:
That was just a type that I had to think about. But looks like we've removed most UUIDs anyways (expect `icingadb_instance.id`).

Al2Klimov:
And.. you expect that we remove that as well?

Al2Klimov:
* [ ] https://git.icinga.com/packaging/rpm-icingadb/-/merge_requests/17
* [ ] https://git.icinga.com/packaging/deb-icingadb/-/merge_requests/7

lippserd:
> * There need to be corresponding changes to our packaging repos adding `type: mysql` to the default config as this is a required setting now. Looks like those were not created so far.

I think `type: mysql` should be the default.

> * both look to me like this wouldn't be done this way if we did the PostgreSQL schema first, i.e. I think both aren't the best way to do this in PostgreSQL and making the PostgreSQL-support look like a second-class citizen to me.

If we had created the PostgreSQL schema first, we would have to do the workarounds in the MySQL schema.
I don't think this discussion is going anywhere as "identical" schemes are a must.

julianbrost:
> And.. you expect that we remove that as well?

No. What I wanted to say is that PostgreSQL has quite a lot of more specific types that might be useful if we decided to make use of them. But seems like I'm the only one who would want to do this anyways.

schema/pgsql/schema.sql
lippserd: Separating the schema versions makes sense to me.

schema/pgsql/schema.sql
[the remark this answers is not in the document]
lippserd: Oh noes!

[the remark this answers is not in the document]
julianbrost: I thought I explicitly checked this in the docs, but looks like that was for domains and not types: https://www.postgresql.org/docs/14/sql-createdomain.html

Al2Klimov:
> If we had created the PostgreSQL schema first, we would have to do the workarounds in the MySQL schema.

Even better: if we had created the PostgreSQL schema first (God preserve us!), we'd even have to degrade that schema for "equality" with MySQL.

schema/pgsql/schema.sql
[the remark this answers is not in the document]
lippserd: I see. Is `CREATE DOMAIN boolenum AS CHAR(1) CHECK (VALUE IN ('n', 'y')) DEFAULT 'n';` an option? I did not check how to force a specific sort order.

schema/pgsql/schema.sql
```diff
@@ -0,0 +1,1912 @@
...
+CREATE DOMAIN bytea20 AS bytea CONSTRAINT exactly_20_bytes_long CHECK ( VALUE IS NULL OR octet_length(VALUE) = 20 );
+CREATE DOMAIN bytea16 AS bytea CONSTRAINT exactly_16_bytes_long CHECK ( VALUE IS NULL OR octet_length(VALUE) = 16 );
+CREATE DOMAIN bytea4 AS bytea CONSTRAINT exactly_4_bytes_long CHECK ( VALUE IS NULL OR octet_length(VALUE) = 4 );
+
+CREATE DOMAIN biguint AS bigint CONSTRAINT positive CHECK ( VALUE IS NULL OR 0 <= VALUE );
+CREATE DOMAIN uint AS bigint CONSTRAINT between_0_and_4294967295 CHECK ( VALUE IS NULL OR VALUE BETWEEN 0 AND 4294967295 );
+CREATE DOMAIN smalluint AS int CONSTRAINT between_0_and_65535 CHECK ( VALUE IS NULL OR VALUE BETWEEN 0 AND 65535 );
+CREATE DOMAIN tinyuint AS smallint CONSTRAINT between_0_and_255 CHECK ( VALUE IS NULL OR VALUE BETWEEN 0 AND 255 );
+
+CREATE TYPE boolenum_t AS ENUM ( 'n', 'y' );
```
julianbrost, line 12:
```
2021-12-06T16:41:09.340+0100	FATAL	icingadb	pq: operator does not exist: boolenum = unknown
can't perform "SELECT id, heartbeat FROM icingadb_instance WHERE environment_id = $1 AND responsible = $2 AND id <> $3 AND heartbeat > $4"
```

Looks like you have to do `WHERE responsible::boolenum_t = 'y'` in queries now, so not sure if this is indeed a good idea.

julianbrost (changes requested):

schema/pgsql/schema.sql
Al2Klimov: You mean `WHERE responsible = 'y'::boolenum_t`?

julianbrost: No? (Don't ask me why it's that way)

```
icingadb=# SELECT COUNT(*) FROM icingadb_instance WHERE responsible::boolenum_t = 'y';
 count 
-------
     0
(1 row)

icingadb=# SELECT COUNT(*) FROM icingadb_instance WHERE responsible = 'y'::boolenum_t;
ERROR:  operator does not exist: boolenum = boolenum_t
LINE 1: ...COUNT(*) FROM icingadb_instance WHERE responsible = 'y'::boo...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

julianbrost: Second line is probably a hint why it doesn't work:
```
icingadb=# SELECT COUNT(*) FROM icingadb_instance WHERE responsible = 'y'::boolenum;
ERROR:  operator does not exist: boolenum = boolenum
LINE 1: ...COUNT(*) FROM icingadb_instance WHERE responsible = 'y'::boo...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

Al2Klimov: And w/o _t?

Al2Klimov: `s/responsible/responsible::boolenum_t/` in Go or ENUM -> CHAR(1) CHECK...?

julianbrost: > `s/responsible/responsible::boolenum_t/` in Go

No way, that would defeat the whole purpose of trying to behave somewhat like MySQL. Web would have to do the same for each query. Then we could just use the standard boolean type with less impact.

lippserd: https://github.com/Icinga/icingadb/pull/221#discussion_r759395335

schema/pgsql/schema.sql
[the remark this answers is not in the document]
Al2Klimov: Sort order.. ah! Right! n < y < sticky e.g.!

schema/pgsql/schema.sql
Al2Klimov: As we're already talking about schema equality and sort order.. what if we explicitly specify the DEFAULTs for both RDBMS', drop the boolenum DOMAIN in favour of the boolenum_t TYPE and.. it will just work?

lippserd: We don't have to change anything for MySQL:

> If an ENUM column is declared to permit NULL, the NULL value is a valid value for the column, and the default value is NULL. If an ENUM column is declared NOT NULL, its default value is the first element of the list of permitted values.

> ENUM values are sorted based on their index numbers, which depend on the order in which the enumeration members were listed in the column specification

https://dev.mysql.com/doc/refman/8.0/en/enum.html

Al2Klimov: OK.. so I'd just do https://github.com/Icinga/icingadb/pull/221#discussion_r765670379 w/o touching the MySQL schema.. ?

lippserd: I don't know. The sort order has not yet been clarified. You should find out what we should do.

Al2Klimov: 1. "The ordering of the values in an enum type is the order in which the values were listed when the type was created." – https://www.postgresql.org/docs/9.1/datatype-enum.html
2. I've just suggested what we should do. To be precise: compared to status quo (DOMAINs) I'd like to make the columns of just the ENUM types and explicitly specify a default per column.

Al2Klimov: Yes, I know, schema equality.. but in this case we'd provide schema equality – effective defaults – by having schema unequality – explicitly (not) specified defaults.

lippserd: We don't have `DOMAIN` for the bools in PostgreSQL at the moment. We have `ENUM`. And the problem with that is that it requires casting, as far as I understand @julianbrost correctly.

Al2Klimov: That's not true:

https://github.com/Icinga/icingadb/blob/74d8bb9a171d31e79e372d31128ebd1aebc768fe/schema/pgsql/schema.sql#L12-L26

https://github.com/Icinga/icingadb/blob/74d8bb9a171d31e79e372d31128ebd1aebc768fe/schema/pgsql/schema.sql#L55-L58

julianbrost: > And the problem with that is that it requires casting, as far as I understand @julianbrost correctly.

I don't know in detail if there's a way to fix it but on PostgreSQL 14 with the current enum + type + domain construction, I wasn't able to do any queries without casting.

Al2Klimov: OK, let's formulate the question as unambiguous as possible:

If I'd do `CREATE TYPE boolenum AS ENUM ( 'n', 'y' );` and `active_checks_enabled boolenum NOT NULL DEFAULT 'n',` for all columns and it would work, would you accept it or not?

Al2Klimov: Damn. Seems its a s/bug/feature/: https://www.postgresql.org/message-id/2971.1294769104%40sss.pgh.pa.us

Al2Klimov: What about this (for all of our domains)?

```
idb=# SELECT COUNT(*) FROM icingadb_instance WHERE responsible = 'y';
ERROR:  operator does not exist: boolenum = unknown
LINE 1: ...ECT COUNT(*) FROM icingadb_instance WHERE responsible = 'y';
                                                                 ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
idb=# CREATE FUNCTION compare(boolenum, unknown) RETURNS boolean AS 'SELECT $1::boolenum_t = $2::char::boolenum_t;' LANGUAGE SQL IMMUTABLE RETURNS NULL ON NULL INPUT;
CREATE FUNCTION
idb=# CREATE OPERATOR = (LEFTARG = boolenum, RIGHTARG = unknown, PROCEDURE = compare);
CREATE OPERATOR
idb=# SELECT COUNT(*) FROM icingadb_instance WHERE responsible = 'y';
 count
-------
     0
(1 row)

idb=#
```

@lippserd @nilmerg Is web using more operators?

julianbrost: Does `ORDER BY` work?

julianbrost: But I wouldn't be surprised if it needed an operator `<`. Have you tested it?

Anyways, I'm not convinced of this idea. Your question shouldn't be if web uses more operators but rather also if web or anything else might want to use other operators in the future. And if we don't get that answer right now, we'd have to upgrade the schema and add operators. So I think just adding `DEFAULT 'n'` to the column definitions is the easier way.

Al2Klimov: Surprisingly the non-ENUM domains just work:

```
idb=# \x
Expanded display is on.
idb=# select * from history order by event_time;
(0 rows)

idb=# select * from history where event_time > 0 order by event_time;
(0 rows)

idb=# select * from history where event_time = 0 order by event_time;
(0 rows)

idb=# select * from host where id = 'x';
(0 rows)

idb=#
```

README.md
```diff
@@ -13,9 +13,9 @@
 
 ## About
 
-Icinga DB serves as a synchronisation daemon between Icinga 2 (Redis) and Icinga Web 2 (MySQL). It synchronises configuration, state and history of an Icinga 2 environment using checksums.
+Icinga DB serves as a synchronisation daemon between Icinga 2 (Redis) and Icinga Web 2 (RDBMS). It synchronises configuration, state and history of an Icinga 2 environment using checksums.
```
lippserd, line 16: I would write `database` instead of `RDBMS` since we're speaking of `database` later on anyways. Also it makes sense to clarify that `database` means MySQL, MariaDB or PostgreSQL.

cmd/icingadb/main.go
```diff
@@ -317,6 +318,14 @@ func run() int {
 
 // checkDbSchema asserts the database schema of the expected version being present.
 func checkDbSchema(ctx context.Context, db *icingadb.DB) error {
+	var expectedDbSchemaVersion uint16
+	switch db.DriverName() {
+	case "icingadb-mysql":
```
lippserd, line 323: I would introduce driver constants for all `icingadb-*` occurrences as in 71f3b494d15ce4ada2e330cdc70fc624923db223.

doc/01-About.md
```diff
@@ -2,6 +2,6 @@
 
 ![Icinga DB Context](images/about/icinga-db-in-icinga-context.png)
 
-Icinga DB serves as a synchronisation daemon between Icinga 2 (Redis) and Icinga Web 2 (MySQL). It synchronises configuration, volatile states and history of an Icinga 2 environment using checksums.
+Icinga DB serves as a synchronisation daemon between Icinga 2 (Redis) and Icinga Web 2 (RDBMS). It synchronises configuration, volatile states and history of an Icinga 2 environment using checksums.
```
lippserd, line 5: See requested changes for `README.md`.

doc/02-Installation.md
```diff
@@ -202,6 +203,80 @@ After creating the database, you can import the Icinga DB schema using the follo
...
+
+#### Installing PostgreSQL database server <a id="installing-database-pgsql-server"></a>
```
lippserd, line 208: Please remove this chapter. I will also remove the MySQL installation instructions in a separate PR.

doc/02-Installation.md
```diff
@@ -202,6 +203,80 @@ After creating the database, you can import the Icinga DB schema using the follo
...
+systemctl start postgresql
+```
+
+Debian/Ubuntu:
+
+```
+apt install postgresql postgresql-client
+```
+
+#### Setting up the PostgreSQL database <a id="setting-up-pgsql-db"></a>
```
lippserd, line 246: I would only speak of `pg_hba.conf` and not list all alternative paths for the different operating systems.

doc/02-Installation.md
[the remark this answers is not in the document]
lippserd: I think it makes sense to categorize it like this:

### Setting Up the Database

Small explanation that either a MySQL/MariaDB or a PostgreSQL database needs to be created. It can also be useful to explain the function of the database, i.e. that is the data source for Icinga Web 2, reporting and so on.

#### MySQL/MariaDB

#### PostgreSQL

doc/03-Configuration.md
```diff
@@ -24,6 +24,7 @@ Configuration of the database used by Icinga DB.
 
 Option                   | Description
 -------------------------|-----------------------------------------------
+type                     | **Required.** Either "mysql" or "pgsql".
```
lippserd, line 27: Introducing `type` as required will break all installations. I would make it optional and leave it as `mysql` by default.

pkg/config/database.go
```diff
@@ -11,14 +12,19 @@ import (
 	"github.com/jmoiron/sqlx/reflectx"
 	"github.com/pkg/errors"
 	"net"
+	"net/url"
+	"strconv"
 	"sync"
 	"time"
 )
 
+var ErrBadDBType = goerrors.New("unknown database type")
```
lippserd, line 21: I would make this a function that reports `unkown database type $type. Must be one of mysql or postgresql` and uses `errors.New()` (from "github.com/pkg/errors") in order to provide the correct stack.

pkg/config/database.go
```diff
@@ -35,30 +41,73 @@ func (d *Database) Open(logger *logging.Logger) (*icingadb.DB, error) {
...
+		}
+
+		dsn = config.FormatDSN()
+	case "pgsql":
+		uri := &url.URL{
+			Scheme:   "postgres",
+			User:     url.UserPassword(d.User, d.Password),
+			Host:     net.JoinHostPort(d.Host, strconv.FormatInt(int64(d.Port), 10)),
+			Path:     "/" + url.PathEscape(d.Database),
+			RawQuery: "connect_timeout=60&binary_parameters=yes",
```
lippserd, line 76: I think using `url.Values` makes query string manipulation much cleaner to read.

Al2Klimov:
IMAO it feels better as-is.

cmd/icingadb/main.go
Al2Klimov: Let's see if/when that gets merged... wait! History **retention**? Never mind, other construction area.

doc/02-Installation.md
```diff
@@ -3,7 +3,8 @@
 ## Requirements <a id="installation-requirements"></a>
 
 * Local Redis instance (Will be installed during this documentation)
-* MySQL/MariaDB server with `icingadb` database, user and schema imports (Will be installed during this documentation)
+* An RDBMS, either MySQL/MariaDB or PostgreSQL, with `icingadb` database,
```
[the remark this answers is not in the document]
Al2Klimov: You didn’t write it that short just to write less (just the RDBMS thing), but I shall **shorten as you did**, right?

cmd/icingadb/main.go
lippserd: This has nothing to do with history retention. It introduces constants for the driver names we support that actually serve a purpose, rather than just using a random looking string over and over again. Also, constants make it easier to see the usage and what's actually available.

doc/02-Installation.md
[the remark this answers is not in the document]
lippserd: I just want to write database instead of RDBMS. Also creating a schema and a user is implicit to me when creating a database for an application.

doc/02-Installation.md
lippserd: You can do it if you want.

lippserd:
> IMAO it feels better as-is.

Are you referring to `IsRetryable()` and `IsDeadlock()` with this answer?

doc/02-Installation.md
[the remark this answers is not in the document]
Al2Klimov: OK, for me the latter is **definitively** another construction area. If we need this, we need this independent of Postgres support. (The small explanation is already there.)

pkg/config/database.go
Al2Klimov: Let me guess: to follow already existing code patterns?

lippserd: > Let me guess: to follow already existing code patterns?

Yes.

doc/02-Installation.md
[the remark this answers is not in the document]
lippserd: Since I choose to introduce the Database Setup chapter in this PR, it only makes sense to add a little explanation as well.

cmd/icingadb/main.go
Al2Klimov: OK. Btw. PGSQL shouldn’t exist in 71f3b494d15ce4ada2e330cdc70fc624923db223.

doc/02-Installation.md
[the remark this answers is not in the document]
Al2Klimov: Not quite sure about the latter (think of Docker-ized apps using a regular RDBMS or Director). Let's hear a ~~second~~ third opinion. @julianbrost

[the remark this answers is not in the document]
lippserd: We don’t state that anywhere else. Just remove it.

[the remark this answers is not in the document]
julianbrost: Regarding RDBMS: I see that this is done as technically, Redis is also a database to distinguish them, but it's probably somewhat confusing for users that just want set up Icinga DB and aren't too familiar with databases. So my suggestion here would be to list the three options where they first appear (introduction/requirements/installation) and refer as "SQL database" to them later on (still differentiates it from Redis and should be understood by more people than RDBMS (yes, both terms don't refer to the same thing, but these details aren't relevant here)).

Regarding the detail level mentioning user and schema: the current version is somewhat too specific, we don't really care if database and user are called `icingadb`. Also, the relevant information here is that you need one of MySQL/MariaDB/PostgreSQL, the users/permissions needed will be explained later.

So I support the "A MySQL/MariaDB or PostgreSQL database" part of the suggestion, but "(will be created during this documentation)" and "Database installation should not be part of this documentation" contradict each other.

doc/02-Installation.md
Al2Klimov: Btw.: why?

Al2Klimov:
All in all: I'd address $DOC_ISSUE separately (before this PR) if $DOC_ISSUE already exists in master.

doc/02-Installation.md
lippserd: There are so many tutorials on how to install a database that it makes no sense to explain it in each of our products that require a database.

doc/02-Installation.md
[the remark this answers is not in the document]
lippserd: This change is still missing.

doc/03-Configuration.md
```diff
@@ -24,6 +24,7 @@ Configuration of the database used by Icinga DB.
 
 Option                   | Description
 -------------------------|-----------------------------------------------
+type                     | **Optional.** Either "mysql" (default) or "pgsql".
```
lippserd, line 27: Please use backticks for `mysql` and `postgresql`.

doc/02-Installation.md
Al2Klimov: "in each of our products"... OK, this is definitely true. But IMAO we should at least provide a link to a tutorial. Maybe you already heard of that discussion, maybe not, so I'll tell you (again):

@gunnarbeutner doubted the need of a such chapter.
@dnsmichi : "I won’t explain to anyone how to setup a database."

To me @dnsmichi seemed to expect a lot of questions. Actually the same question.

I have no problem with not providing even a link, but... I won’t explain to anyone how to setup a database. :P

doc/02-Installation.md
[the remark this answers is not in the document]
Al2Klimov: I consider this PR is already large enough. Why don’t we improve the DB docs separately and I'll rebase later on? Also you seem to understand far better than me what/how/why to do on that.

doc/03-Configuration.md
lippserd: Ping.

lippserd (changes requested):

pkg/config/database.go
```diff
@@ -35,30 +38,74 @@ func (d *Database) Open(logger *logging.Logger) (*icingadb.DB, error) {
...
+				query["sslrootcert"] = []string{d.TlsOptions.Ca}
+			}
+		} else {
+			query["sslmode"] = []string{"disable"}
+		}
+
+		uri.RawQuery = query.Encode()
+		dsn = uri.String()
+	default:
+		return nil, errors.Errorf(`unknown database type %q, must be one of: "mysql", "pgsql"`, d.Type)
```
lippserd, line 105: I would use a function for both error constructions as with https://github.com/Icinga/icingadb/blob/master/pkg/logging/logging.go#L129

doc/02-Installation.md
```diff
@@ -133,7 +133,11 @@ Debian/Ubuntu:
 apt-get install icingadb-redis
 ```
 
-### Setting up the MySQL database <a id="setting-up-mysql-db"></a>
+### Setting up the database <a id="setting-up-db"></a>
```
lippserd, line 136: I know it wasn't the case before, but please write database with a capital D.

lippserd (changes requested):

pkg/config/database.go
Al2Klimov: For finding usages?

lippserd: And for just a single place to change code whenever we need to change something.

lippserd (approved):

julianbrost:
Also, I'm missing the ability to configure a connection using Unix Domain Sockets. But I'll probably create an issue for this afterwards as I'd rather have this merged instead of blocking it because of this.

doc/02-Installation.md
```diff
@@ -133,7 +133,11 @@ Debian/Ubuntu:
 apt-get install icingadb-redis
 ```
 
-### Setting up the MySQL database <a id="setting-up-mysql-db"></a>
+### Setting up the Database <a id="setting-up-db"></a>
+
+Either a MySQL/MariaDB database or a PostgreSQL one is required.
```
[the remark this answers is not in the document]
Al2Klimov: I find your strange finding strange.

lippserd:
> Just a note to my future self or anyone else coming back to this PR asking who approved this: I still don't think these `boolenum` and `*uint` types in the schema are a good idea but I was outvoted there.

Just for the records, what would you propose?

julianbrost:
The standard `boolean` type and for integers just the smallest one that fits the required range and maybe some constraints as they make sense (as defined by the constraints given by icinga2, not by the MySQL schema).

But anyways, I didn't want to resume that discussion, GitHub just doesn't allow to get rid of the "changes requested" state without explicit approval.

PR #226, opened 2020-10-27
Keep testing w/ MariaDB v10.4

Al2Klimov:
... after v10.5 was released.


julianbrost (approved):
But keep in mind that a matrix can't generate more than 256 jobs. ;)

PR #249, opened 2021-02-02
Add cleanup package for housekeeping history tables

### LINKED ISSUE #163: History cleanup
Icingadb should have options to specify max age for notifications/history similar to the 'DB IDO Cleanup' feature to properly cleanup old entries in the following Icingadb database tables:

history
state_history
notification_history
user_notification_history

raviks789:
History tables in icingadb must be cleaned up every some period mentioned in housekeeping.ini file,
and all the records before specific time duration must be removed.

fixes #163


lippserd (changes requested):
Please rebase.

pkg/config/config.go
```diff
@@ -9,8 +9,9 @@ import (
 
 // Config defines Icinga DB config.
 type Config struct {
-	Database Database `yaml:"database"`
-	Redis    Redis    `yaml:"redis"`
+	Database *Database `yaml:"database"`
```
lippserd, line 12: Don't use pointers.

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,235 @@
...
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/icinga/icingadb/pkg/icingadb"
+	"github.com/icinga/icingadb/pkg/utils"
+	"go.uber.org/zap"
+	"sync"
+	"time"
+)
+
+// Tables type defines mapping of history tables with corresponding period.
+type Tables map[string]string
```
lippserd, line 16: Rename to `HistoryRetention` and use map[string]uint. Uint for days. time.Duration does not really make sense. Make sure to adjust all other code and docs as well.

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+
+		c.wg.Wait()
+	}
+
+	return nil
+}
+
+// controller calls the cleanup method periodically after c.opts.Interval.
+func (c *Cleanup) controller(ctx context.Context, tbl tableconfig, cufunc cleanupFunc) {
```
lippserd, line 91: This function should return any errors instead of logging them. Also you should reset the timer. A generic function for that could look like this:

```
func StablePeriodic(ctx context.Context, tick time.Duration, callback func(ctx context.Context, time time.Time) error) (context.CancelFunc, <-chan error) {
	ctx, cancelCtx := context.WithCancel(ctx)
	g, ctx := errgroup.WithContext(ctx)
	ticker := time.NewTicker(tick)

	g.Go(func() error {
		defer ticker.Stop()

		for {
			select {
			case t := <-ticker.C:
				if err := callback(ctx, t); err != nil {
					return err
				}

				ticker.Reset(tick)
			case <-ctx.Done():
				return ctx.Err()
			}
		}
	})

	return cancelCtx, com.WaitAsync(g)
}

**Review by lippserd — CHANGES_REQUESTED** [review]
Also, please rebase.

File: doc/03-Configuration.md [hunk 4]
```diff
@@ -35,4 +61,15 @@ database:
   password: icingadb
 redis:
   address: redis:6380
+cleanup:
+  history:
+    acknowledgement: 10
```
julianbrost, line 66: Suggestion: I think the configuration would be way more self-documenting if this was structured more like the following:
```yaml
history:
  retention-days:
    acknowledgement: 10
```

Also, it is nice to be able to have this fine-grained configuration, but I think for most users it would be sufficient to have a single "retain 100 days of history" option. So maybe have one option that's used as a default for all these detailed options.

doc/04-Housekeeping.md
```diff
@@ -0,0 +1,19 @@
+# Icinga DB Housekeeping
```
julianbrost, line 1: Should not have the "Icinga DB" prefix. Also, maybe something like "history cleanup" or "history retention" would be more descriptive.

doc/04-Housekeeping.md
```diff
@@ -0,0 +1,19 @@
...
+
+![Icinga Logo](https://icinga.com/wp-content/uploads/2014/06/icinga_logo.png)
+
+#### Table of Contents
+
+- [About](#about)
```
julianbrost, line 7: Other files in `doc/` don't have this at the top.

julianbrost (changes requested):
Needs a rebase. As quite a few changes happened in the meantime, I expect this to add quite some changes here, so I didn't have a too close look at the code so far.

julianbrost:
Superseded by #435 (which reuses some of the code from this PR), so I don't think there's a reason for keeping this one open.

PR #251, opened 2021-02-02
Config Sync: Iterate over config using cursor (replace HKEYS with HSCAN)

julianbrost:
Redis documentation states that it is preferable to use `*SCAN` instead of `*KEYS` (https://redis.io/commands/keys):

<blockquote>
Warning: consider KEYS as a command that should only be used in production environments with extreme care. It may ruin performance when it is executed against large databases. This command is intended for debugging and special operations, such as changing your keyspace layout.  Don't use KEYS in your regular application code. If you're looking for a way to find keys in a subset of your keyspace, consider using SCAN or sets.
</blockquote>

While the documentation is for the KEYS command and not HKEYS, the same warning should apply here to as the hashes we are iterating over can get quite large. Therefore use the HSCAN command that exists as an alternative.

In contrast to HKEYS, HSCAN not only returns the keys but also the values in the hash. Therefore now HSCAN is done on `icings:checksum:*` if available and on `icinga:config:*` otherwise. In both cases, the values are later reused.

## To be determined
* COUNT value for HSCAN: 1000? 2000? 10000? see https://github.com/Icinga/icingadb/pull/251#issuecomment-772458835

### DIFF
--- a/configobject/configsync/configsync.go
+++ b/configobject/configsync/configsync.go
@@ -36,19 +36,19 @@ func Operator(super *supervisor.Supervisor, chHA chan int, objectInformation *co
 		done chan struct{}
 		// Used by this Operator to provide the InsertPrepWorker with IDs to insert
 		// Operator -> InsertPrepWorker
-		chInsert chan []string
+		chInsert chan *connection.ConfigChunk
 		// Used by the JsonDecodePool to provide the InsertExecWorker with decoded rows, ready to be inserted
 		// JsonDecodePool -> InsertExecWorker
 		chInsertBack chan []connection.Row
 		// Used by this Operator to provide the DeleteExecWorker with IDs to delete
 		// Operator -> DeleteExecWorker
-		chDelete chan []string
+		chDelete chan *connection.ConfigChunk
 		// Used by this Operator to provide the UpdateCompWorker with IDs to compare
 		// Operator -> UpdateCompWorker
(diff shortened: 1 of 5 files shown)


julianbrost:
Did some basic evaluation on a fairly large config (~1 million service customwars). According to Redis `SLOWLOG GET`, a `icinga:config:service:customvar` takes about 350ms to 384ms on my machine. So this is the time for which the execution of other Redis queries is blocked.

Something else about `HSCAN`: there you are way more sensitive to network latency as each time you advance the cursor, you have to perform a full round-trip (you have to pass a cursor value that you receive from the previous call). This can probably be tuned with the `COUNT` parameter (which I for now set to the arbitrary value of 1000). According to the [documentation](https://redis.io/commands/scan#number-of-elements-returned-at-every-scan-call), this is just a hint to the implementation, but in my test with this large hash, it seems to respect the value. So this is something, were some fine-tuning can be done.

julianbrost:
Regarding the COUNT paramater of HSCAN: 2000 seems to be a good value if we want to stay below 10ms per query (that's what Redis SLOWLOG considers a slow query by default), but it would also be reasonable to go up to about 10000. These queries take 16ms to 35ms. All numbers are from my machine, so they may differ in detail, but should give the right order of magnitude.

N-o-X (approved):

Issue #209, opened 2020-08-28
icingadb mysql FATA[0001] Error 1064


mb-77:
I get the follow error after a while icingadb is running:

INFO[0000] Connecting to Redis
INFO[0000] Connecting to MySQL
INFO[0000] Redis connection established
INFO[0000] Starting heartbeat listener
INFO[0000] SQL connection established
INFO[0000] Got initial environment.                      UUID=04bafde1-3d3a-4ee0-9a90-ac17affacdf3 context=HA environment=da39a3ee5e6b4b0d3255bfef95601890afd80709
INFO[0000] Taking over.                                  UUID=04bafde1-3d3a-4ee0-9a90-ac17affacdf3 context=HA environment=da39a3ee5e6b4b0d3255bfef95601890afd80709
INFO[0001] Inserted 2 customvar_flats in 1.856756ms      action=insert benchmark=1.856756ms count=2 type=customvar_flat
**FATA[0001] Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '' at line 1**

I've try to drop and re-create the DB and its schema, but the problem persist

Obihoernchen:
I experience the same issue.

ziirish:
Same error.

```
Icinga DB version: v1.0.0-rc1
icinga2 - The Icinga 2 network monitoring daemon (version: r2.12.3-1)

ii  icingadb                             1.0.0~rc1-2.buster           amd64        database connector for Icinga 2
ii  icingadb-redis                       1.0.0~rc1-2.buster           amd64        Redis service for Icinga DB
ii  mariadb-common                   1:10.4.17+maria~buster      all          MariaDB database common files (e.g. /etc/mysql/conf.d/mariadb.cnf)
```

ziirish:
With the debug logs:

``` 
Jan 07 10:05:48 master-eu-1-vision icingadb[27092]: time="2021-01-07T10:05:48Z" level=debug msg="notes_url: Insert execution worker started"
Jan 07 10:05:48 master-eu-1-vision icingadb[27092]: time="2021-01-07T10:05:48Z" level=debug msg="notes_url: Delete execution worker started"
Jan 07 10:05:48 master-eu-1-vision icingadb[27092]: time="2021-01-07T10:05:48Z" level=debug msg="notes_url: Update comparison worker started"
[...]

PR #255, opened 2021-02-09
ChunkRows(): ensure to never get an empty chunk

Al2Klimov:
refs #209

### DIFF
--- a/connection/mysql_utils_test.go
+++ b/connection/mysql_utils_test.go
@@ -175,4 +175,15 @@ func TestChunkRows(t *testing.T) {
 	}
 	chunks = ChunkRows(rows, 2)
 	assert.Equal(t, want, chunks)
+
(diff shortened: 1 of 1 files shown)


Al2Klimov:
## Analysis

[The broken query](https://github.com/Icinga/icingadb/issues/209#issuecomment-769191274) may only originate from this statement:

https://github.com/Icinga/icingadb/blob/6dd577ea89bda8b326a722baa470189869bc0161/configobject/objecttypes/customvar/customvar.go#L76

... which is fired only here:

https://github.com/Icinga/icingadb/blob/6dd577ea89bda8b326a722baa470189869bc0161/connection/mysql.go#L708-L732

The only thing I can think of is an empty chunk.

N-o-X (approved):

PR #256, opened 2021-02-09
ChunkRows: replace float with int arithmetic

julianbrost:
Avoiding float arithmetic avoids precision problems. I don't think this results in any bugs for values we realistically reach in Icinga DB, but why use floats if you don't have to? Then you also don't have to think about whether doing so may blow up.

### DIFF
--- a/connection/mysql_utils.go
+++ b/connection/mysql_utils.go
@@ -10,7 +10,6 @@ import (
 	"github.com/go-sql-driver/mysql"
 	"io/ioutil"
 	oldlog "log"
-	"math"
 	"sort"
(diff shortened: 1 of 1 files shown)


Al2Klimov (approved):
Great, even the eventual edge case #255 works as expected:

```
Python 3.9.0 (default, Dec  6 2020, 18:02:34)
[Clang 12.0.0 (clang-1200.0.32.27)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 0 // 1
0
>>> 0 % 1
0
>>>
```

PR #259, opened 2021-02-15
Add foreign key with cascade delete constraints to history tables

raviks789:
Earlier we did not have any foreign keys for history table. But when we delete a record from the parent
history table the corresponding records in their child tables must also be deleted. This is done with the
introduction of foreign key constraints with on cascade delete.

refs #163


lippserd (changes requested):
Please rename the foreign keys according to the following scheme: `fk_$table_$referenced-table` and also add the keys to the upgrade script.

Al2Klimov:
I'm afraid this would make the history sync much more complicated.

lippserd:
> I'm afraid this would make the history sync much more complicated.

And why exactly?

Al2Klimov:
All of our Redis->DB syncs depend on the ability to run out-of-order. I.e. 1. service, 2. its host. Same for history.

I suggest to instead cleanup the tables independently from each other by timestamp.

lippserd:
Got your point now. But I think it makes sense to keep the foreign keys for referential integrity. Otherwise, we would have to select IDs before removing them from the child and parent tables. The foreign key columns are NULLable, so we have no problem inserting them out of order. What would only happen if they were separated because of the chunk size. Which we could also fix. Given that we are not going to clean up these events in the next fraction of a second, I think we can neglect that too.

lippserd:
@cla-bot check

Al2Klimov:
> Otherwise, we would have to select IDs before removing them from the child and parent tables.

Why exactly?

> The foreign key columns are NULLable, so we have no problem inserting them out of order.

But you'd have to insert NULL in the referring column for the case the referred row isn’t inserted yet, right?

lippserd:
> > Otherwise, we would have to select IDs before removing them from the child and parent tables.
> 
> Why exactly?

Because of referential integrity. `history` is the leading table in Web. You don't want to delete by timestamp which may delete different IDs.

> But you'd have to insert NULL in the referring column for the case the referred row isn’t inserted yet, right?

I think you're right and the out of order won't work. So we would have to fix this in our code or select the IDs first.

Al2Klimov:
> delete by timestamp which may delete different IDs

Depending on how exactly you delete or do I oversee something?

lippserd:
> > delete by timestamp which may delete different IDs
> 
> Depending on how exactly you delete or do I oversee something?

It cannot be guaranteed that two different delete statements will be executed exactly at the same time and therefore affect exactly the same IDs.

Al2Klimov:
It can be guaranteed that two different delete statements will be executed exactly at the same time by putting them into one transaction.

Even w/o the above it can be guaranteed that two different delete statements will affect exactly the same IDs by deleting an obvious past time period.

lippserd:
> Even w/o the above it can be guaranteed that two different delete statements will affect exactly the same IDs by deleting an obvious past time period.

That was a bit poorly worded from my side. I meant that there is no guarantee that the separate deletes by timestamp remove the correct records, e.g. if we only want to remove downtimes that have actually ended, we can only do this reliably if we select the IDs from the `downtime_history` table beforehand. Of course, we could also do a delete with a join from the history table, but I don't understand why we want to introduce such complexity. On the other hand, without referential integrity in the first place, there is a chance that bugs in Icinga DB or Icinga 2 could result in a differing state of the history tables. And with the delete by timestamp we may delete records that do not belong to each other. So, I really think it's best to introduce referential integrity for the history tables.

Al2Klimov:
> why we want to introduce such complexity

Because IMAO not having the ability to insert out of order is more complex.

lippserd:
> Because IMAO not having the ability to insert out of order is more complex.

I think it's better to add "complexity" in one place than to shift the complexity to each individual consumer.
Not only the cleanup but also reporting [1] and other history consumers require an integral state of the history tables:

> On the other hand, without referential integrity in the first place, there is a chance that bugs in Icinga DB or Icinga 2 could result in a differing state of the history tables.

We are not that far away from referential integrity here. We are already inserting first into the child table and then into the history table. We just need to make sure that these statements are not separated by the chunk size.

I don't know if I am alone with my opinion here, but I think integrity is very important. All other tables can be (re)synchronized after restarting Icinga DB. The history tables just don't. /cc @julianbrost @N-o-X 

[1] This is reporting in general, not our SLA, as these are separate tables. I'm thinking of event timelines in the history and detail views and other things we don't know yet.

Al2Klimov:
Huh? Didn’t we agree on using JOINs (whereever) and (yet) missing data across tables... just vanishes from the result set?

julianbrost:
> All of our Redis->DB syncs depend on the ability to run out-of-order. I.e. 1. service, 2. its host. Same for history.

I don't think this is a good comparison. Services and hosts are different objects in Redis, so f we'd add a foreign key constraint for `service.host_id`, we'd have to enforce some ordering between these two object types. However, with history, there's one Redis stream item for each event and Icinga DB splits this into two rows for different tables. I haven't checked the code but right now, it probably just produces rows for two different bulk inserters and this would have to be changed to happen in the right order, but I don't think the amount of work needed to achieve that is anywhere near what it would take to do the same for hosts and services.

> The foreign key columns are NULLable, so we have no problem inserting them out of order.

If you consider this a valid option, applications would have to be able to handle NULL foreign keys. Couldn't they then just handle dangling foreign keys the same way?

Anyways, let me try to understand the intentions of this PR first. If I understand correctly, you want to use the following queries for history cleanup and want these to implicitly (by the means of `ON DELETE CASCADE`) clean the main `history` timeline table:

https://github.com/Icinga/icingadb/blob/88dffc3fdca723c164931ad0a720ece6602b3e04/pkg/cleanup/cleanup.go#L48-L53

So from my point of view (i.e. coming to the project at a later stage, notably after most design decision), there are intentionally no foreign key constraints at all in the schema. Changing this is possible of course but should be well considered.

I think the best argument for changing this is the following (the very specific use-case of history cleanup could probably also be handled good enough without foreign key constraints):
> On the other hand, without referential integrity in the first place, there is a chance that bugs in Icinga DB or Icinga 2 could result in a differing state of the history tables.

If the following indeed describes all changes required, it might be worth doing it:
> We are not that far away from referential integrity here. We are already inserting first into the child table and then into the history table. We just need to make sure that these statements are not separated by the chunk size.

As I'm going to look more into the history sync code due to SLA Reporting soon anyways, I'll try to keep this issue here in mind and can hopefully come up with a more informed opinion then.

lippserd:
> > The foreign key columns are NULLable, so we have no problem inserting them out of order.
> 
> If you consider this a valid option, applications would have to be able to handle NULL foreign keys. Couldn't they then just handle dangling foreign keys the same way?

That was wrong. If you try to insert a row in the `history` table with a non-NULL foreign key, it has to exist in the referenced table.

> So from my point of view (i.e. coming to the project at a later stage, notably after most design decision), there are intentionally no foreign key constraints at all in the schema. Changing this is possible of course but should be well considered.

We don't have constraints and auto-increment IDs because we want to maximize parallel inserts.
We actually introduced this for configuration and status data over which Icinga has control.
With the history tables, however, Icinga DB has control, because it is Icinga DB that decides to insert into several tables and the data cannot simply be re-synchronized after a restart. What is in the database remains in the database.
Without referential integrity, there is no guarantee to consumers that the history tables will be consistent.
When the tables differ, it would be very difficult for us to understand why they are not consistent.
It becomes even more difficult if the cleanup then simply deletes this data by using just a timestamp.

schema/mysql/schema.sql
```diff
@@ -909,7 +909,9 @@ CREATE TABLE user_notification_history (
   notification_history_id binary(16) NOT NULL COMMENT 'UUID notification_history.id',
   user_id binary(20) NOT NULL COMMENT 'user.id',
 
-  PRIMARY KEY (id)
+  PRIMARY KEY (id),
+
+  CONSTRAINT fk_user_notification_history_notification_history FOREIGN KEY (notification_history_id) REFERENCES notification_history (id) ON DELETE CASCADE
```
julianbrost, line 914: This constraint actually adds a dependency between the `icinga:history:stream:notification` and `icinga:history:stream:usernotification` streams. Thus, what I said in https://github.com/Icinga/icingadb/pull/259#issuecomment-895957599 is wrong and to support this schema change, we'd indeed have to change our code to wait for entries from other streams.

julianbrost (changes requested):
I'm explicitly setting this to "changes requested" as this PR can't be merged anyways until the code is updated so that history events are inserted in the proper order to actually satisfy these constraints.

(Or we drop this idea altogether as this turned out to require more extensive changes as previously assumed (https://github.com/Icinga/icingadb/pull/259#pullrequestreview-736126161))

Al2Klimov:
I opt for the latter.

PR #261, opened 2021-02-19
Set an explicit timeout for go test

julianbrost:
Our GitHub Actions run with a timeout of 10 minutes which is also the default timeout of `go test`. Therefore if a test hangs, we will never reach the timeout of `go test` which would then panic and print the stack of all goroutines. This PR reduces the timeout of `go test` to 5 minutes so that we get this useful information.


julianbrost:
Actually, the failed CI is the proof that it works as expected: https://github.com/Icinga/icingadb/pull/261/checks?check_run_id=1934329729

N-o-X (approved):

Issue #263, opened 2021-02-23
Panic in PipeConfigChunks


julianbrost:
Looks like my PR #251 introduced a bug which can lead to a panic:

```
panic: runtime error: slice bounds out of range [:3] with capacity 1

goroutine 2886 [running]:
github.com/Icinga/icingadb/connection.(*RDBWrapper).PipeConfigChunks.func3(0xc0003b1860, 0xc00006f630, 0x1f4, 0xc0007733e0)
	/idbcp/connection/redis.go:661 +0x33a
created by github.com/Icinga/icingadb/connection.(*RDBWrapper).PipeConfigChunks
	/idbcp/connection/redis.go:653 +0x145
```

N-o-X:
Shouldn't be a problem anymore with the RC2 changes in place.

PR #264, opened 2021-02-23
PipeConfigChunks(): don't crash on incomplete chunks

Al2Klimov:


Al2Klimov:
## Scenario

1. Dump a large config
2. Clear conf.d
3. Reload

## Before

panic: runtime error: slice bounds out of range [:197] with capacity 1

## After

julianbrost:
The idea when I wrote this code was that all slices have the same length or are nil. I'd prefer to check where the lengths diverge rather than fix it this way. I can also take a look at this.

Also fixes #263 I created today (noticed the same today).

Al2Klimov:
Yes, please have a look!

N-o-X:
This PR isn't needed anymore after our RC2 rework.

PR #289, opened 2021-05-31
Error handling

lippserd:


internal/internal.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// CantParseUint64 wraps the given error with the specified string that cannot be parsed into uint64.
+func CantParseUint64(err error, s string) error {
+	return errors.Wrapf(err, "can't parse '%s' into uint64", s)
+}
+
+// CantPerformQuery wraps the given error with the specified query that cannot be executed.
+func CantPerformQuery(err error, q string) error {
+	return errors.Wrap(err, "can't perform "+q)
```
julianbrost, line 25: This could use a `%q` format string to get a nice quoted query.

internal/internal.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// CantUnmarshalJSON wraps the given error with the specified data that cannot be unmarshalled into the designated value.
+func CantUnmarshalJSON(err error, data []byte, v interface{}) error {
+	return errors.Wrapf(err, "can't unmarshal JSON '%s' (%#v) into %T", data, data, v)
```
julianbrost, line 30: `%q` instead of `'%s'`. Also, why is `data` logged twice?

pkg/icingadb/overdue/sync.go
```diff
@@ -246,8 +250,10 @@ func (s Sync) updateOverdue(
 		op = s.redis.SRem
 	}
 
-	_, err := op(ctx, "icingadb:overdue:"+objectType, ids...).Result()
-	return err
+	cmd := op(ctx, "icingadb:overdue:"+objectType, ids...)
+	_, _ = cmd.Result() // Error is handled by icingaredis.WrapCmdErr().
```
julianbrost, line 254: Maybe have a function that both calls `.Result()` and performs the error wrapping? Would avoid having to explain why there's no error handling here.

pkg/icingaredis/v1/stats_message.go
```diff
@@ -24,25 +25,25 @@ func (m StatsMessage) IcingaStatus() (*IcingaStatus, error) {
...
 		if err := json.Unmarshal([]byte(s), &envelope); err != nil {
-			return nil, errors.Wrap(err, "can't parse Icinga 2 status")
+			return nil, internal.CantUnmarshalJSON(err, []byte(s), envelope)
 		}
 
 		return &envelope.Status.IcingaApplication.IcingaStatus, nil
 	}
 
-	return nil, errors.New("bad message")
+	return nil, errors.Errorf("bad message %#v. %s missing", m, "IcingaApplication")
```
julianbrost, line 34: Why not just inline the hardcoded string constant parameter into the format string?

pkg/icingaredis/v1/stats_message.go
```diff
@@ -24,25 +25,25 @@ func (m StatsMessage) IcingaStatus() (*IcingaStatus, error) {
...
 		if err := json.Unmarshal([]byte(s), &t); err != nil {
-			return nil, errors.Wrap(err, "can't parse timestamp")
+			return nil, internal.CantUnmarshalJSON(err, []byte(s), t)
 		}
 
 		return &t, nil
 	}
 
-	return nil, errors.New("bad message")
+	return nil, errors.Errorf("bad message %#v. %s missing", m, "timestamp")
```
julianbrost, line 48: Same here.

pkg/types/bool.go
```diff
@@ -40,14 +41,14 @@ func (b Bool) MarshalJSON() ([]byte, error) {
...
-	return json.Marshal(b.Bool)
+	return internal.MarshalJSON(b.Bool)
 }
 
 // UnmarshalText implements the encoding.TextUnmarshaler interface.
 func (b *Bool) UnmarshalText(text []byte) error {
 	parsed, err := strconv.ParseUint(string(text), 10, 64)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "can't parse '%s' into uint64", text)
```
julianbrost, line 51: `%q` instead of `'%s'`

pkg/types/comment_type.go
```diff
@@ -33,7 +34,7 @@ func (ct *CommentType) UnmarshalText(bytes []byte) error {
 
 	i, err := strconv.ParseUint(text, 10, 64)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "can't parse '%s' into uint64", text)
```
julianbrost, line 37: `%q` instead of `'%s'`

pkg/types/float.go
```diff
@@ -22,14 +24,14 @@ func (f Float) MarshalJSON() ([]byte, error) {
...
-	return json.Marshal(v)
+	return internal.MarshalJSON(v)
 }
 
 // UnmarshalText implements the encoding.TextUnmarshaler interface.
 func (f *Float) UnmarshalText(text []byte) error {
 	parsed, err := strconv.ParseFloat(string(text), 64)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "can't parse '%s' into float64", text)
```
julianbrost, line 34: `%q` instead of `'%s'`

julianbrost (commented):
Just what I've noticed so far from going over the code once.

pkg/types/binary.go
```diff
@@ -52,7 +53,7 @@ func (binary *Binary) UnmarshalText(text []byte) error {
 	b := make([]byte, hex.DecodedLen(len(text)))
 	_, err := hex.Decode(b, text)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "can't decode hex %q into %T", text, b)
```
julianbrost, line 56: `%T` of `b` will always be `[]byte`, so I think there's little value in adding this to the message.

internal/internal.go
```diff
@@ -0,0 +1,46 @@
...
+	return errors.Wrapf(err, "can't unmarshal JSON %q into %T", data, v)
+}
+
+// CantUnmarshalYAML wraps the given error with the specified data that cannot be unmarshalled.
+func CantUnmarshalYAML(err error, data interface{}) error {
+	return errors.Wrapf(err, "can't unmarshal YAML %#v", data)
+}
+
+// MarshalJSON calls json.Marshal and wraps any resulting errors.
+func MarshalJSON(v interface{}) ([]byte, error) {
```
julianbrost, line 39: Why this but no `UnmarshalJSON()`?

julianbrost (changes requested):

internal/internal.go
```diff
@@ -0,0 +1,46 @@
...
+	return errors.Wrapf(err, "can't perform %q", q)
+}
+
+// CantUnmarshalJSON wraps the given error with the specified data that cannot be unmarshalled into the designated value.
+func CantUnmarshalJSON(err error, data []byte, v interface{}) error {
+	return errors.Wrapf(err, "can't unmarshal JSON %q into %T", data, v)
+}
+
+// CantUnmarshalYAML wraps the given error with the specified data that cannot be unmarshalled.
+func CantUnmarshalYAML(err error, data interface{}) error {
```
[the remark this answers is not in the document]
lippserd: That makes sense. I don't think there is any other way to access the raw YAML other than adding it to the error in `FromYAMLFile`. But the error message contains the path to the file, which should be enough:

> can't parse YAML file ./config.yml: can't unmarshal YAML into *config.Database: yaml: unmarshal errors:
  line 3: mapping key "host" already defined at line 2

Also, I don't think it's a good idea to post passwords in logs.

pkg/config/redis.go
```diff
@@ -69,19 +70,24 @@ func dialWithLogging(logger *zap.SugaredLogger) func(context.Context, string, st
 			backoff.NewExponentialWithJitter(1*time.Millisecond, 1*time.Second),
 			5*time.Minute,
 		)
+
+		if err != nil {
+			err = errors.Wrap(err, "can't connect to Redis")
+		}
```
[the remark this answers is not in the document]
lippserd: Nice, good catch! I will address all the places where we can unconditionally call `errors.Wrap *`.

pkg/icingaredis/error_handling.go
```diff
@@ -0,0 +1,57 @@
+package icingaredis
+
+import (
+	"context"
+	"github.com/go-redis/redis/v8"
+	"github.com/icinga/icingadb/pkg/utils"
+	"github.com/pkg/errors"
+)
+
+type ErrorHandling struct{}
```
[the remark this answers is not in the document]
lippserd: `WrapErrors` sounds good to me.

internal/internal.go
lippserd: Good question 😆 I'll add it.

internal/internal.go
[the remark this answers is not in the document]
julianbrost: > Also, I don't think it's a good idea to post passwords in logs

Good point, I just noticed that it's inconsistent with the functions for JSON. The JSON could also contain sensitive information though (mainly custom vars, like SNMP credentials for example).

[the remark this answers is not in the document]
lippserd: I also removed it JSON contents from the error.

julianbrost (approved):

PR #295, opened 2021-06-01
Assert the database schema of the expected version being present

Al2Klimov:


Al2Klimov:
## Before

```
2021-06-01T17:34:28.640+0200    INFO    icingadb/main.go:51     Starting Icinga DB
2021-06-01T17:34:28.640+0200    INFO    icingadb/main.go:56     Connecting to database
2021-06-01T17:34:28.651+0200    INFO    icingadb/main.go:65     Connecting to Redis
2021-06-01T17:34:28.657+0200    DEBUG   icingadb/ha.go:105      Starting HA     {"instance_id": "1fc6196cc62f4baf9cda049e5726bac1"}
```

## After

```
database schema is absent
```

```
expected database schema v2, got v1
```

```
2021-06-01T17:25:33.931+0200    INFO    icingadb/main.go:52     Starting Icinga DB
2021-06-01T17:25:33.931+0200    INFO    icingadb/main.go:57     Connecting to database
2021-06-01T17:25:33.964+0200    INFO    icingadb/main.go:71     Connecting to Redis
```

cmd/icingadb/main.go
```diff
@@ -60,6 +61,10 @@ func run() int {
 		}
 	}
 
+	if err := checkDbSchema(context.Background(), db, cmd.Config.Database.Database); err != nil {
+		panic(err)
```
lippserd, line 65: Please use `logger.Fatalf("%+v", err)`.

cmd/icingadb/main.go
```diff
@@ -253,3 +258,42 @@ func run() int {
 		cancelHactx()
 	}
 }
+
+// checkDbSchema asserts the database schema of the expected version being present.
+func checkDbSchema(ctx context.Context, db *icingadb.DB, dbName string) error {
+	{
+		var one []struct{ One uint8 }
```
lippserd, line 265: I wouldn't check if the schema is installed. The `SELECT` will fail anyway if there is no table` icingadb_schema`.

cmd/icingadb/main.go
```diff
@@ -253,3 +258,42 @@ func run() int {
...
+
+		if len(one) < 1 {
+			return errors.New("database schema is absent")
+		}
+	}
+
+	var version []struct{ Version uint16 }
+
+	err := db.SelectContext(
+		ctx, &version, "SELECT version FROM icingadb_schema ORDER BY `timestamp` DESC LIMIT 1",
```
lippserd, line 284: Please order by `id`. I wouldn't expect `timestamp` to be correct 100 percent of the time.

cmd/icingadb/main.go
```diff
@@ -253,3 +258,42 @@ func run() int {
...
+	)
+	if err != nil {
+		return errors.Wrap(err, "can't check database schema version")
+	}
+
+	if len(version) < 1 {
```
lippserd, line 290: I don't think that's necessary. `version` could be`0`, but that is handled with the next `if` anyway.

cmd/icingadb/main.go
```diff
@@ -253,3 +258,42 @@ func run() int {
...
+		}
+
+		if len(one) < 1 {
+			return errors.New("database schema is absent")
+		}
+	}
+
+	var version []struct{ Version uint16 }
+
+	err := db.SelectContext(
```
lippserd, line 283: Could be the following instead:
```
var version uint16
err := db.QueryRowxContext(ctx, `SELECT...`).Scan(&version)

lippserd (changes requested):

cmd/icingadb/main.go
Al2Klimov: Hm.. clever **and** smart :)

cmd/icingadb/main.go
Al2Klimov: With only the next if you'd get a SEGV.

lippserd (approved):

PR #300, opened 2021-06-18
Driver context

lippserd:


pkg/driver/driver.go
```diff
@@ -72,3 +108,32 @@ func Register(logger *zap.SugaredLogger) {
 	// TODO(el): Don't discard but hide?
 	_ = mysql.SetLogger(log.New(ioutil.Discard, "", 0))
 }
+
+// ctxConnector adds driver.DriverContext support for drivers that do not implement it.
+type ctxConnector struct {
+	driver driver.Driver
+	name   string
+}
```
julianbrost, line 116: Why is this needed? The MySQL driver seems to support `DriverContext`. At least the implementation of `Connect()` doesn't seem ideal as it doesn't stop any operations if the context is canceled but only checks when done. Thus, `context.WithTimeout(...)` won't really work as expected.

lippserd: The alternative would be to panic in `OpenConnector()` which is not an option for me as this code is library code. In the end, we're just wrapping here. I can't see how to change `Connect ()` any further as it's the driver that doesn't actually handle contexts.

lippserd: We could check the context before calling Open though.

julianbrost: I see two more options:
1. Just always use the `DriverContext` interface here as `github.com/go-sql-driver/mysql` implements it. So `ctxConnector` shouldn't be used anyways at the moment and this could be asserted at compile time. However, `github.com/lib/pq` currently does not support that interface. There's a PR for that (https://github.com/lib/pq/pull/900) but it didn't make any progress since May 2020.
2. Execute `Open()` in a Goroutine so that connect will return immediately in case the context is canceled and if this happens, let the Goroutine close and throwaway the connection after `Open()` returns.

Given that this should be dead code right now anyways, I wouldn't bother with it now and see if we want to implement (2) when we add PostgreSQL support at a later point if that PR hasn't made any progress until then.

julianbrost (commented):
Does the Go `sql` package guarantee to always use `DriverContext` if available? Or would it make sense to also explicitly implement the `Open()` function from the `Driver` interface (simply calling `Connect(context.Background())`) to ensure that `Open()` also retries?

lippserd:
> Does the Go `sql` package guarantee to always use `DriverContext` if available? Or would it make sense to also explicitly implement the `Open()` function from the `Driver` interface (simply calling `Connect(context.Background())`) to ensure that `Open()` also retries?

Yep, `DriverContext` will always be used if available:

https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/database/sql/driver/driver.go#L10
https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/database/sql/driver/driver.go#L97
https://github.com/golang/go/blob/10d625d5b41f17c118da18a592c683e60fcdcb3b/src/database/sql/sql.go#L824

julianbrost (approved):

PR #257, opened 2021-02-09
Schema: introduce downtime#creator

Al2Klimov:
fixes #231

PR #301, opened 2021-06-23
Introduce downtime#scheduled_by

### LINKED ISSUE #231: Downtime tables should include downtime source/origin
Just faced again the issue that [Icinga 2 does not complain about a failed downtime removal](https://github.com/Icinga/icingaweb2/issues/4247) if it's coming from a scheduled downtime config object.

Icinga DB Web could hide the removal button in such a case, though there doesn't seem to be a column in the schema indicating what created the downtime. I'd like to have that included.

Al2Klimov:
fixes #231
closes #257

### DIFF
--- a/schema/mysql/schema.sql
+++ b/schema/mysql/schema.sql
@@ -594,6 +594,7 @@ CREATE TABLE downtime (
   is_in_effect enum('n', 'y') NOT NULL,
   start_time bigint unsigned DEFAULT NULL COMMENT 'Time when the host went into a problem state during the downtimes timeframe',
   end_time bigint unsigned DEFAULT NULL COMMENT 'Problem state assumed: scheduled_end_time if fixed, start_time + flexible_duration otherwise',
+  scheduled_by varchar(767) DEFAULT NULL COMMENT 'Name of the ScheduledDowntime which created this Downtime. 255+1+255+1+255, i.e. "host.name!service.name!scheduled-downtime-name"',
 
   zone_id binary(20) DEFAULT NULL COMMENT 'zone.id',
 
@@ -958,6 +959,7 @@ CREATE TABLE downtime_history (
   scheduled_end_time bigint unsigned NOT NULL,
   start_time bigint unsigned NOT NULL COMMENT 'Time when the host went into a problem state during the downtimes timeframe',
   end_time bigint unsigned NOT NULL COMMENT 'Problem state assumed: scheduled_end_time if fixed, start_time + duration otherwise',
+  scheduled_by varchar(767) DEFAULT NULL COMMENT 'Name of the ScheduledDowntime which created this Downtime. 255+1+255+1+255, i.e. "host.name!service.name!scheduled-downtime-name"',
   has_been_cancelled enum('n', 'y') NOT NULL,
   trigger_time bigint unsigned NOT NULL,
   cancel_time bigint unsigned DEFAULT NULL,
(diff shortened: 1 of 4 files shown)


Al2Klimov:
## Blocked by

* [ ] Icinga/icinga2#8631

schema/mysql/schema.sql
```diff
@@ -588,6 +588,7 @@ CREATE TABLE downtime (
   is_in_effect enum('n', 'y') NOT NULL,
   start_time bigint unsigned DEFAULT NULL COMMENT 'Time when the host went into a problem state during the downtimes timeframe',
   end_time bigint unsigned DEFAULT NULL COMMENT 'Problem state assumed: scheduled_end_time if fixed, start_time + flexible_duration otherwise',
+  scheduled_by varchar(255) DEFAULT NULL,
```
julianbrost, line 591: 255 is probably not long enough. The value will be "host_name!service_name!downtime_name", so probably something like 3*255 + 2 is required.

julianbrost (changes requested):

schema/mysql/schema.sql
lippserd: And there should be a `COMMENT` that explains that and what the column reflects.

Al2Klimov: @lippserd Can you confirm that length?

lippserd: We have no limit on host and service names other than the implicit file name length limit for API-created objects, right?
I think this should be a text column just to be on the safe side. MySQL will not execute the statement if values need to be truncated and it is running in strict mode. Which in turn would stop Icinga DB.

julianbrost: `host.name` and `service.name` are both `varchar(255)`. Assuming the same length for the names of ScheduledDowntimes, that yields the limit I suggested above. Haven't checked if Icinga 2 enforces these limits, but if it doesn't, maybe it should if we have database schemas with those limits?

lippserd: If the downtime name can't be longer than 255 chars 3*255+2 is fine. For host and service names we already have the 255 char length restriction in other tables. But yes, Icinga 2 should enforce those limits.

Al2Klimov: I.e. Icinga 2 shall reject config of objects w/ names >255?

Al2Klimov:
## TODO

* [ ] Reason column length

lippserd (changes requested):
I would rewrite the comment to

> Name of the ScheduledDowntime which created this downtime. 255+1+255+1+255, i.e. "host.name!service.name!scheduled-downtime-name".

lippserd (approved):

lippserd (changes requested):
Please also add it to the history.

Al2Klimov:
## TODO

* https://github.com/Icinga/icinga2/pull/8966

Al2Klimov:
## Purrs

```
MariaDB [idb]> select scheduled_by from downtime_history;
+---------------------------------------------------+
| scheduled_by                                      |
+---------------------------------------------------+
| Alexanders-MacBook-Pro.local!load!backup-downtime |
+---------------------------------------------------+
1 row in set (0.004 sec)

MariaDB [idb]>
```

julianbrost (approved):

bobapple (approved):

lippserd (approved):

PR #303, opened 2021-06-29
Fix SEGV due to empty config

Al2Klimov:


Al2Klimov:
Tested with the config `{}` which may be emitted by our entrypoint.

lippserd (changes requested):
Still does not work for `---`.

julianbrost (approved):

lippserd (approved):

PR #304, opened 2021-07-01
Activate and use bulk upsert

lippserd:
closes #298

### DIFF
--- a/pkg/icingadb/db.go
+++ b/pkg/icingadb/db.go
@@ -114,7 +114,7 @@ func (db *DB) BuildUpsertStmt(subject interface{}) (stmt string, placeholders in
 	set := make([]string, 0, len(updateColumns))
 
 	for _, col := range updateColumns {
-		set = append(set, fmt.Sprintf("%s = :%s", col, col))
+		set = append(set, fmt.Sprintf("%s = VALUES(%s)", col, col))
 	}
 
 	return fmt.Sprintf(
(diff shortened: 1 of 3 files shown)


pkg/icingadb/db.go
```diff
@@ -98,7 +98,7 @@ func (db *DB) BuildUpdateStmt(update interface{}) string {
 		`UPDATE %s SET %s WHERE id = :id`,
 		utils.TableName(update),
 		strings.Join(set, ", "),
-	)
+	), len(columns)
```
julianbrost, line 101: With `:id` in the `WHERE` clause, there is one more placeholder.

pkg/icingadb/db.go
```diff
@@ -371,7 +371,9 @@ func (db *DB) CreateStreamed(ctx context.Context, entities <-chan contracts.Enti
 	}
 
 	sem := db.getSemaphoreForTable(utils.TableName(first))
-	return db.NamedBulkExec(ctx, db.BuildInsertStmt(first), 1<<15/len(db.BuildColumns(first)), sem, forward, nil)
+	stmt, placeholders := db.BuildInsertStmt(first)
+
+	return db.NamedBulkExec(ctx, stmt, 1<<15/placeholders, sem, forward, nil)
```
julianbrost, line 376: `1<<15` could be made a named constant to better explain why this is done here (not directly related to this PR, but if you're touching the code anyways).

julianbrost (changes requested):

pkg/icingadb/db.go
lippserd: True!

pkg/icingadb/db.go
lippserd: Will be addressed in the code-quality PR.

julianbrost (approved):

PR #310, opened 2021-07-11
Fix/enhance logging

raviks789:
Enhance logging by providing configuration for logging in configuration file `config.yml`.


internal/logging/logging.go
```diff
@@ -0,0 +1,197 @@
...
+	case "systemd":
+		return getJsonEncoder(), getJournaldWriteSyncer()
+	default:
+		panic(fmt.Sprintf("%s is not a valid logger output", output))
+	}
+}
+
+// getConsoleEncoder returns console encoder used for console logging.
+func getConsoleEncoder() zapcore.Encoder {
+	return zapcore.NewConsoleEncoder(zap.NewDevelopmentEncoderConfig())
```
lippserd, line 123: Why `NewDevelopmentEncoderConfig`? You have a `defaultConfig`.

internal/logging/logging.go
```diff
@@ -0,0 +1,197 @@
...
+	return zapcore.Lock(os.Stderr)
+}
+
+// Fatal introduces a short buffer time between Fatal log and os.Exit(1) by sleeping
+// for a short interval after logging fatal error. This is done to avoid the loss of fatal level log messages
+// on exit on failure.
+func (l *Logging) Fatal(err error) {
+	l.enabled.Store(false)
+
+	// Fatal would exit after write, but we want to sleep first.
```
lippserd, line 183: Don't sleep if it's not `systemd-journal`.

internal/logging/logging.go
```diff
@@ -0,0 +1,197 @@
...
+		var childlvl zapcore.Level
+		if err := childlvl.Set(level); err != nil {
+			panic(err)
+		}
+		atom := zap.NewAtomicLevelAt(childlvl)
+
+		encoder, syncer := selectEncoderSyncer(l.output)
+
+		// child logger cloned from default logger and using childlvl for LevelEnablerFunc.
+		logger := l.logger.Desugar().With(zap.Skip()).WithOptions(
```
lippserd, line 84: You don't need `.With(zap.Skip())`.

internal/logging/journal.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	consoleEncEntry := zapcore.Entry{
+		Level:      stringToZapLevel[line.Level],
+		LoggerName: line.LoggerName,
+		Message:    line.Message,
+		Stack:      line.Stack,
+	}
+
+	message, err := zapcore.NewConsoleEncoder(defaultConfig.EncoderConfig).EncodeEntry(consoleEncEntry, nil)
```
lippserd, line 94: Make no sense to create an encoder for every message. Pass the encoder config to `NewJournalWriter` and create an encoder from that.

internal/logging/logging.go
```diff
@@ -0,0 +1,197 @@
...
+	Development: true,
+	Sampling: &zap.SamplingConfig{
+		Initial:    100,
+		Thereafter: 100,
+	},
+
+	Encoding: "json",
+
+	// copied from "zap.NewProductionEncoderConfig" with some updates
+	EncoderConfig: zapcore.EncoderConfig{
```
lippserd, line 138: Only save `EncoderConfig` for now and don't disable `TimeKey`. You can override that in `NewJournalWriter` later. Make sure to pass the struct not a pointer to the struct!

internal/logging/logging.go
lippserd: Also, you don't need this method as you're only calling it once.

internal/logging/logging.go
```diff
@@ -0,0 +1,197 @@
...
+		EncodeCaller:   zapcore.ShortCallerEncoder,
+	},
+
+	// Use "/dev/null" to discard all
+	OutputPaths:      []string{"stderr"},
+	ErrorOutputPaths: []string{"stderr"},
+}
+
+// getJsonEncoder returns json encoder used for systemd-journald logging.
+func getJsonEncoder() zapcore.Encoder {
```
lippserd, line 158: You don't need this method as you're only calling it once.

internal/logging/logging.go
```diff
@@ -0,0 +1,197 @@
...
+	return zapcore.NewJSONEncoder(defaultConfig.EncoderConfig)
+}
+
+// getJournaldWriteSyncer returns WriteSyncer for systemd-journald logging.
+func getJournaldWriteSyncer() zapcore.WriteSyncer {
```
lippserd, line 163: Since it's `journal` everywhere, remove the `d`.

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+		l.loggers[name] = logger
+		return logger
+	} else {
+		logger := l.logger.Named(name)
+		l.loggers[name] = logger
+		return logger
+	}
+}
+
+// GetLogger returns the default sugared logger for the initialized Logging instance.
```
lippserd, line 107: GetLogger returns the default logger.

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	// if the logger is already stored then use the stored child logger.
+	if logger, ok := l.loggers[name]; ok {
+		return logger
+	}
+
+	if level, found := l.options[name]; found {
+		var childlvl zapcore.Level
```
lippserd, line 80: `lvl` instead of `childlvl`

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+		atom := zap.NewAtomicLevelAt(childlvl)
+
+		// child logger cloned from default logger and using childlvl for LevelEnablerFunc.
+		logger := l.logger.Desugar().WithOptions(
+			zap.WrapCore(func(c zapcore.Core) zapcore.Core {
+				return zapcore.NewCore(
+					l.encoder,
+					l.syncer,
+					zap.LevelEnablerFunc(func(childlvl zapcore.Level) bool {
```
lippserd, line 92: `lvl` instead of `childlvl`

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+						// Always log fatals, but allow other levels to be completely disabled.
+						return zap.FatalLevel.Enabled(childlvl) || l.enabled.Load() && atom.Enabled(childlvl)
+					}))
+			})).Sugar().Named(name)
+
+		l.loggers[name] = logger
+		return logger
```
lippserd, line 99: newline

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	} else {
```
lippserd, line 100: You don't need `else` here.

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+		logger := l.logger.Named(name)
+		l.loggers[name] = logger
+		return logger
```
lippserd, line 103: newline

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	encoder zapcore.Encoder
+	// encoder stores zapcore.WriteSyncer; used to build the logger
+	syncer zapcore.WriteSyncer
+	mu     sync.Mutex
+	// loggers stores named child loggers.
+	loggers map[string]*zap.SugaredLogger
+	options Options
+}
+
+// Options stores a `child-logger-name: level` pair mapping.
```
lippserd, line 34: Options define child loggers with their desired log level.

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	"sync"
+	"time"
+)
+
+// Logging implements logging using: a switch to disable logging on Fatal log, default logger and level,
+// and stored child loggers and their options.
+type Logging struct {
+	// enabled implements a switch to disable logging on Fatal log.
+	enabled *atomic.Bool
+	// level defines default log-level.
```
lippserd, line 18: I think you don't need a reference to level.

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	))
+
+	return &Logging{
+		enabled: enabled,
+		level:   atom,
+		output:  output,
+		logger:  logger.Sugar(),
+		encoder: encoder,
+		syncer:  syncer,
+		mu:      sync.Mutex{},
```
lippserd, line 63: You don't need that initialization. The zero value of `sync.Mutex` is already usable.

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	case "console":
+		return getConsoleEncoder(zap.NewDevelopmentEncoderConfig()), getConsoleWriteSyncer()
+	case "systemd-journal":
+		return zapcore.NewJSONEncoder(defaultEncConfig), getJournaldWriteSyncer()
+	default:
+		panic(fmt.Sprintf("%s is not a valid logger output", output))
+	}
+}
+
+// getConsoleEncoder returns console encoder used for console logging.
```
lippserd, line 124: You don't need that function. Just use `zapcore.NewConsoleEncoder()`.

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	// output defines default log-level.
+	output string
+	// logger defines default logger.
+	logger *zap.SugaredLogger
+	// encoder stores zapcore.Encoder; used to build the logger
+	encoder zapcore.Encoder
+	// encoder stores zapcore.WriteSyncer; used to build the logger
+	syncer zapcore.WriteSyncer
+	mu     sync.Mutex
+	// loggers stores named child loggers.
```
lippserd, line 29: remove comment

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+)
+
+// Logging implements logging using: a switch to disable logging on Fatal log, default logger and level,
+// and stored child loggers and their options.
+type Logging struct {
+	// enabled implements a switch to disable logging on Fatal log.
+	enabled *atomic.Bool
+	// level defines default log-level.
+	level zap.AtomicLevel
+	// output defines default log-level.
```
lippserd, line 20: remove comment

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	output string
+	// logger defines default logger.
```
lippserd, line 22: remove comment

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	logger *zap.SugaredLogger
+	// encoder stores zapcore.Encoder; used to build the logger
```
lippserd, line 24: // encoder defines the zapcore.Encoder,
// which is used to create the default logger and the child loggers

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	encoder zapcore.Encoder
+	// encoder stores zapcore.WriteSyncer; used to build the logger
```
lippserd, line 26: // syncer defines the zapcore.WriterSyncer,
// which is used to create the default logger and the child loggers

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	"go.uber.org/zap/zapcore"
+	"os"
+	"sync"
+	"time"
+)
+
+// Logging implements logging using: a switch to disable logging on Fatal log, default logger and level,
+// and stored child loggers and their options.
+type Logging struct {
+	// enabled implements a switch to disable logging on Fatal log.
```
lippserd, line 16: Please also document why we have that switch.

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+		logger:  logger.Sugar(),
+		encoder: encoder,
+		syncer:  syncer,
+		mu:      sync.Mutex{},
+		loggers: map[string]*zap.SugaredLogger{},
+		options: options,
+	}
+}
+
+// GetChildLogger returns a named child sugared logger for the default logger `Logging.logger`.
```
lippserd, line 69: GetChildLogger returns a named child logger.

Please also document how the log level is obtained.

pkg/config/logging.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/icinga/icingadb/internal"
+	"github.com/icinga/icingadb/internal/logging"
+)
+
+// Logging defines Logger configuration.
+type Logging struct {
+	Level  string `yaml:"level" default:"debug"`
+	Output string `yaml:"output" default:"console"`
+
+	logging.Options `yaml:"options,omitempty"`
```
lippserd, line 14: Does `omitempty` have any effect?

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	mu     sync.Mutex
+	// loggers stores named child loggers.
+	loggers map[string]*zap.SugaredLogger
+	options Options
+}
+
+// Options stores a `child-logger-name: level` pair mapping.
+type Options map[string]string
+
+// NewLogging returns a initialized Logging
```
lippserd, line 37: NewLogging returns a new Logging.

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+	"fmt"
+	"go.uber.org/atomic"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"os"
+	"sync"
+	"time"
+)
+
+// Logging implements logging using: a switch to disable logging on Fatal log, default logger and level,
```
lippserd, line 13: Logging implements access to a default logger and named child loggers.
Log levels can be configured per child via Options.
Logs either to the console or the systemd journal.
Provides Fatal log to stop all loggers before the error is logged and the program exits.

internal/logging/logging.go
```diff
@@ -0,0 +1,185 @@
...
+		return logger
+	}
+}
+
+// GetLogger returns the default sugared logger for the initialized Logging instance.
+func (l *Logging) GetLogger() *zap.SugaredLogger {
+	return l.logger
+}
+
+// selectEncoderSyncer returns Encoder and WriteSyncer for the logger based on output value.
```
lippserd, line 112: You don't need that method. You're only calling it once.

lippserd (changes requested):

pkg/config/logging.go
raviks789: No effect here. It will be removed.

internal/logging/journal.go
```diff
@@ -0,0 +1,105 @@
...
+	zapcore.ErrorLevel:  journal.PriErr,
+	zapcore.FatalLevel:  journal.PriCrit,
+	zapcore.PanicLevel:  journal.PriCrit,
+	zapcore.DPanicLevel: journal.PriCrit,
+}
+
+// logLine reads the logger fields which are used to construct and send log messages to journal
+type logLine struct {
+	Level      string `json:"level"`
+	Time       string `json:"ts"`
```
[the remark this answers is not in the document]
lippserd: ?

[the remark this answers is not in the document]
raviks789: No, we would have to write a custom method to unmarshal time.Time with custom type. So decided to have it as a string and parse it later in Write() method.

lippserd (changes requested):
Aside from the changes requested, I suggest removing the `output` option and changing the systemd service type from `simple` to `notify`. Systemd then sets the environment variable `NOTIFY_SOCKET`. We can then check that this variable is set and set the output to `systemd-journal`. The notification type requires that we actually notify systemd, so we need to do the following as well:

```
import (
   ...
    "github.com/coreos/go-systemd/daemon"
)
...
func run() int {
    cmd := command.New

    _, _ = daemon.SdNotify(false, daemon.SdNotifyReady)
...
```

And to set the `output` value we can do the following:

```
var output string
if _, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
    output = "systemd-journal"
} else {
    output = "console"
}

File: internal/logging/logging.go [hunk 30]
```diff
@@ -0,0 +1,185 @@
...
+	case "systemd-journal":
+		wr, err := newJournalWriter(os.Stderr, defaultEncConfig)
+
+		if err != nil {
+			panic(err)
+		}
+		encoder = zapcore.NewJSONEncoder(defaultEncConfig)
+		syncer = zapcore.AddSync(wr)
+	default:
+		panic(fmt.Sprintf("%s is not a valid logger output", output))
```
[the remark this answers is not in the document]
raviks789: I think default case not be required as we are going to use only either systemd-journal or console. As you have mentioned in this comment.
> Aside from the changes requested, I suggest removing the `output` option and changing the systemd service type from `simple` to `notify`. Systemd then sets the environment variable `NOTIFY_SOCKET`. We can then check that this variable is set and set the output to `systemd-journal`. The notification type requires that we actually notify systemd, so we need to do the following as well:
> 
> ```
> import (
>    ...
>     "github.com/coreos/go-systemd/daemon"
> )
> ...
> func run() int {
>     cmd := command.New
> 
>     _, _ = daemon.SdNotify(false, daemon.SdNotifyReady)
> ...
> ```
> 
> And to set the `output` value we can do the following:
> 
> ```
> var output string
> if _, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
>     output = "systemd-journal"
> } else {
>     output = "console"
> }
> ```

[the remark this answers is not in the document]
lippserd: No, I could still pass something wrong when using this package somewhere else.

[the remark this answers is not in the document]
raviks789: That makes sense as well.

doc/03-Configuration.md
```diff
@@ -24,6 +24,15 @@ database                 | **Required.** Database database.
...
 password                 | **Required.** Database password.
 
+## Logging Configuration <a id="configuration-logging"></a>
+
+Configuration of the logging component used by Icinga DB.
+
+Option                   | Description
+-------------------------|-----------------------------------------------
+level                    | **Optional.** Logger default level (debug, info, warn, error, dpanic, panic or fatal). Usually set to `debug` by default.
+options                  | **Optional.** Child loggers, with `Icinga Db component` => `logging level` map.<br /> `Icinga Db component`: database, redis, heartbeat, high-availability, config-sync, history, runtime-updates, overdue-sync, dump-signals, delta. <br /> `logging level`: debug, info, warn, error, dpanic, panic or fatal.
```
lippserd, line 34: I would not write long text in tables, but create a separate section under logging and document the components with a small description of what they do there. The description for `options` could then be something like this:

> **Optional.** Map of component name to logging level in order to set a different logging level for each component instead of the default one. See [logging components](#logging-components) for details.

lippserd (changes requested):

internal/logging/journal.go
```diff
@@ -0,0 +1,99 @@
...
+	LoggerName string `json:"logger"`
+	Caller     string `json:"caller"`
+	Message    string `json:"msg"`
+	Stack      string `json:"stacktrace"`
+}
+
+func (w *journalWriter) Write(p []byte) (int, error) {
```
N-o-X, line 52: Missing description/comments

N-o-X (changes requested):
Works great and the code is also fine. Just needs those 3 requested changes and a rebase and it's ready to merge.

N-o-X:
Packaging PRs needed for this PR:
https://git.icinga.com/packaging/rpm-icingadb/-/merge_requests/14
https://git.icinga.com/packaging/deb-icingadb/-/merge_requests/4

internal/logging/journal.go
```diff
@@ -0,0 +1,101 @@
...
+	message, err := w.journalEncoder.EncodeEntry(entry, nil)
+
+	if err != nil {
+		panic(err)
+	}
+
+	err = journal.Send(message.String(), pri, map[string]string{
+		"PACKAGE":           filepath.Dir(line.Caller),
+		"SYSLOG_IDENTIFIER": filepath.Base(os.Args[0]),
+	})
```
julianbrost, line 84: Shouldn't journald already set `SYSLOG_IDENTIFIER` by itself?

Also, to be honest, is that `PACKAGE` extra field really worth the effort of implementing anything systemd-specific? The [log/syslog](https://pkg.go.dev/log/syslog) package already provides a Writer that could just be plugged into zap and should give mostly the same effect as journald provides syslog compatibility.

In addition, when testing with `NOTIFY_SOCKET= go run ./cmd/icingadb`, this even just logs `PACKAGE=.` (haven't tried if this behaves differently when properly compiling and running as a systemd unit).

internal/logging/journal.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil {
+			panic(err)
+		}
+		entry.Time = tm
+		message, err := w.consoleEncoder.EncodeEntry(entry, nil)
+		if err != nil {
+			panic(err)
+		}
+
+		return w.Writer.Write(message.Bytes())
```
julianbrost, line 97: Also here, most likely `len(p) != len(message.Bytes())`, so incorrect implementation of `io.Writer`.

julianbrost (changes requested):

cmd/icingadb/main.go
```diff
@@ -184,7 +215,7 @@ func run() int {
 							com.ErrgroupReceive(g, dbErrs)
 
 							g.Go(func() error {
-								return s.ApplyDelta(ctx, icingadb.NewDelta(ctx, actualCvs, cvs1, cv, logger))
+								return s.ApplyDelta(ctx, icingadb.NewDelta(ctx, actualCvs, cvs1, cv, logs.GetChildLogger("delta")))
```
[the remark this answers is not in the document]
lippserd: `config-sync` for those components you've mentioned sounds good to me.

cmd/icingadb/main.go
```diff
@@ -33,47 +35,76 @@ func main() {
...
+	// When started by systemd, NOTIFY_SOCKET is set by systemd for Type=notify supervised services, which is the
+	// default setting for the Icinga DB service. So we notify that Icinga DB finished starting up and also configure
+	// logging into the systemd journal instead of stderr.
+	// Always add this line before output assignment
+	_, _ = daemon.SdNotify(false, daemon.SdNotifyReady)
+	var output string
+	// When the type in icingadb.service is set to notify then set output to "systemd-journal",
+	// else set it to console, that is write log messages to stdErr.
+	if _, ok := os.LookupEnv("NOTIFY_SOCKET"); ok {
+		output = "systemd-journal"
```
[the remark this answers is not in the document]
lippserd: Yes, `Type = notify` is only there to say if it runs under systemd. I would leave it that way, because the alternative can write to the journal if systemd is available and Icinga DB is started somehow differently.

cmd/icingadb/main.go
```diff
@@ -33,47 +35,76 @@ func main() {
 
 func run() int {
 	cmd := command.New()
+	// When started by systemd, NOTIFY_SOCKET is set by systemd for Type=notify supervised services, which is the
+	// default setting for the Icinga DB service. So we notify that Icinga DB finished starting up and also configure
+	// logging into the systemd journal instead of stderr.
+	// Always add this line before output assignment
+	_, _ = daemon.SdNotify(false, daemon.SdNotifyReady)
```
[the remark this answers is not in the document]
lippserd: I would not wait for the 5 minute reconnect. I would try to connect to Redis and the database before it notifies the ready state and cancel if it is not a retryable error. Would that be ok?

doc/03-Configuration.md
```diff
@@ -24,6 +24,31 @@ database                 | **Required.** Database database.
 user                     | **Required.** Database username.
 password                 | **Required.** Database password.
 
+## Logging Configuration <a id="configuration-logging"></a>
+
+Configuration of the logging component used by Icinga DB.
+
+Option                   | Description
+-------------------------|-----------------------------------------------
+level                    | **Optional.** Specifies the default logging level. Can be set to `fatal`, `error`, `warning`, `info` or `debug`. Defaults to `debug`.
```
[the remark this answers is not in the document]
lippserd: Makes sense.

internal/logging/journal.go
```diff
@@ -0,0 +1,101 @@
...
+	Stack      string `json:"stacktrace"`
+}
+
+// Write converts the given byte slice into logLine before trying to send the log entry to journald.
+// Should sending the data to journald fail, Write falls back to stderr for error logging.
+func (w *journalWriter) Write(p []byte) (int, error) {
+	line := &logLine{}
+	if err := json.NewDecoder(bytes.NewReader(p)).Decode(line); err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
lippserd: I believe there is no better way right now without implementing a custom encoder. CoreOS does the same thing.

internal/logging/logging.go
```diff
@@ -0,0 +1,195 @@
...
+
+		logger := l.logger.Desugar().WithOptions(
+			zap.WrapCore(func(c zapcore.Core) zapcore.Core {
+				return zapcore.NewCore(
+					l.encoder,
+					l.syncer,
+					zap.LevelEnablerFunc(func(lvl zapcore.Level) bool {
+						// Always log fatals, but allow other levels to be completely disabled.
+						return zap.FatalLevel.Enabled(lvl) || l.enabled.Load() && atom.Enabled(lvl)
+					}))
```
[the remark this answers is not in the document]
lippserd: Makes sense.

internal/logging/logging.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// GetLogger returns the default logger.
+func (l *Logging) GetLogger() *zap.SugaredLogger {
+	return l.logger
+}
+
+const (
+	ExitFailure = 1
+)
```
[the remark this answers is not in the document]
lippserd: That was my intention when writing https://github.com/Icinga/icingadb/pull/338#discussion_r685108746:

>  I opt for moving the constants to a new package, e.g. `exitcode`. 

Go for that?

internal/logging/logging.go
```diff
@@ -0,0 +1,195 @@
...
+		logger = logger.WithOptions(zap.OnFatal(zapcore.WriteThenGoexit))
+	}
+
+	logged := make(chan struct{})
+	go func() {
+		defer close(logged)
+		defer logger.Sync()
+		logger.Fatal(fmt.Sprintf("%+v", err))
+	}()
+	<-logged
```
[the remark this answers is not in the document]
lippserd: Zap doesn't let you use `WriteThenNoop` in` Fatal`.

internal/logging/journal.go
lippserd: That is true. Should we just return `len(p)` here?

internal/logging/journal.go
```diff
@@ -0,0 +1,101 @@
...
+		entry.Time = tm
+		message, err := w.consoleEncoder.EncodeEntry(entry, nil)
+		if err != nil {
+			panic(err)
+		}
+
+		return w.Writer.Write(message.Bytes())
+	}
+
+	return 0, nil
```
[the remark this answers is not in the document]
lippserd: Again, just return `len(p)`?

cmd/icingadb/main.go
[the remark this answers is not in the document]
julianbrost: Would be an improvement but still somewhat strange if you restart Icinga DB, it reports OK and dies after 5 minutes because it was never connected to the database.

cmd/icingadb/main.go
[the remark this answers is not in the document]
julianbrost: Well, what about just doing `ExecStart=/usr/bin/icingadb -log journald`? Would make it obvious why it logs to which destination.

internal/logging/journal.go
julianbrost: On success, yes, on error, maybe that, maybe 0. `io.Writer` docs doesn't seem to forbid returning `len(p)` with `err != nil`, however it sound strange to say everything was written and return an error at the same time.

internal/logging/journal.go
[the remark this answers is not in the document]
julianbrost: Yes

internal/logging/journal.go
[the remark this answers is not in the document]
julianbrost: Do you have a link to where CoreOS does that? Also haven't looked to much into zap but what would be the problem with implementing an encoder?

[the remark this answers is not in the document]
lippserd: Sry, it was etcd: https://github.com/etcd-io/etcd/blob/main/client/pkg/logutil/zap_journal.go#L54

internal/logging/logging.go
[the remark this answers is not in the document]
julianbrost: Another option might be to just sleep in the writer (or encoder should we implement that) right after sending the message to journald. Then zap could just exit as normal.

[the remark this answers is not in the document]
lippserd: Is there really a benefit since we would have to pass `logging.enabled` to the writer and always check whether we're dealing with the `fatal` level?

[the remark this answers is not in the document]
julianbrost: Apart from getting rid of this `WriteThenGoexit` hack, it would also allow using the regular `logger.Fatal*()` functions that currently wouldn't do the sleep.

cmd/icingadb/main.go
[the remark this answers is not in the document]
lippserd: Without `Type = notify` we would have the same situation. To me, signaling readiness means that Icinga DB has finished parsing its configuration and is ready to do its job. The connection to the database and Redis is part of this job and I would not make the ready state dependent on it. Because if the database connection or Redis is gone during runtime, we also do not signal that we are no longer doing anything. I would rather send a `STATUS` string that we're retrying those connections.

PR #312, opened 2021-07-13
Bulk statement size

lippserd:


julianbrost (commented):
`1<<14` could be made a named constant like `stmtMaxPlaceholders` with a comment stating the source for that number or some explanation why it was chosen. Seems like the limit imposed by MySQL would be `1<<16`.

lippserd:
> `1<<14` could be made a named constant like `stmtMaxPlaceholders` with a comment stating the source for that number or some explanation why it was chosen. Seems like the limit imposed by MySQL would be `1<<16`.

I've documented the options and made them configurable too.

pkg/icingadb/db.go
```diff
@@ -373,7 +384,7 @@ func (db *DB) CreateStreamed(ctx context.Context, entities <-chan contracts.Enti
 	sem := db.getSemaphoreForTable(utils.TableName(first))
 	stmt, placeholders := db.BuildInsertStmt(first)
 
-	return db.NamedBulkExec(ctx, stmt, 1<<15/placeholders, sem, forward, nil)
+	return db.NamedBulkExec(ctx, stmt, db.options.MaxPlaceholdersPerStatement/placeholders, sem, forward, nil)
```
julianbrost, line 387: Could become 0 if a small value is set. There's no documentation on what the required minimum would be and even if it was documented, it potentially would have to be updated when we add columns and so on. So maybe go with something like `max(maxPlaceholders/placeholders, 1)` here and in the other places?

julianbrost (changes requested):

julianbrost (approved):

PR #313, opened 2021-07-13
Update authors and remove authors whose code is no longer in the repo

lippserd:
- [ ] Decide how to keep this file up to date automatically, e.g. https://github.com/Icinga/icingadb/pull/313#issuecomment-892981183

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -2,3 +2,4 @@ Alexander A. Klimov <alexander.klimov@icinga.com> <alexander.klimov@icinga.com>
 Diana Flach <crunsher@bamberg.ccc.de> <Crunsher@users.noreply.github.com>
 Diana Flach <crunsher@bamberg.ccc.de> <jean-marcel.flach@icinga.com>
 Diana Flach <crunsher@bamberg.ccc.de> <jean.flach@icinga.com>
+Henrik Triem <henrik.triem@netways.de> <43344334+htriem@users.noreply.github.com>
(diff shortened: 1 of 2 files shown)


julianbrost:
Any plans for how we'll keep this file up-to-date in the long run? Like if we do it like in the icinga2 repo (i.e. automatically populate from Git authors), these people would show up again.

lippserd:
With something like that:

`git ls-tree --name-only -r -z HEAD | xargs -0 -n1 git blame --line-porcelain | grep -ae "^author " -A1`

Al2Klimov:
Are they such a burden for us? I'd just enable Probot and let it do its thing.

lippserd:
> Are they such a burden for us? I'd just enable Probot and let it do its thing.

We have rewritten the whole thing 100%. So it is a lie to claim them as authors.

lippserd:
Can we configure the probot to do its thing after a certain commit?

Al2Klimov:
I could implement it. (But I'd consult colleague Drescher first (just to be sure).)

julianbrost:
I think we should update the AUTHORS file in some form for RC2, even if it's not yet automated. Not sure about the mailmap though, I mean the commits are still somewhere in the history of this repo, so I wouldn't remove stuff from that file.

htriem (approved):
I really should switch to my icinga.com-Mail, but this is looking good for now.

PR #314, opened 2021-07-13
Fix DUPLICATE KEY errors after HA handover/takeover

lippserd:


julianbrost:
How would the old code lead to SQL errors? From the diff, I'd rather suspect this would fix something like a "you're writing to a nil map" panic.

lippserd:
The maps are set here: https://github.com/Icinga/icingadb/pull/314/files#diff-1bf222be586d5eccdb10e889435785fe46f18c390e56bf0f4a2d368c64914dd9R135

Had a look at the code again. I though that the delta is hanging around after handover/takeover but we're calling NewDelta so this fix really does not do anything. Sorry for the noise.

Duplicate key errors are still there w/ this "fix". I was just lucky when testing it previously.

PR #318, opened 2021-07-16
AcknowledgementHistory#Init(): ensure #IsPersistent and #IsSticky are not NULL

### LINKED ISSUE #305: Removing an acknowledgement crashes Icinga DB
When deleting an acknowledgement in the web interface, the latest Icinga DB master crashes:

```
2021-07-01T13:49:26.144Z	FATAL	icingadb/main.go:211	Error 1048: Column 'is_sticky' cannot be null
can't perform "INSERT INTO acknowledgement_history (environment_id,object_type,service_id,cleared_by,is_sticky,is_persistent,clear_time,comment,expire_time,host_id,set_time,endpoint_id,id,author) VALUES (:environment_id,:object_type,:service_id,:cleared_by,:is_sticky,:is_persistent,:clear_time,:comment,:expire_time,:host_id,:set_time,:endpoint_id,:id,:author) ON DUPLICATE KEY UPDATE clear_time = :clear_time,cleared_by = :cleared_by"
```

Setting an acknowledgement works fine, but when removing it, the above error happens. Maybe the event message has fewer JSON keys set? (But that's just speculation, haven't investigated further.)

Al2Klimov:
fixes #305

### DIFF
--- a/pkg/icingadb/v1/history/ack.go
+++ b/pkg/icingadb/v1/history/ack.go
@@ -29,6 +29,11 @@ type AcknowledgementHistory struct {
 	IsSticky                 types.Bool      `json:"is_sticky"`
 }
 
+func (ah *AcknowledgementHistory) Init() {
+	ah.IsPersistent = types.Bool{Bool: false, Valid: true}
+	ah.IsSticky = types.Bool{Bool: false, Valid: true}
+}
+
 type HistoryAck struct {
 	HistoryMeta              `json:",inline"`
 	AcknowledgementHistoryId types.Binary `json:"id"`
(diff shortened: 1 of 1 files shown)


lippserd:
Shouldn't Icinga provide those fields?

Al2Klimov:
Not necessary. Icinga already provides them on ack additions. On ack removals they don’t matter, but MySQL formally requires them to comply with NOT NULL. 🤷‍♂️

lippserd:
Then I would question the Upserter logic, as this should really just be an UPDATE. If the entry is not there, it would be created with wrong values.

PR #319, opened 2021-07-19
Pointer receivers, Cond usage, pass ctx and Godoc for Heartbeat

### LINKED ISSUE #316: Heartbeat.Err() will always return nil
Most receiver functions for `Heartbeat` operate on a copy instead of a pointer.

Even though `Heartbeat.setError()` operates on a pointer:
https://github.com/Icinga/icingadb/blob/cf3a13d3f50214efa90e3676076240cf15e6dea0/pkg/icingaredis/heartbeat.go#L149-L153

It is only called from within `Heartbeat.controller()` which itself operates on a copy:
https://github.com/Icinga/icingadb/blob/cf3a13d3f50214efa90e3676076240cf15e6dea0/pkg/icingaredis/heartbeat.go#L79

So `Heartbeat.Err()` could only return something not `nil` when called on the copy within `Heartbeat.controller()` (there is no such call). If any external caller would call it (or `Heartbeat.Close()` which in turn returns `Heartbeat.Err()`), the return value would always be `nil`.

Given that neither seems to be called somewhere else in the code base, this won't cause any problems right now, but still looks broken.

lippserd:
Heartbeat now uses pointer receivers for its methods because some methods actually change the heartbeat values. The context is no longer stored in the structure, but passed to the controller loop. The beat and the lost channels are replaced by Cond and the last heartbeat is stored independently to not be affected by a slow HA receiver. If the database connections are occupied by the config, HA cannot update the instance and does not read from the beat channel in time. In addition, heartbeat errors are no longer swallowed, but handled in HA.

fixes #316

Please note that I think it also makes sense to use a separate connection pool for HA, but that's subject to a different PR.

### DIFF
--- a/pkg/icingaredis/heartbeat.go
+++ b/pkg/icingaredis/heartbeat.go
@@ -3,6 +3,7 @@ package icingaredis
 import (
 	"context"
 	"github.com/go-redis/redis/v8"
+	"github.com/icinga/icingadb/pkg/com"
 	v1 "github.com/icinga/icingadb/pkg/icingaredis/v1"
 	"github.com/icinga/icingadb/pkg/utils"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


pkg/icingaredis/heartbeat.go
```diff
@@ -12,79 +13,100 @@ import (
...
+	beat      *com.Cond
+	cancel    context.CancelFunc
+	client    *Client
+	done      chan struct{}
+	errMu     *sync.Mutex
+	err       error
+	logger    *zap.SugaredLogger
+	lost      *com.Cond
+	message   *v1.StatsMessage
+	messageMu *sync.Mutex
```
julianbrost, line 33: Given that now everything operates in `Heartbeat` pointers, there is no need for the two mutexes to be pointers. Then their zero value is a fine default and they even don't have to be initialized explicitly.

julianbrost (changes requested):

julianbrost (approved):

PR #323, opened 2021-07-22
Downtime: Add parent_id

N-o-X:
This PR adds the downtime attribute `parent_id` added to Icinga 2 with https://github.com/Icinga/icinga2/pull/8913 to Icinga DB.

### DIFF
--- a/pkg/icingadb/v1/downtime.go
+++ b/pkg/icingadb/v1/downtime.go
@@ -10,6 +10,7 @@ type Downtime struct {
 	EnvironmentMeta    `json:",inline"`
 	NameMeta           `json:",inline"`
 	TriggeredById      types.Binary    `json:"triggered_by_id"`
+	ParentId           types.Binary    `json:"parent_id"`
 	ObjectType         string          `json:"object_type"`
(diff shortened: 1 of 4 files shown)


lippserd:
Also with # 301 in mind, are `trigger_by_id`, `parent_downtime_id` and `scheduled_by` mutually exclusive or could two or all of them be set?

N-o-X:
As we already found out in our meeting a few hours ago, `trigger_by_id` and `scheduled_by` will both be set, when using `child_options` in a `ScheduledDowntime`.

lippserd (changes requested):
Even if the other columns don't have a meaningful comment at the moment, it would be good if the comment for `parent_id` describes what the column represents.

N-o-X:
> Even if the other columns don't have a meaningful comment at the moment, it would be good if the comment for `parent_id` describes what the column represents.

I'd create a separate PR to add descriptions for all 3 columns.

N-o-X:
@cla-bot check

lippserd (approved):

PR #325, opened 2021-07-28
Stream state updates from icinga:runtime:state

lippserd:
Icinga now sends runtime updates in two separate channels,
icinga:runtime for config updates and icinga:runtime:state for
state updates. With this change, Icinga DB reads from these two
streams. This is a preparation so that state updates can be
streamed directly after a (re)start of Icinga or Icinga DB without
waiting for the config sync, as it is currently done.

refs Icinga/icinga2#8915

### DIFF
--- a/pkg/icingaredis/utils.go
+++ b/pkg/icingaredis/utils.go
@@ -12,6 +12,25 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+// Streams represents a Redis stream key to ID mapping.
+type Streams map[string]string
+
+// Option returns the Redis stream key to ID mapping
+// as a slice of stream keys followed by their IDs
+// that is compatible for the Redis STREAMS option.
+func (s Streams) Option() []string {
+	// len*2 because we're appending the IDs later.
+	streams := make([]string, 0, len(s)*2)
+	ids := make([]string, 0, len(s))
+
+	for key, id := range s {
+		streams = append(streams, key)
+		ids = append(ids, id)
+	}
+
+	return append(streams, ids...)
+}
+
 func CreateEntities(ctx context.Context, factoryFunc contracts.EntityFactoryFunc, pairs <-chan HPair, concurrent int) (<-chan contracts.Entity, <-chan error) {
 	entities := make(chan contracts.Entity)
 	g, ctx := errgroup.WithContext(ctx)
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -94,17 +94,19 @@ func run() int {
 						// Runtime updates must wait for configuration synchronization to complete.
 						wg := sync.WaitGroup{}
 
+						// Get the last IDs of the runtime update streams before starting anything else,
+						// otherwise updates may be lost.
+						runtimeUpdateStreams, err := rt.Streams(ctx)
+						if err != nil {
+							logger.Fatalf("%+v", err)
+						}
+
 						dump := icingadb.NewDumpSignals(rc, logger)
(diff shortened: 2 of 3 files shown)


julianbrost (changes requested):

julianbrost (changes requested):
Seems like `XRead().Result()` only returns the streams with messages. I added `r.logger.Infow("xread", zap.Strings("streams", streams))` right before the call to `XRead()` and this is what I've got:
```
2021-07-29T14:29:59.030+0200	INFO	icingadb/runtime_updates.go:102	xread	{"streams": ["icinga:runtime", "icinga:runtime:state", "1627561541341-9", "1627561591601-0"]}
2021-07-29T14:30:06.061+0200	INFO	icingadb/runtime_updates.go:102	xread	{"streams": ["icinga:runtime", "1627561806059-0"]}
2021-07-29T14:30:06.062+0200	INFO	icingadb/runtime_updates.go:102	xread	{"streams": ["icinga:runtime", "1627561806060-1"]}
```

So looks like the current code will only sync either of the streams now (unless you happen to have new messages in both streams each time).

Suggestion: make the last IDs a `map[string]string` from key to ID throughout the code (I find the []string{key1, key2, ..., id1, id2, ...)` a strange format anyways) and convert it to the required `[]string` for each `XRead()`. Other option would be to just update the IDs in-place in `streams`, but that would be some ugly indexing work.

lippserd:
> Seems like `XRead().Result()` only returns the streams with messages. I added `r.logger.Infow("xread", zap.Strings("streams", streams))` right before the call to `XRead()` and this is what I've got:
> 
> ```
> 2021-07-29T14:29:59.030+0200	INFO	icingadb/runtime_updates.go:102	xread	{"streams": ["icinga:runtime", "icinga:runtime:state", "1627561541341-9", "1627561591601-0"]}
> 2021-07-29T14:30:06.061+0200	INFO	icingadb/runtime_updates.go:102	xread	{"streams": ["icinga:runtime", "1627561806059-0"]}
> 2021-07-29T14:30:06.062+0200	INFO	icingadb/runtime_updates.go:102	xread	{"streams": ["icinga:runtime", "1627561806060-1"]}
> ```
> 
> So looks like the current code will only sync either of the streams now (unless you happen to have new messages in both streams each time).
> 
> Suggestion: make the last IDs a `map[string]string` from key to ID throughout the code (I find the `[]string{key1, key2, ..., id1, id2, ...)` a strange format anyways) and convert it to the required `[]string` for each `XRead()`. Other option would be to just update the IDs in-place in `streams`, but that would be some ugly indexing work.

Good catch. I introduced a new type for the streams:

https://github.com/Icinga/icingadb/pull/325/files#diff-dc1ad4347acc2a73e8d830a986f72b63a442ae0f678711ee3c91651d53cf17fbR16

pkg/icingaredis/utils.go
```diff
@@ -12,6 +12,25 @@ import (
...
+
+// Option returns the Redis stream key to ID mapping
+// as a slice of stream keys followed by their IDs
+// that is compatible for the Redis STREAMS option.
+func (s Streams) Option() []string {
+	// len*2 because we're appending the IDs later.
+	streams := make([]string, 0, len(s)*2)
+	ids := make([]string, 0, len(s))
+
+	for k, i := range s {
```
julianbrost, line 26: I find using `i` as a non-counter variable in a loop counterintuitive. Spelling out the variable(s) here won't hurt I guess.

julianbrost (approved):

PR #326, opened 2021-07-29
Don't append(x,y) if x is not to be modified

### LINKED ISSUE #324: Use append(x,y) only to modify x in-place
## Describe the bug

https://yourbasic.org/golang/gotcha-append/

## Expected behavior

Not `append(x,y)`, but `append(append([]T(nil),x...),y)` for derivative slice creation.

Al2Klimov:
... as append(x,y) would modify x.

fixes #324


julianbrost (changes requested):
Merge conflict due to #325. But looks like a good change otherwise.

julianbrost (approved):

PR #327, opened 2021-07-30
Insert environment

### LINKED ISSUE #292: Environment table not populated
Since merging #277, the `environment` table is no longer populated with a row for the current environment.

1. `TRUNCATE environment;` (in case it's still there from an earlier version)
2. Restart Icinga DB
3. Wait for things to settle
4. `SELECT * FROM environment;`

No rows are returned, even though there should be one.

lippserd:
With this change Icinga DB will insert the environment after each
heartbeat takeover if it does not already exist in the database as
the environment may have changed, although this is likely to happen
very rarely.

Instead of checking whether the environment already exists,
uses an INSERT statement that does nothing if it does.

fixes #292

### DIFF
--- a/pkg/icingadb/db.go
+++ b/pkg/icingadb/db.go
@@ -78,6 +78,20 @@ func (db *DB) BuildInsertStmt(into interface{}) (string, int) {
 	), len(columns)
 }
 
+// BuildInsertIgnoreStmt returns an INSERT statement for the specified struct for
+// which the database ignores rows that have already been inserted.
+func (db *DB) BuildInsertIgnoreStmt(into interface{}) (string, int) {
+	columns := db.BuildColumns(into)
+
+	return fmt.Sprintf(
+		// MySQL treats UPDATE id = id as a no-op.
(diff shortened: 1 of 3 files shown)


julianbrost:
Previously, Icinga DB bailed out if the environment suddenly changed: https://github.com/Icinga/icingadb/blob/d9e3c8b195ccacdca0d565239f035b3de9a381ea/ha/ha.go#L338-L342

The history provides no explicit reasoning for why that was done (got introduced by 80e1726cdfc690fbcef262a0f3308660103d4b77). Does anyone remember a particular reason for this? Otherwise I'll think about whether anything would break due to a sudden environment change with the current implementation.

> although this is likely to happen very rarely

Are there any cases where you'd expect this to happen for a valid reason?

lippserd:
> Are there any cases where you'd expect this to happen for a valid reason?

For example, I've already started Icinga DB, but I don't like the empty string as an environment. Or I have already started Icinga DB and after a while I want to add another environment to Icinga DB.

Sure, after renaming an existing environment, I lose my history, but that's something for a feature or our documentation.

PR #329, opened 2021-08-03
Use SET SESSION for schema modes

lippserd:
As of version 8, MySQL is stricter when it comes to setting system
and session variables and requires that the user importing our schema
be granted the SESSION_VARIABLES_ADMIN permission. Although the session
scope is already the default, we will now use SET SESSION for the modes
in our scheme for the sake of clarity.

### DIFF
--- a/schema/mysql/schema.sql
+++ b/schema/mysql/schema.sql
@@ -1,7 +1,7 @@
 -- IcingaDB | (c) 2019 Icinga GmbH | GPLv2+
 
-SET sql_mode = 'STRICT_ALL_TABLES,NO_ENGINE_SUBSTITUTION';
-SET innodb_strict_mode = 1;
+SET SESSION sql_mode = 'STRICT_ALL_TABLES,NO_ENGINE_SUBSTITUTION';
+SET SESSION innodb_strict_mode = 1;
 
 CREATE TABLE host (
   id binary(20) NOT NULL COMMENT 'sha1(environment.name + name)',


julianbrost:
If I understand correctly, `SET` without an additional scope keyword should already set the session variable:

>  To assign a value to a session system variable, precede the variable name by the `SESSION` or `LOCAL` keyword, by the `@@SESSION.`, `@@LOCAL.`, or `@@` qualifier, or by no keyword or no modifier at all:

-- https://dev.mysql.com/doc/refman/8.0/en/set-variable.html#set-variable-system-variables

Is that documentation wrong? Is there a way to see a difference in behavior for both versions of the schema file before/after this PR?

lippserd:
No, the documentation is correct. I tested some MySQL containers for the Docker Compose Icinga playground and ran into problems there, which (in the end) are not related to the session change, but rather to a lack of permissions: https://github.com/lippserd/docker-compose-icinga/blob/master/env/icingadb-mysql/privileges.sql

Since I then tested with both changes, I thought that I also need the session change and created a PR.

Now tested again with only the session admin permission and it works, as it should be expected from the docs.

So we can close the PR or keep the changes to quickly see that we are actually changing session variables. (I would update the commit and description then.)

julianbrost:
Well, given that I had to search the documentation for the default scope, adding it for clarity won't hurt I guess.

lippserd:
Updated the commit and the description. Ready to review!

julianbrost:
I don't see why the commit still mentions MySQL 8 and `SESSION_VARIABLES_ADMIN`. How does that error show? Like does it help to actually know a permission for session variables is missing if it says `SET SESSION` instead of just `SET`?

lippserd:
MySQL 8 complains, that the user either needs the `SESSION_VARIABLES_ADMIN` or the `SYSTEM_VARIABLES_ADMIN` permission. And since `SESSION_VARIABLES_ADMIN` has less permissions than `SYSTEM_VARIABLES_ADMIN` it is mentioned for clarity.
If you start a MySQL 8 docker container for example which is tied to a user and a database, this user won't have any of these permissions. For me this is the reason for the change.

julianbrost (approved):
So the exact error message with MySQL 8 is:
```
mysql> SET innodb_strict_mode = 1;
ERROR 1227 (42000): Access denied; you need (at least one of) the SYSTEM_VARIABLES_ADMIN or SESSION_VARIABLES_ADMIN privilege(s) for this operation
```
Maybe `SET SESSION` helps someone to know that `SESSION_VARIABLES_ADMIN` might be more appropriate.

PR #336, opened 2021-08-04
Environments

lippserd:
closes #327


Al2Klimov (approved):
I guess that’s your expected result:

```
2021-08-05T16:50:39.921+0200	WARN	icingadb/ha.go:155	Got new environment	{"current": "env1", "new": "env2"}
2021-08-05T16:50:39.921+0200	INFO	icingadb/main.go:236	Restarting HA
2021-08-05T16:50:39.921+0200	DEBUG	icingadb/db.go:181	Executed DELETE FROM usergroup WHERE id IN (?) with 0 rows in 1m5.404771912s
```

```
MariaDB [idb]> select environment_id,count(*)from service group by environment_id;
+--------------------------------------------+----------+
| environment_id                             | count(*) |
+--------------------------------------------+----------+
| 0x0D028BE76AA3724FA7F09610A8732C920B744733 |       11 |
| 0x42627E015B320B884B199D9046545D1DC8300212 |       11 |
+--------------------------------------------+----------+
2 rows in set (0.008 sec)

MariaDB [idb]>
```

julianbrost (changes requested):

pkg/icingadb/db.go
```diff
@@ -89,12 +89,33 @@ func (db *DB) BuildInsertStmt(into interface{}) (string, int) {
...
+func (db *DB) BuildSelectStmt(from interface{}, into interface{}) string {
+	q := fmt.Sprintf(
 		`SELECT %s FROM %s`,
 		strings.Join(db.BuildColumns(into), ", "),
 		utils.TableName(from),
 	)
+
+	if scoper, ok := from.(contracts.Scoper); ok {
+		where, _ := db.BuildWhere(scoper.Scope())
+		q = q + ` WHERE ` + where
```
Al2Klimov, line 115: Btw. I'd let every of my trainees dig through half of Google on how to shorten this line by three characters... but I'm afraid you're my boss. 🤷‍♂️

julianbrost: Hmm, let me guess! ;)
```go
wh, _ := db.BuildWhere(scoper.Scope())
q = q + ` WHERE ` + wh
```

Al2Klimov: Not the three characters I meant :)

pkg/icingadb/ha.go
```diff
@@ -78,6 +78,11 @@ func (h *HA) Done() <-chan struct{} {
 	return h.done
 }
 
+// Environment returns the current environment.
+func (h *HA) Environment() *v1.Environment {
+	return &h.environment
+}
```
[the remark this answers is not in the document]
lippserd: True, fixed!

pkg/icingadb/db.go
lippserd: > but I'm afraid you're my boss. 🤷‍♂️

But I also want to learn, so what line are we talking about exactly?

pkg/icingadb/v1/environment.go
```diff
@@ -1,10 +1,42 @@
...
+	return context.WithValue(parent, environmentContextKey, e)
+}
+
+// FromCtx swaps this Environment to the Environment value stored in ctx, if any:
+//
+// 	var e Environment
+// 	if !e.FromCtx(ctx) {
+// 		// Error handling.
+// 	}
+func (e *Environment) FromCtx(ctx context.Context) bool {
```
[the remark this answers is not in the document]
lippserd: I also looked at the examples, but they have a package scope for the user and we have a broader scope. This is why I added the methods to the env struct to have scope. Otherwise we would have `EnvironmentFromContext` and` NewContextWithEnvironment` or something like that. And if we introduce more of this, we would have more and more unscoped functions. I think it's cleaner this way and it could satisfy an interface too if we ever need to.

pkg/icingadb/db.go
Al2Klimov: 115

pkg/icingadb/v1/environment.go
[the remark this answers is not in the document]
julianbrost: IMO this is pretty similar to a `func NewEnvironment() *Environment` function, which is also not really scoped and you probably wouldn't make a `func (e *Environment) Init()` instead.

pkg/icingadb/db.go
lippserd: Ok, I have no clue! 😆 Please help!

julianbrost: If there only was a way to append to a string without typing the variable name twice!

pkg/icingadb/v1/environment.go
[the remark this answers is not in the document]
lippserd: I find the comparison a little bit far-fetched as this is a constructor, a single function and in your example is guaranteed to return a new environment while `FromContext` belongs to `NewContext` and extracts an existing environment, if any. And by the loss of scope I mean that one function is then in the package and the other is attached to the struct. For me, both functions belong together and that's why I put both in the same scope. But I can just change that.
Also, I wouldn't just return `* Environment` and require a a non-nil check. To me this is unexpected behavior and the bool clearly indicates that this is an operation that could fail.

pkg/icingadb/db.go
lippserd: 🐘

pkg/icingadb/v1/environment.go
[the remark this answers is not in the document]
lippserd: Ready to review.

pkg/icingadb/ha.go
```diff
@@ -78,10 +81,18 @@ func (h *HA) Done() <-chan struct{} {
 	return h.done
 }
 
+// Environment returns the current environment.
+func (h *HA) Environment() *v1.Environment {
+	h.environmentMu.Lock()
+	defer h.environmentMu.Unlock()
+
+	return &h.environment
+}
```
julianbrost, line 90: Locking here doesn't solve the issue. The controller loop might still update `h.environment.Id` after the pointer was returned here. Options would be:
1. Return a copy here (taking the copy is synchronized, afterwards they're independent objects, so no more need to sync)
2. Make `h.environment` a `*Environment` and never update existing `Environment` instances but instead always create a new one (after initialization, all accesses are read-only and need no further synchronization)
3. Expose `h.environmentMu` to allow the caller to properly synchronize (wouldn't recommend doing this but would theretically be possible)

julianbrost (changes requested):

pkg/icingadb/ha.go
lippserd: Implemented option 2, thanks!

julianbrost:
So how do we want to continue here? The problem I described in the previous comment is now covered by #349 I guess. I only observed this with this PR, however @lippserd said the same happened for him after heartbeat loss+recovery. So that problem is not necessarily related to this PR. I could just review everything but environment changes with this PR, but that's somewhat strange given that code for handling environment changes is explicitly part of this PR and trying to test that crashes Icinga DB for me quite reliably. Or rather wait with this PR until someone found that bug so that I can properly test this PR?

julianbrost:
There's a problem with how we use Redis on environment changes: Icinga DB reads two different Redis streams: `icinga:stats` for heartbeats and `icinga:dump` for dump signals. Events on both can trigger a new initial sync on the Icinga DB side: a heartbeat restarts the HA for the new environment and this starts a new sync. Also, when dump dome signals are received after an in progress signal, this also starts a new sync. These are independent Redis streams, and there's no extra synchronization on the Go side, so right now a dump done signal may be processed before the heartbeat with the new environment, starting a new sync with the data from the new environment but still under the assumption of operating with the old environment ID from the old heartbeat. This then uses a mix of both environments in the delta calculation and in the case of object deletions, there's no real way to real way to even detect that it's from a wrong environment (the object is just missing in Redis, there's no environment ID you could compare).

So I think these are the options for moving on here:
1. Make the environment name an explicit configuration option in icingadb just as in icinga2, so if you want to change that, you have to change it for both daemons and restart both. Then we can try to do our best in detecting a mismatch there and just bail out if we see objects from another environment.
2. Bail out on environment changes, just like RC1 did. However there's still a risk that some unintentional operations are executed until a new environment ID is seen for the first time. This isn't much worse than the previous option from a reliability perspective, but in the first case, if something goes wrong, this can be attributed to a misconfiguration.
3. Make heartbeat and dump signal handling more aware of each other: icinga2 ensures to send a heartbeat with the current environment before starting a dump, so after receiving a dump done signal, icingadb could ensure that any heartbeat written until then was processed (worst case read the last element from the `icinga:stats` again and check the environment ID). However when doing this, one has to be careful not to use a heartbeat that is newer than the dump signals (i.e. have a new heartbeat in Redis but no dump signals from that environment yet).
4. Change how/what icinga2 writes to Redis so that this can be handled more reliably.

Given that changing the environment should be a rather rare operation, I think it's hard to justify putting much time into option 3 or even 4. I'm not sure about the other two options. Option 1 requires extra configuration (but only if you actually want to use an environment name other than the empty string, which should be pretty rare anyways). This won't make it much more reliable, it just allows to blame a misconfiguration for problems due to environment changes.

julianbrost:
Discussing this in our team meeting today resulted in the conclusion that we don't want to go the (for now) easy way and bail out when the environment changes, but instead put in the extra effort required to get this handled properly.

PR #337, opened 2021-08-05
Don't print stack on CLI parsing failure or --help

### LINKED ISSUE #334: `icingadb --help` should not display a stack trace at the end
When running `icingadb --help` on the current master, the output looks like this:

```
Usage:
  icingadb [OPTIONS]

Application Options:
  -c, --config=  path to config file (default: ./config.yml)
      --datadir= path to the data directory (default: ./)

Help Options:
  -h, --help     Show this help message

panic: can't parse CLI flags: Usage:
  icingadb [OPTIONS]

Application Options:
  -c, --config=  path to config file (default: ./config.yml)
      --datadir= path to the data directory (default: ./)

Help Options:
  -h, --help     Show this help message


goroutine 1 [running]:
github.com/icinga/icingadb/pkg/utils.Fatal(...)
	/home/jbrost/dev/icingadb/pkg/utils/utils.go:181
github.com/icinga/icingadb/internal/command.New(0x7d7300)
	/home/jbrost/dev/icingadb/internal/command/command.go:21 +0x33b
main.run(0x0)
	/home/jbrost/dev/icingadb/cmd/icingadb/main.go:35 +0x6b
main.main()
	/home/jbrost/dev/icingadb/cmd/icingadb/main.go:31 +0x25
exit status 2
```

There should be no stack trace at the end as it is pointless and confusing.

Al2Klimov:
fixes #334


Al2Klimov:
## Purrs

```
$ go run ./cmd/icingadb -x
unknown flag `x'
```

julianbrost (changes requested):
Now swallows the error message you should receive with `icingadb -c /does/not/exist`

Al2Klimov:
Is this still the case with the current master?

internal/command/command.go
```diff
@@ -18,7 +19,7 @@ type Command struct {
 func New() *Command {
 	flags, err := config.ParseFlags()
 	if err != nil {
-		utils.Fatal(err)
+		os.Exit(2)
```
julianbrost, line 22: Why exit code 2? If you want to have a different exit code for every exit reason, I think it would make sense to have all of them as an enum/iota const. But I'm not sure if that's even worth the effort.

julianbrost (commented):
> Is this still the case with the current master?

No, after merging th current master, that error is still shown. However, if `config.ParseFlags()` would return other errors, they would still be swallowed. It doesn't right now but we have to ensure that it stays this way.

internal/command/command.go
Al2Klimov: Bad CLI args = exit code 2. We shouldn’t re-invent this wheel.

Al2Klimov:
> However, if `config.ParseFlags()` would return other errors, they would still be swallowed.

No, they'd be printed.

internal/command/command.go
julianbrost: Does it? [citation needed]

julianbrost:
> > However, if `config.ParseFlags()` would return other errors, they would still be swallowed.
> 
> No, they'd be printed.

How would `os.Exit(2)` print them? I'm not talking about errors originating by the flag parser but that **we** must not any other new error conditions in **our** `config.ParseFlags()` function. Right now we don't, so that's fine, we just have to keep that in mind.

internal/command/command.go
Al2Klimov: https://github.com/golang/go/blob/1f9c9d853067635305f72e247c5b49e3fa5da8af/src/flag/flag.go#L317

julianbrost: The current implementation does not follow that convention as it also exits with code 2 when the help text is printed. Anyways, this is only the behavior of the Go flag package, not some standard (or even common AFAICT) behavior, so I'd say it's valid to discuss this.

I'd at the very least add a comment that this follows the exit code convention of the Go flag package (if it actually does), or simply go with exit code 1 as probably nobody will check this exit code anyways.

julianbrost (changes requested):

Al2Klimov:
Btw. it’s not a Go-only thing: https://www.baeldung.com/linux/status-codes#misuse-shell-builtin

julianbrost (approved):

julianbrost:
For sure you will find other software that uses the same exit code in this case. But if I take a random program from `/usr/bin` and run it with `--help` and `--foobar`, I wouldn't expect to get exit code 0 and 2 in most cases.

PR #338, opened 2021-08-05
Support --version

### LINKED ISSUE #335: Support `icingadb --version` again
Icinga DB 1.0.0 RC1 supported a `--version` command line switch whereas the current master does not. This should be added again.

Al2Klimov:
fixes #335

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -1,12 +1,15 @@
 package command
 
 import (
+	"fmt"
+	"github.com/icinga/icingadb/internal"
 	"github.com/icinga/icingadb/pkg/config"
 	"github.com/icinga/icingadb/pkg/icingadb"
 	"github.com/icinga/icingadb/pkg/icingaredis"
 	"github.com/icinga/icingadb/pkg/utils"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
+	"os"
 )
 
 type Command struct {
@@ -19,6 +22,11 @@ func New() *Command {
 	flags, err := config.ParseFlags()
 	if err != nil {
 		utils.Fatal(err)
+	}
+
+	if flags.Version {
+		fmt.Println("Icinga DB version:", internal.Version)
+		os.Exit(0)
 	}
 
 	cfg, err := config.FromYAMLFile(flags.Config)
--- a/internal/version.go
+++ b/internal/version.go
@@ -1,0 +1,3 @@
+package internal
+
+var Version = "v1.0.0-rc2-dev"
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -65,9 +65,5 @@ func ParseFlags() (*Flags, error) {
 		return nil, errors.Wrap(err, "can't parse CLI flags")
 	}
 
-	if err := ValidateFile(f.Config); err != nil {
-		return nil, errors.Wrap(err, "invalid config file "+f.Config)
-	}
-
 	return f, nil
 }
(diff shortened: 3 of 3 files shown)


Al2Klimov:
## Purrs

```
$ go run ./cmd/icingadb --version
Icinga DB version: v1.0.0-rc1
```

## Still purrs

```
$ go run ./cmd/icingadb   
panic: can't open YAML file ./config.yml: open ./config.yml: no such file or directory
```

lippserd (changes requested):
I opt for moving version.go to `internal`. Also it would be cool if we could inject the version (and build time) on go build [1]. Does this work with a const?

[1] https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications

Al2Klimov:
Directly quoting the website you linked:

> The value cannot be a const

But... does that matter?

lippserd:
> Directly quoting the website you linked:
> 
> > The value cannot be a const
> 
> But... does that matter?

Is it possible to override the version now?

Al2Klimov:
Not yet. I'd have to s/const/var/. Shall I do?

lippserd:
> Not yet. I'd have to s/const/var/. Shall I do?

Yes, please.

julianbrost:
What do you think of adding some `-dev` suffix in the source, that you still get a somewhat meaningful version when using `go run` or `go build` yourself but still is distinguishable from (release) packages?

Al2Klimov:
## Purrs

```
$ go run -ldflags '-X github.com/icinga/icingadb/internal.Version=LOLCAT' ./cmd/icingadb --version
Icinga DB version: LOLCAT
```

Al2Klimov:
> when using `go run` or `go build` yourself

... we know that we’ve just used it by ourselves.

julianbrost:
Yes, we do. But we also ask for the output of `--version` in bug reports and then we have to rely on the user telling us that they actually built it themselves (maybe they did a month ago to work around another bug and forgot about that). It's just about reducing the chance of having build that look like they are released versions when they aren't (yes, of course you can also deliberately set the version to `1.0.0`, but we can't protect against everything).

lippserd:
I also think `v1.0.0-rc2-dev` makes more sense.

internal/command/command.go
```diff
@@ -21,6 +24,11 @@ func New() *Command {
 		utils.Fatal(err)
 	}
 
+	if flags.Version {
+		fmt.Println("Icinga DB version:", internal.Version)
+		os.Exit(0)
```
lippserd, line 29: In our `main.go` we have two constants for the exit code. The logging PR also calls `os.Exit()` outside of main after a fatal error has been logged. I think we should do this consistently and either use these constants everywhere or drop them. I opt for moving the constants to a new package, e.g. `exitcode`. @julianbrost Since you introduced these constants, what's your opinion here?

julianbrost: I added them because otherwise `int run()` would have magic `return 0;` statements without an indication what this number is supposed to mean. They were inspired by the `EXIT_SUCCESS`/`EXIT_FAILURE` constants you have in C.

So from this point of view, just using `os.Exit(0)` here is fine, as it's pretty clear that it's an exit code.

Yet another possibility would be to have an `type ExitCode int` and use `return ExitCode(0);` instead.

I don't have a strong opinion towards any of these options, but I'd prefer one where there are `return 0;` statements without any indication that it's an exit code.

lippserd: I see. Makes sense!

lippserd (approved):

julianbrost (approved):

PR #339, opened 2021-08-05
Enlarge {comment,downtime}#name 255 -> 548

### LINKED ISSUE #273: Comment and downtime name columns too small
Adding a comment or downtime for a host/service with a long name will crash Icinga DB because the value doesn't fit into the column.

Host and service names have currently a limit of 255 chars.

* For host comments/downtimes, the respective name column needs to be that big plus the id length
* For service comments/downtimes, the respective name column has to be 255*2 chars long plus the id length

Al2Klimov:
I.e. large enough for "host.name!service.name!UUID".

fixes #273

### DIFF
--- a/schema/mysql/schema.sql
+++ b/schema/mysql/schema.sql
@@ -551,7 +551,7 @@ CREATE TABLE comment (
 
   name_checksum binary(20) NOT NULL COMMENT 'sha1(name)',
   properties_checksum binary(20) NOT NULL,
-  name varchar(255) NOT NULL,
+  name varchar(548) NOT NULL COMMENT '255+1+255+1+36, i.e. "host.name!service.name!UUID"',
 
   author varchar(255) NOT NULL COLLATE utf8mb4_unicode_ci,
   text text NOT NULL,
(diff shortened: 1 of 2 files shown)


lippserd (commented):
For services this should be `255+1+255+1+36` or am I missing something?

Al2Klimov:
`service#name` is 255 large.

nilmerg:
But it's `host#name!service#name` for service downtimes and comments, isn't it?

Al2Klimov:
I’m afraid... yes.

https://github.com/Icinga/icinga2/blob/5af40fa5330460ef519bf8a79d4f49f5d11dc66b/lib/icingadb/icingadb-objects.cpp#L1253-L1258

🤦‍♂️ 🐘

lippserd (changes requested):
Please use host_name and service_name in the comment.

Al2Klimov:
I kept it consistent with the other comments.

lippserd:
> I kept it consistent with the other comments.

I see. Then it should be host.name and service.name. Sry!

lippserd (approved):

PR #344, opened 2021-08-10
Schema: Add description to downtime columns 'triggered_by_id' and 'parent_id'

N-o-X:
This PR adds improved comments to columns 'triggered_by_id' and 'parent_id' to better tell them apart.


schema/mysql/schema.sql
```diff
@@ -573,8 +573,8 @@ CREATE TABLE downtime (
   id binary(20) NOT NULL COMMENT 'sha1(environment.name + name)',
   environment_id binary(20) NOT NULL COMMENT 'environment.id',
 
-  triggered_by_id binary(20) DEFAULT NULL COMMENT 'downtime.id',
-  parent_id binary(20) DEFAULT NULL COMMENT 'downtime.id',
+  triggered_by_id binary(20) DEFAULT NULL COMMENT 'The ID of the downtime that triggered this downtime. This is set when creating downtimes on a host higher up in the dependency chain using the "child_option" "DowntimeTriggeredChildren" and can also be set manually via the API.',
```
lippserd, line 576: Icinga Web does not provide these options for service downtimes but via the API it's possible, right? I would then write **on a host or service**.

schema/mysql/schema.sql
```diff
@@ -573,8 +573,8 @@ CREATE TABLE downtime (
...
+  parent_id binary(20) DEFAULT NULL COMMENT 'The ID of this downtimes parent downtime. This is set by using "all_services" while creating host downtimes via the API',
```
lippserd, line 577: I would rephrase this to the following:

> For service downtimes, the ID of the host downtime that created this downtime by using the "all_services" flag of the schedule-downtime API.

lippserd (changes requested):
Please also adjust the upgrade script.

lippserd (approved):

PR #345, opened 2021-08-11
Move etc/icingadb.service to packaging repositories

N-o-X:
We've decided that we don't want to manage service and config files for our packages inside of the source directory but in our packaging repositories instead.


julianbrost:
What's the state of this on the packaging side? Was the service file added there so that this can be merged?

N-o-X:
Yes, that's all done:
https://git.icinga.com/packaging/rpm-icingadb/-/commit/b387a62fa50cd429b0cfd59a8acab7a1024674b4
https://git.icinga.com/packaging/deb-icingadb/-/commit/d9535c6bb5b552a5c2cee3705e740fd708f325b4

PR #346, opened 2021-08-11
Make sure to set default values for config options

lippserd:
If no option is specified in the configuration,
UnmarshalYAML is not called for the Options structs and
therefore no default values are set.

In the config structs we now check whether Options is nil,
which means that no options are set in the configuration.
Then we call unmarshal ourselves to trigger the
Options.UnmarshalYAML call to set the default values.

Alternatively, we could also call defaults.Set() here, but then,
when changing the UnmarshalYAML functions,
we would also have to check here whether changes need to be made.

### DIFF
--- a/pkg/config/database.go
+++ b/pkg/config/database.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"fmt"
-	"github.com/creasty/defaults"
 	"github.com/icinga/icingadb/internal"
 	"github.com/icinga/icingadb/pkg/driver"
 	"github.com/icinga/icingadb/pkg/icingadb"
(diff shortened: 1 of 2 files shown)


pkg/config/database.go
```diff
@@ -47,5 +48,26 @@ func (d *Database) Open(logger *zap.SugaredLogger) (*icingadb.DB, error) {
...
+	self := (*T)(d)
+
+	if err := unmarshal(&self); err != nil {
+		return internal.CantUnmarshalYAML(err, d)
+	}
+	if self.Options == nil {
+		// Options is nil if no option is set in the configuration.
+		// in order for the default values to be set.
+		// We have to call unmarshal ourselves to trigger the Options.UnmarshalYAML call
+		if err := unmarshal(&self.Options); err != nil {
```
julianbrost, line 67: This shouldn't call `unmarshal()` here as this will decode the whole `database:` block into the `options:` sub-block, i.e. it now also uses this value (without the `options:` sub-key):
```yaml
database:
  max_connections: 42
```

julianbrost (changes requested):

PR #348, opened 2021-08-12
Integration tests

julianbrost:
This PR starts implementing integration tests for Icinga DB. It spawns Docker containers for Icinga DB, Icinga 2, Redis and MySQL and then runs test cases written using the Go `testing` package against these services. These tests make use of the new [icinga-testing](https://github.com/Icinga/icinga-testing) repo that contains all the code for spawning these Docker containers as required. This exists as a separate repo so that it can also be used for other projects in the future (for example testing Icinga 2 cluster functionality).

There are quite a few TODO items in the code, mostly showing the need that many individual tests still need to be implemented. But I think it makes no sense waiting for all tests to be implemented before merging this but instead merge what exists now, have it run for each PR and add more tests over time. Currently implemented tests are:
* Object sync: Start Icinga 2 with a config having Host/Service/User objects with all kinds of attribute variations and check that all attributes are properly written to MySQL
* Runtime updates: create/update/delete Service and User objects using the Icinga 2 API and check that these changes make it to the MySQL database
* Multiple environments: test that multiple Icinga 2 + Icinga DB instances can write to the same database and create their own environment each (test currently skipped until #292 is fixed)


N-o-X (changes requested):
Besides the one question below, I really like your approach on this. Tests are easy to write, understand and maintain and, thanks to the great work in `icinga-testing`, don't need a lot of preparation.

Good job! 👍

julianbrost:
I've changed this PR to draft state as #358 will be updated and therefore this PR will have to change before merging. But you can have a look at is nonetheless as I plan no more code changes to the tests but just wait until #358 is merged and then rebase this PR on top of the new master removing the config defaults commit from this PR.

N-o-X (approved):

PR #358, opened 2021-08-25
Ensure extra config options are properly initialized

julianbrost:
YAML is decoded by the structure of the YAML source document, not the Go destination data structure. Therefore, the old code did not always call `UnmarshalYAML()` on all sub-structs. Therefore, defaults were not always set but zero values were used, resulting in all kind of strange behavior.

This commit changes the code so that it no longer relies on individual `UnmarshalYAML()` functions to set the defaults for each sub-struct but instead just sets all of them when creating the surrounding Config instance. It also moves the config validation to separate Validate() functions.

### Tests

<details>
<summary>Extra patch to show parsed config</summary>

```diff
diff --git a/cmd/icingadb/main.go b/cmd/icingadb/main.go
index daf7461..dcd8fbf 100644
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -38,6 +38,7 @@ func run() int {
        defer logger.Sync()
 
        logger.Info("Starting Icinga DB")
+       logger.Debugw("Loaded config", zap.Any("config", cmd.Config))
 
        db := cmd.Database()
        defer db.Close()
```
</details>

Defaults are now set even on an empty config:
```console
% go run ./cmd/icingadb -c <(printf ---) 
2021-08-25T17:10:30.238+0200	INFO	icingadb/main.go:40	Starting Icinga DB
2021-08-25T17:10:30.238+0200	DEBUG	icingadb/main.go:41	Loaded config	{"config": {"Database":{"Host":"","Port":0,"Database":"","User":"","Password":"","Options":{"MaxConnections":16,"MaxConnectionsPerTable":8,"MaxPlaceholdersPerStatement":8192,"MaxRowsPerTransaction":8192}},"Redis":{"Address":"","Password":"","Options":{"Timeout":30000000000,"MaxHMGetConnections":4096,"HMGetCount":4096,"HScanCount":4096}}}}
```

MySQL options are still validated:
```console
$ go run ./cmd/icingadb -c <(printf 'database:\n  options:\n    max_connections: 0') 
panic: invalid configuration: max_connections cannot be 0. Configure a value greater than zero, or use -1 for no connection limit
```

Redis options are still validated:
```console
$ go run ./cmd/icingadb -c <(printf 'redis:\n  options:\n    hmget_count: 0') 
panic: invalid configuration: hmget_count must be at least 1
```

closes #346

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"github.com/creasty/defaults"
 	"github.com/jessevdk/go-flags"
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v3"
@@ -11,6 +12,17 @@ import (
 type Config struct {
 	Database Database `yaml:"database"`
 	Redis    Redis    `yaml:"redis"`
+}
+
+// Validate checks constraints in the supplied configuration and returns an error if they are violated.
+func (c *Config) Validate() error {
+	if err := c.Database.Validate(); err != nil {
+		return err
+	}
+	if err := c.Redis.Validate(); err != nil {
+		return err
+	}
+	return nil
 }
 
 // Flags defines CLI flags.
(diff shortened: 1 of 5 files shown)


lippserd (changes requested):
I like this approach. But the `Validate()` methods lack documentation.

lippserd (approved):

PR #361, opened 2021-09-02
Redis HYield: remove duplicates returned by HSCAN

### LINKED ISSUE #349: Duplicate key errors after HA restart oder handover/takeover
Immediate takeovers after calling `cancelHaCtx` which happens after HA handovers or restart (#327) may result in duplicate key errors. This might happen for any type and is solely related to the config sync. To my knowledge, this does only happen if the config sync before calling `cancelHaCtx` did not finish. It seems that there can be multiple config syncs running at the same time.

julianbrost:
The first iteration of the PR handled this in the delta algorithm, this iteration simply discards duplicates right when reading from Redis.

<details>
<summary>Original PR description for reference</summary>

Initially I hoped that one could simply read both channels until the end, just writing the items in a map each (thereby removing duplicates) would suffice, but my tests showed that with a million of objects, this already takes over a second so I think it makes sense to keep the approach to compute the delta on the fly while receiving the data.

The new implementation is now single-threaded, eliminating the need for locking. My tests showed that this gives a slight penalty when only considering the wall clock time from start to finish (about 20% to 30% in my tests), however, the CPU time even decreased, so if enough delta computations would run to use all available CPU resources, this implementation should even perform better. In reality, I'd expect reading the data from Redis and MySQL will be the limiting factor anyways, so there probably won't be a big impact, this implementation should just use a little less CPU.

I am aware that the implementation is somewhat complex, so I've tried to add quite some comments to help understanding it. I also have other implementations lying around that are easier to understand, but have their own drawbacks:
* Fully read both channels until closed, then the delta can be computed with a few simple set operations (which have to be written out in loops since we're using Go): adds extra latency as the computation can only start once everything is received.
* Still perform the computation on the fly but keep more state to make the on the fly updates easier to understand: has an additional set for IDs received from both streams so that for each ID, both the actual and desired value are known and the ID can be added to/removed from the update set depending on whether their Checksums match. This implementation takes more RAM (additional map to keep) and CPU (more map operations with more maps).

In case this implementation is deemed too complex for those benefits, we can also discuss using one of the others.
</details>

fixes #349

### DIFF
--- a/pkg/icingaredis/client.go
+++ b/pkg/icingaredis/client.go
@@ -70,6 +70,8 @@ func (c *Client) HYield(ctx context.Context, key string) (<-chan HPair, <-chan e
 	return pairs, com.WaitAsync(contracts.WaiterFunc(func() error {
 		defer close(pairs)
 
+		seen := make(map[string]struct{})
+
 		var cnt uint64
 		defer utils.Timed(time.Now(), func(elapsed time.Duration) {
 			c.logger.Infof("Fetched %d elements of %s in %s", cnt, key, elapsed)
@@ -88,6 +90,13 @@ func (c *Client) HYield(ctx context.Context, key string) (<-chan HPair, <-chan e
 			}
 
 			for i := 0; i < len(page); i += 2 {
+				if _, ok := seen[page[i]]; ok {
+					// Ignore duplicate returned by HSCAN.
+					continue
+				}
+
+				seen[page[i]] = struct{}{}
+
 				select {
 				case pairs <- HPair{
 					Field: page[i],


lippserd (changes requested):
I like the removal of the locks but I suggest to handle duplicates right where they originate, i.e. in `HYield`. There is little to no cost in memory usage, for large environments it's even lower. Since `HYield` should be considered a library function it makes sense to not require deduplication in each and every consumer:

# `HYield` changes

```
	dedup := make(chan HPair)
	g.Go(func() error {
		defer close(dedup)

		known := make(map[string]bool)

		for {
			select {
			case pair, ok := <-pairs:
				if !ok {
					return nil
				}

				if _, ok := known[pair.Field]; ok {
					break
				}

				known[pair.Field] = true

				select {
				case dedup <- pair:
				case <-ctx.Done():
					return ctx.Err()
				}
			case <-ctx.Done():
				return ctx.Err()
			}
		}
	})

	return dedup, com.WaitAsync(g)
```

# Memory

This PR:

```
          5588557824  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
             5470084  page reclaims
                3126  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                1015  messages sent
              116652  messages received
                5289  signals received
               18466  voluntary context switches
             4026451  involuntary context switches
          3049406165  instructions retired
          2723933174  cycles elapsed
            16125952  peak memory footprint
```

With dedup in `HYield`:

```
          5137096704  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
             7098216  page reclaims
               26368  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                1075  messages sent
              116728  messages received
                5502  signals received
               26696  voluntary context switches
             3759844  involuntary context switches
          3263071036  instructions retired
          3223771744  cycles elapsed
            15966208  peak memory footprint
```

julianbrost:
That will make a difference in which value is used, the first one or the last one. But that shouldn't really matter as if this happens, that object will receive a runtime update shortly after anyways.

So your suggestion is to keep the change that the delta computation is single-threaded but remove the ability to handle duplicates (should give some improvement in CPU time and memory usage)?

Which delta implementation did you use for your numbers with dedup in `HYield`?

Anyways, can we get #365 done first? Shouldn't be much left to do there and this would avoid creating merge conflicts.

lippserd:
> That will make a difference in which value is used, the first one or the last one. But that shouldn't really matter as if this happens, that object will receive a runtime update shortly after anyways.

True. Also, I'm not exactly sure what happens if you update an already iterated key without changing the size of the hash. As far as I understand the iteration algorithm, it may only produce duplicates if the size of the hash changes.

> So your suggestion is to keep the change that the delta computation is single-threaded but remove the ability to handle duplicates (should give some improvement in CPU time and memory usage)?

Yes, keep the single thread but move deduplication to `HYield`.

> Which delta implementation did you use for your numbers with dedup in `HYield`?

Yours.

> Anyways, can we get #365 done first? Shouldn't be much left to do there and this would avoid creating merge conflicts.

Yes!

julianbrost:
> > That will make a difference in which value is used, the first one or the last one. But that shouldn't really matter as if this happens, that object will receive a runtime update shortly after anyways.
> 
> True. Also, I'm not exactly sure what happens if you update an already iterated key without changing the size of the hash. As far as I understand the iteration algorithm, it may only produce duplicates if the size of the hash changes.

I wouldn't make too many assumptions there, we can only really rely on the [documented behavior](https://redis.io/commands/scan#scan-guarantees) anyways. So when the value for a key is changed, it could probably be in the overall result with any value, both values, or not at all. So this must be covered by a runtime update anyways.

Original spirit of the PR was just to make the delta as robust as possible, with the assumption that the last value we see is probably the most up-to-date one. But that's not something to rely on.

julianbrost:
I've pushed a version that just removes duplicates in `HYield` and will create a second PR with the changes to the delta implementation shortly.

julianbrost:
> will create a second PR with the changes to the delta implementation shortly

And that's #373.

lippserd (approved):
Good idea to use the empty struct here 👍

PR #236, opened 2020-12-09
Support TLS

Al2Klimov:
fixes #188

PR #364, opened 2021-09-15
Feature/history sync foreign keys

julianbrost:


schema/mysql/upgrades/1.0.0-rc2.sql
```diff
@@ -141,3 +141,14 @@ ALTER TABLE service_state
...
     MODIFY long_output mediumtext DEFAULT NULL,
     MODIFY performance_data mediumtext DEFAULT NULL;
+
+ALTER TABLE history
+    ADD CONSTRAINT fk_history_acknowledgement_history FOREIGN KEY (acknowledgement_history_id) REFERENCES acknowledgement_history (id) ON DELETE CASCADE,
+    ADD CONSTRAINT fk_history_comment_history FOREIGN KEY (comment_history_id) REFERENCES comment_history (comment_id) ON DELETE CASCADE,
+    ADD CONSTRAINT fk_history_downtime_history FOREIGN KEY (downtime_history_id) REFERENCES downtime_history (downtime_id) ON DELETE CASCADE,
+    ADD CONSTRAINT fk_history_flapping_history FOREIGN KEY (flapping_history_id) REFERENCES flapping_history (id) ON DELETE CASCADE,
+    ADD CONSTRAINT fk_history_notification_history FOREIGN KEY (notification_history_id) REFERENCES notification_history (id) ON DELETE CASCADE,
+    ADD CONSTRAINT fk_history_state_history FOREIGN KEY (state_history_id) REFERENCES state_history (id) ON DELETE CASCADE;
```
Al2Klimov, line 151: Don't they panic if only one of them refs table XY and the others are NULL?

Al2Klimov, line 151: What's the exact point of ON DELETE CASCADE here?

pkg/icingadb/history/sync.go
```diff
@@ -2,12 +2,15 @@ package history
 
 import (
 	"context"
+	"encoding/json"
```
Al2Klimov, line 5: Let me guess: it makes no sense to review the diff as such, but just the new file?

julianbrost: Yes.

schema/mysql/upgrades/1.0.0-rc2.sql
julianbrost: For details on that commit, you'd have to ask @raviks789, that one is just taken from #259. But from my understanding, it's for #249. That's the reason why we started considering foreign key constraints in the first place.

Al2Klimov: OK, let's assume your guess is right. IYHO: how **exactly** are these FKs supposed to work for history cleanup?

julianbrost: Events are deleted from the details table and should delete the corresponding entries in the `history` table with them (from #249): https://github.com/Icinga/icingadb/blob/201197b187e05beef5bc93a719caa882bfdd752d/pkg/cleanup/cleanup.go#L143-L148

Al2Klimov: Wouldn't it be smarter to reverse that rel-ship to only have to delete from one table and everything else cascades?

julianbrost: The queries I linked only delete related events (i.e. downtime start and end) after both are beyond the deletion threshold. Also not sure how you'd implement this in detail as there's not a 1:1 relation but a 1:N (with N limited to a maximum of 2 though I guess).

Al2Klimov: > delete related events (i.e. downtime start and end) after both are beyond the deletion threshold

OK, then we can't do this:

> delete from one table and everything else cascades

julianbrost: In which scenario? What the PR currently does or what you suggested?

Al2Klimov: I just threw my idea (second quote) away as the more important feature (first quote) blocks it. Nothing to worry about.

pkg/icingadb/history/sync.go
```diff
@@ -32,274 +35,366 @@ func NewSync(db *icingadb.DB, redis *icingaredis.Client, logger *zap.SugaredLogg
...
+	for {
+		timeoutReached := false
+		inputClosed := false
+
+		select {
+		case e, ok := <-input:
+			if ok {
+				ids = append(ids, e.ID)
+				if len(ids) == 1 {
+					timeout = time.After(flushTimeout)
```
Al2Klimov, line 151: If you've not just C&Ped my code, but wrote this more or less from scratch, consider (re-)using an actual timer instead. Rationale: see time.After docs.

pkg/icingadb/history/sync.go
```diff
@@ -32,274 +35,366 @@ func NewSync(db *icingadb.DB, redis *icingaredis.Client, logger *zap.SugaredLogg
...
-		{"state", []v1.UpserterEntity{(*v1.StateHistory)(nil), (*v1.HistoryState)(nil)}},
-		{"downtime", []v1.UpserterEntity{(*v1.DowntimeHistory)(nil), (*v1.HistoryDowntime)(nil)}},
-		{"comment", []v1.UpserterEntity{(*v1.CommentHistory)(nil), (*v1.HistoryComment)(nil)}},
-		{"flapping", []v1.UpserterEntity{(*v1.FlappingHistory)(nil), (*v1.HistoryFlapping)(nil)}},
-		{"acknowledgement", []v1.UpserterEntity{(*v1.AcknowledgementHistory)(nil), (*v1.HistoryAck)(nil)}},
-	} {
-		var structifiers []structify.MapStructifier
-		for _, structPtr := range rhs.structPtrs {
-			structifiers = append(structifiers, structify.MakeMapStructifier(reflect.TypeOf(structPtr).Elem(), "json"))
+	baseUUID := uuid.MustParse("235396c3-a525-4806-a3bc-3806c25baaeb")
```
Al2Klimov, line 340: A namespace inside a table? Do you fear a collision with the migration tool?

pkg/icingadb/history/sync.go
```diff
@@ -32,274 +35,366 @@ func NewSync(db *icingadb.DB, redis *icingaredis.Client, logger *zap.SugaredLogg
...
-	return streams
-}()
+		var userNotifications []v1.UpserterEntity
+
+		for _, user := range users {
+			userNotifications = append(userNotifications, &v1.UserNotificationHistory{
+				HistoryTableEntity: v1.HistoryTableEntity{
+					// Generate a deterministic ID based on the other fields.
+					Id: types.UUID{UUID: uuid.NewSHA1(baseUUID, append(append(append([]byte(nil),
+						notificationHistory.EnvironmentId...), notificationHistory.Id.UUID[:]...), user...))},
```
Al2Klimov, line 362: notificationHistory.EnvironmentId? Isn't notificationHistory.Id unique enough?

Al2Klimov: user? The whole user ID? Wouldn't i be enough? (i from `s/for _, user/for i, user/`)

pkg/icingadb/history/sync.go
julianbrost: My idea was to avoid changing the schema here for this PR as we will likely change this when addressing HA + history. If keeping an explicit ID column, I'd have chosen `SHA1(env_ud, notification_history_id, user_id)` but the column is only 16 bytes, so a hash-based UUID is probably the closest I can get. Another option would be to drop the ID column and use a composite primary key over all remaining columns as this is a relation table and the relation is what identifies the row.

But all in all, I'd postpone the final ID choice to the HA + history task and see this as an intermediate solution.

pkg/icingadb/history/sync.go
julianbrost: What would be the benefit? Apart from that you'd have to sort the user list to generate consistent IDs on both instances in a HA setup (if that's what we end up doing). Apart from that, most from the explanation in https://github.com/Icinga/icingadb/pull/364#discussion_r716819291 also applies here.

Al2Klimov: Less stuff to hash. Good for performance and checksums quality.

Al2Klimov (changes requested):
If https://github.com/Icinga/icingadb/pull/364#discussion_r716819291 is the solution, I either want the problem back or – even better: **change the schema**.

Al2Klimov:
... either `s/UUID/SHA1/` or (my preferred solution) keep (generate) actual UUIDs, make notification_history_id+user_id unique and let ON DUPLICATE do its thing.

pkg/icingadb/history/sync.go
julianbrost: SHA1 operates on 512 bit blocks, so it doesn't matter if I put in 448 bits (current code), 352 bits (use 64 bits for uid) or 320 bits (32 bits for uid).

julianbrost:
> I either want the problem back or – even better: **change the schema**.

As I said, we will have to go over history sync + HA anyways which will likely involve some change to the IDs we use. So far we haven't decided on what exactly we want to do here. So making a schema change here just to throw it over in a week feels like a waste of time to me.

Al2Klimov:
Then let's decide that.

julianbrost:
I had hoped that we could unblock #247 and #249 earlier. Both of these don't really care about how exactly the IDs look like but rely on this PR.

Al2Klimov:
Third solution: throw user_notification_history.id away.

lippserd (commented):
All in all, I think that's very well written and documented! 👍

pkg/icingadb/history/sync.go
```diff
@@ -32,274 +35,366 @@ func NewSync(db *icingadb.DB, redis *icingaredis.Client, logger *zap.SugaredLogg
...
 
-		xra := &redis.XReadArgs{
-			Streams: []string{stream, "0-0"},
-			Count:   bulkSize,
-			Block:   10 * time.Second,
-		}
+	xra := &redis.XReadArgs{
+		Streams: []string{"icinga:history:stream:" + key, "0-0"},
+		Count:   bulkSize,
+		Block:   10 * time.Second,
```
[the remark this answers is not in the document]
julianbrost: Most other XReads block indefinitely, non-blocking would be `Block < 0`, but we want to wait for new data, not poll it?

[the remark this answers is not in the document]
lippserd: Shame on me! I'm sorry for mistaking blocking and non-blocking. Actually I wanted to ask why we don't also use `0` here? So, yes, block indefinitely like our other `XRead`s. I'll update my initial comment.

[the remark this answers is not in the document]
julianbrost: I don't see a problem with also blocking indefinitely here, the current parameters are just as they were before. Regarding keep alive, https://github.com/Icinga/icingadb/pull/236#discussion_r713891389 might also be relevant.

pkg/icingadb/history/sync.go
```diff
@@ -32,274 +35,366 @@ func NewSync(db *icingadb.DB, redis *icingaredis.Client, logger *zap.SugaredLogg
...
-			select {
-			case <-ctx.Done():
-				return ctx.Err()
-			case message, ok := <-redis2struct:
-				if !ok {
-					return nil
+// deleteFromRedis is the last stage of the history sync pipeline. It receives history entries from the second to last
+// pipeline stage and then deletes the stream entry from Redis as all pipeline stages successfully processed the entry.
+func (s Sync) deleteFromRedis(ctx context.Context, key string, input <-chan redis.XMessage) error {
+	const logInterval = 20 * time.Second
```
lippserd, line 131: Since we want to more periodic logging, we could make this a configuration option, e.g. `logging.interval` with a default value of 20 seconds.

pkg/icingadb/history/sync.go
```diff
@@ -32,274 +35,366 @@ func NewSync(db *icingadb.DB, redis *icingaredis.Client, logger *zap.SugaredLogg
...
-		for {
-			select {
-			case <-ctx.Done():
-				return ctx.Err()
-			case message, ok := <-redis2struct:
-				if !ok {
-					return nil
+// deleteFromRedis is the last stage of the history sync pipeline. It receives history entries from the second to last
+// pipeline stage and then deletes the stream entry from Redis as all pipeline stages successfully processed the entry.
+func (s Sync) deleteFromRedis(ctx context.Context, key string, input <-chan redis.XMessage) error {
```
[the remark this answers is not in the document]
julianbrost: Hmm, `Bulker` seems to implement mostly the same logic, but operates on `chan interface{}` so I'd either have to use `interface{}` in most places that use `redis.XMessage` right now or have another goroutine that just converts from `chan redis.XMessage` to `chan interface{}`.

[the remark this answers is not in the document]
lippserd: Yeah, that's true. We already introduced `EntityBulker` because of that. Is something like this worth it?

- Introduce `type Bulkable interface {}`
- s/interface{}/Bulkable
- Embed interfaces
- Remove `EntityBulker`
- Use `Bulker` here

But then we would have to use type assertion everywhere which is not so nice.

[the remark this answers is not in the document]
julianbrost: Doesn't change anything, still would need type assertions everywhere. Generics would help.

Al2Klimov (changes requested):

julianbrost:
@Al2Klimov I don't get what you're trying to ask. Can you please rephrase your review in complete sentences?

Al2Klimov:
I mean: didn't we decide to https://github.com/Icinga/icingadb/pull/364#issuecomment-929010876 and to panic() in the ID getter and setter of the respective struct?

julianbrost:
If I remember correctly, we considered doing so but then said that keeping that ID column is the cleaner solution than partially implementing an interface and panic in some of its functions.

Al2Klimov:
Sure, it's less clean. But it is fine. I remember that we finished on just put a panic() there.

lippserd:
> Sure, it's less clean. But it is fine. I remember that we finished on just put a panic() there.

I think we said both solutions are fine.

pkg/icingadb/history/sync.go
```diff
@@ -32,274 +37,359 @@ func NewSync(db *icingadb.DB, redis *icingaredis.Client, logger *zap.SugaredLogg
...
+		}
+
+		var userNotifications []v1.UpserterEntity
+
+		for _, user := range users {
+			userNotifications = append(userNotifications, &v1.UserNotificationHistory{
+				EntityWithoutChecksum: v1types.EntityWithoutChecksum{
+					IdMeta: v1types.IdMeta{
+						Id: utils.Checksum(append(append(append([]byte(nil),
+							notificationHistory.EnvironmentId...), notificationHistory.Id.UUID[:]...), user...)),
```
Al2Klimov, line 354: I'd not include the redundant envId which is already covered by userId. No matter with how large blocks SHA formally works. TBH I'd just XOR the other two ones.

Al2Klimov (commented):
TBH I prefer the panic() solution. But I won't be a (request-changes-)frog on this.

pkg/icingadb/history/sync.go
julianbrost: I can also remove the environment ID here. It's just there because putting it in there trivially ensures that it's in there without having to look at the structure of the other IDs.

Regarding XOR: I don't think this would be a good idea. It might be here, but I'd just stick with the generic solution to producing an ID from multiple values used throughout Icinga DB (which is hashing).

Al2Klimov: Fine.

Al2Klimov (approved):
Absolutely not my preferred solution (see https://github.com/Icinga/icingadb/pull/364#issuecomment-933514547), but tolerable.

pkg/icingadb/history/sync.go
```diff
@@ -32,274 +37,359 @@ func NewSync(db *icingadb.DB, redis *icingaredis.Client, logger *zap.SugaredLogg
...
 
-// insertedMessage represents a just inserted row.
-type insertedMessage struct {
-	// redisId specifies the origin Redis message.
-	redisId string
-	// structType represents the table the row was inserted into.
-	structType reflect.Type
-}
+// redisCount is the number of history entries to read at once (using XREAD) or delete at once (using XDEL).
+const redisCount = 1 << 10
```
lippserd, line 41: We already have values for count `HSCAN` and `HMGET`. We could combine those and this const into just one count.

pkg/icingadb/history/sync.go
```diff
@@ -32,274 +37,359 @@ func NewSync(db *icingadb.DB, redis *icingaredis.Client, logger *zap.SugaredLogg
...
 
-const bulkSize = 1 << 14
+// insertBufSize is the buffer size for the channel used for writing history entities to the database.
+const insertBufSize = 1 << 14
```
lippserd, line 44: I would go with `db.options.MaxPlaceholdersPerStatement` here.

pkg/icingadb/history/sync.go
lippserd, line 354: For the sake of completeness: Please remove the environment ID from the checksum and also from the schema comments.

lippserd (changes requested):

pkg/icingadb/history/sync.go
julianbrost: I've added a new option for now. We could still unify this in a later PR, but I think the intention of these is to have some tuning knobs should the defaults be suboptimal. So it doesn't hurt to have then fine-grained in my opinion.

lippserd (approved):

PR #366, opened 2021-09-20
Child loggers

lippserd:
This PR implements user-configurable child logger levels and the default log level via a new configuration section called `logging`. In addition, each component in Icinga DB now uses a specific child logger, e.g. `ha`, `*-sync`, etc.
With this PR the default log level is set to `INFO`.

- [x] #377 should be merged before this PR.

### DIFF
--- a/config.yml.example
+++ b/config.yml.example
@@ -1,8 +1,28 @@
+# This is the configuration file for Icinga DB.
+
 database:
   host: icingadb
   port: 3306
   database: icingadb
   user: icingadb
   password: icingadb
+
 redis:
   address: redis:6380
+
+logging:
+  # Default logging level. Can be set to 'fatal', 'error', 'warning', 'info' or 'debug'.
+  # If not set, defaults to 'info'.
+  level:
+
+  # Map of component-logging level pairs to define a different log level than the default value for each component.
+  options:
+    database:
+    redis:
+    heartbeat:
+    high-availability:
+    config-sync:
+    history-sync:
+    runtime-updates:
+    overdue-sync:
+    dump-signals:
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -24,3 +24,26 @@ port                     | **Required.** Database port.
 database                 | **Required.** Database database.
 user                     | **Required.** Database username.
 password                 | **Required.** Database password.
+
+## Logging Configuration <a id="configuration-logging"></a>
+
+Configuration of the logging component used by Icinga DB.
+
+Option                   | Description
+-------------------------|-----------------------------------------------
+level                    | **Optional.** Specifies the default logging level. Can be set to `fatal`, `error`, `warning`, `info` or `debug`. Defaults to `info`.
+options                  | **Optional.** Map of component name to logging level in order to set a different logging level for each component instead of the default one. See [logging components](#logging-components) for details.
+
+### Logging Components <a id="logging-components"></a>
+
+Component                | Description
+-------------------------|-----------------------------------------------
+database                 | Database connection status and queries.
+redis                    | Redis connection status and queries.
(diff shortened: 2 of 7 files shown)


doc/03-Configuration.md
```diff
@@ -35,4 +59,6 @@ database:
   password: icingadb
 redis:
   address: redis:6380
+logging:
+  level: info
```
Al2Klimov, line 63: Some overrides (options) would be nice to have here.

Al2Klimov, line 63: And config.yml.example?

Al2Klimov (changes requested):

doc/03-Configuration.md
lippserd: `config.yml` is subject to remove.

Al2Klimov: * #375

lippserd: There is #377 now and I updated the `config.yml.example`.

config.yml.example
```diff
@@ -1,8 +1,28 @@
...
+
 redis:
   address: redis:6380
+
+logging:
+  # Default logging level. Can be set to 'fatal', 'error', 'warning', 'info' or 'debug'.
+  # If not set, defaults to 'info'.
+  level:
+
+  # Map of component-logging level pairs to define a different log level than the default value for each component.
```
Al2Klimov, line 18: Do we want to duplicate our config docs here?

config.yml.example
```diff
@@ -1,8 +1,28 @@
...
+  options:
+    database:
+    redis:
+    heartbeat:
+    high-availability:
+    config-sync:
+    history-sync:
+    runtime-updates:
+    overdue-sync:
+    dump-signals:
```
Al2Klimov, line 28: I miss example levels here. (Only a complete example is a good example.)

config.yml.example
lippserd: Why not?

config.yml.example
lippserd: Example levels are already document above.

config.yml.example
Al2Klimov: For the same reason why you've removed the example config from the docs. Because... it's a duplicate.

lippserd: So you want to remove the comments? If so, I'm sorry, no.

config.yml.example
Al2Klimov: And in the docs. And all of the other options are also documented in the docs. Nevertheless there are example values here like icingadb or 3306.

lippserd: This file should be self-explanatory or it will be useless. So I don't really get your point.

Al2Klimov: I agree and my point is to make it a bit more self-explanatory, i.e. `s/level:/level: info/`, etc..

config.yml.example
Al2Klimov: Because of https://github.com/Icinga/icingadb/pull/366#discussion_r717630965 ?

config.yml.example
lippserd: That would be counterproductive as it would override our built-in default mechanics.

Al2Klimov: ... if users just C&P the file?

config.yml.example
lippserd: I don't want to look into the docs. So there are some comments.

config.yml.example
lippserd: I think this is a use case of this file. Of course, we should update the connection configuration to sane defaults prior to the release.

Al2Klimov: Damn.

Al2Klimov (approved):

PR #367, opened 2021-09-21
schema: Add indices needed for detail histories in Web

nilmerg:
refs https://github.com/Icinga/icingadb-web/issues/265#issuecomment-909197010


nilmerg:
@lippserd This speeds up the detail histories significantly, but will mess up the main history. It ignores `idx_history_event_time` and uses now `idx_history_host_service_id_object_type_event_time` instead. If I add an index hint (`FROM history USE INDEX (idx_history_event_time)`), the main history is back to normal speeds but I don't know if that's the only way to solve this. Please take a look here. You'll need https://github.com/Icinga/icingadb-web/pull/364 for this.

PR #368, opened 2021-09-22
On shutdown: give up HA handover after 3s, not 5m

### LINKED ISSUE #350: Icinga DB does not exit when reconnecting to the database
At exit, Icinga DB will remove its instance from the database. This is done with an unbound ctx which will try to reconnect for 5 minutes even if Icinga DB was requested to shut down.

# Steps to reproduce

- Stop database service.
- Interrupt or kill Icinga DB.

Al2Klimov:
... not to hang for 5m if DB is down.

fixes #350

### DIFF
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -19,6 +19,7 @@ import (
 	"os/signal"
 	"sync"
 	"syscall"
+	"time"
 )
 
 const (
@@ -67,9 +68,17 @@ func run() int {
 
 	heartbeat := icingaredis.NewHeartbeat(ctx, rc, logger)
 	ha := icingadb.NewHA(ctx, db, heartbeat, logger)
+
 	// Closing ha on exit ensures that this instance retracts its heartbeat
 	// from the database so that another instance can take over immediately.
-	defer ha.Close(context.Background())
+	defer func() {
+		// Give up after 3s, not 5m (default) not to hang for 5m if DB is down.
+		ctx, cancelCtx := context.WithTimeout(context.Background(), 3*time.Second)
+
+		ha.Close(ctx)
+		cancelCtx()
+	}()
+
 	s := icingadb.NewSync(db, rc, logger)
 	hs := history.NewSync(db, rc, logger)
 	rt := icingadb.NewRuntimeUpdates(db, rc, logger)
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -69,7 +69,7 @@ func run() int {
 	ha := icingadb.NewHA(ctx, db, heartbeat, logger)
 	// Closing ha on exit ensures that this instance retracts its heartbeat
 	// from the database so that another instance can take over immediately.
-	defer ha.Close()
+	defer ha.Close(context.Background())
 	s := icingadb.NewSync(db, rc, logger)
 	hs := history.NewSync(db, rc, logger)
 	rt := icingadb.NewRuntimeUpdates(db, rc, logger)
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -61,14 +61,14 @@ func NewHA(ctx context.Context, db *DB, heartbeat *icingaredis.Heartbeat, logger
 	return ha
 }
 
-// Close implements the io.Closer interface.
-func (h *HA) Close() error {
+// Close shuts h down.
+func (h *HA) Close(ctx context.Context) error {
 	// Cancel ctx.
 	h.cancelCtx()
 	// Wait until the controller loop ended.
 	<-h.Done()
 	// Remove our instance from the database.
-	h.removeInstance(context.Background())
+	h.removeInstance(ctx)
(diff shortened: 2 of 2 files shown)


Al2Klimov:
## Before

```
2021-09-22T12:42:07.764+0200  INFO    icingadb/main.go:247    Exiting due to signal   {"signal": "interrupt"}
2021-09-22T12:42:07.765+0200    DEBUG   icingadb/ha.go:263      Removing our row from icingadb_instance {"instance_id": "bbfdb09c922e464cbbffa512d4dc7855"}
...
2021-09-22T12:47:07.765+0200    WARN    icingadb/ha.go:268      Could not remove instance from database {"error": "can't connect to database: can't retry: dial tcp 127.0.0.1:6033: connect: connection refused", "errorVerbose": ...
```

## After

```
2021-09-22T13:02:42.571+0200  INFO    icingadb/main.go:256    Exiting due to signal   {"signal": "interrupt"}
...
2021-09-22T13:02:42.572+0200    DEBUG   icingadb/ha.go:263      Removing our row from icingadb_instance {"instance_id": "777acaa97e424f789c63c62107f7c429"}
...
2021-09-22T13:02:45.577+0200    WARN    icingadb/ha.go:268      Could not remove instance from database {"error": "can't connect to database: can't retry: dial tcp 127.0.0.1:6033: connect: connection refused", "errorVerbose": ...
```

cmd/icingadb/main.go
```diff
@@ -67,9 +68,17 @@ func run() int {
...
 	// Closing ha on exit ensures that this instance retracts its heartbeat
 	// from the database so that another instance can take over immediately.
-	defer ha.Close()
+	{
+		defer func() {
+			ctx, cancelCtx := context.WithTimeout(context.Background(), 3*time.Second)
+			ha.Close(ctx)
+			cancelCtx()
+		}()
+	}
```
julianbrost, line 80: What's the point of these curly braces? That defer statement won't pollute the surrounding scope with any names or something like that.

cmd/icingadb/main.go
```diff
@@ -67,9 +68,17 @@ func run() int {
 
 	heartbeat := icingaredis.NewHeartbeat(ctx, rc, logger)
 	ha := icingadb.NewHA(ctx, db, heartbeat, logger)
+
 	// Closing ha on exit ensures that this instance retracts its heartbeat
 	// from the database so that another instance can take over immediately.
-	defer ha.Close()
+	{
+		defer func() {
+			ctx, cancelCtx := context.WithTimeout(context.Background(), 3*time.Second)
```
julianbrost, line 76: Were 3 seconds chosen arbitrarily or is there some reasoning behind it (i.e. gives time for a certain number of retries or something like that)?

pkg/icingadb/ha.go
```diff
@@ -259,11 +259,11 @@ func (h *HA) realize(s *icingaredisv1.IcingaStatus, t *types.UnixMilli, shouldLo
 	return nil
 }
 
-func (h *HA) removeInstance() {
+func (h *HA) removeInstance(ctx context.Context) {
 	h.logger.Debugw("Removing our row from icingadb_instance", zap.String("instance_id", hex.EncodeToString(h.instanceId)))
 	// Intentionally not using a context here as this is a cleanup task and h.ctx is already cancelled.
```
julianbrost, line 264: That comment is now outdated as some context is used now. IMHO it would also be fine to just create the context with timeout right here.

julianbrost (changes requested):

cmd/icingadb/main.go
Al2Klimov: That was an 🐘.

cmd/icingadb/main.go
Al2Klimov: Arbitrarily.

julianbrost (approved):
I find it a bit strange to pass through a context all the way to implement a timeout (maybe someone can give a second opinion on that), but does the job, so fine for me I guess.

Al2Klimov:
Errm... that's one of its main purposes?

julianbrost:
Yes, but for this you don't have to pass it through all the functions but could also create it right within `removeInstance()`.

cmd/icingadb/main.go
```diff
@@ -67,9 +68,15 @@ func run() int {
 
 	heartbeat := icingaredis.NewHeartbeat(ctx, rc, logger)
 	ha := icingadb.NewHA(ctx, db, heartbeat, logger)
+
 	// Closing ha on exit ensures that this instance retracts its heartbeat
 	// from the database so that another instance can take over immediately.
-	defer ha.Close()
+	defer func() {
+		ctx, cancelCtx := context.WithTimeout(context.Background(), 3*time.Second)
```
lippserd, line 75: I'd like to see a comment, why we have a timeout here.

lippserd (changes requested):
I'd like to see a comment somewhere why we're having ctx with a timeout here.

lippserd (approved):

PR #371, opened 2021-09-22
HA#controller(): fix race condition

### LINKED ISSUE #360: Heartbeat events could be lost
When using a `github.com/icinga/icingadb/pkg/com.Cond` by doing `case <-cond.Wait()` in a select statement where another case could be taken in a loop may potentially lose events.

This happens because the channel returned by `Wait()` may be closed after the receive operation was aborted by another case and the next call to `Wait()` already returns the next channel that was not closed yet.

This problem is demonstrated by the following simple program:
```go
package main

import (
	"context"
	"fmt"
	"github.com/icinga/icingadb/pkg/com"
)

func main() {
	cond1 := com.NewCond(context.Background())
	cond2 := com.NewCond(context.Background())

	go func() {
		for {
			cond1.Broadcast()
			cond2.Broadcast()
		}
	}()

	for i := 0; i < 10; i++ {
		select {
		case <-cond1.Wait():
			fmt.Println("cond1")
		case <-cond2.Wait():
			fmt.Println("cond2")
		}
	}
}
```

Possible output:
```
cond1
cond2
cond1
cond2
cond1
cond1
cond1
cond1
cond1
cond2
```

The `case <-h.heartbeat.Beat()`/`case <-h.heartbeat.Lost()` in `ha.go` was where I noticed this potential issue (`Beat()`/`Lost()` return the result of a `Cond.Wait()`). For `Beat()` this isn't really an issue as this event happens periodically so a beat may be lost but the next one will probably be caught. On the other hand `Loss()` only signals once, so that won't automatically fix itself with the same event a few seconds later. Also the loss should be handled as soon as possible.

I haven't checked if there are other places in the code where this could also be an issue.

Al2Klimov:
... not to loose heartbeat-lost events.

fixes #360

### DIFF
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -113,6 +113,7 @@ func (h *HA) controller() {
 	h.logger.Debugw("Starting HA", zap.String("instance_id", hex.EncodeToString(h.instanceId)))
 
 	oldInstancesRemoved := false
+	var lost <-chan struct{}
 
 	logTicker := time.NewTicker(time.Second * 60)
 	defer logTicker.Stop()
@@ -155,10 +156,16 @@ func (h *HA) controller() {
 			}
 
 			shouldLog = false
-		case <-h.heartbeat.Lost():
+
+			if lost == nil {
+				lost = h.heartbeat.Lost()
+			}
+		case <-lost:
+			lost = nil
 			h.logger.Error("Lost heartbeat")
 			h.signalHandover()
 		case <-h.heartbeat.Done():
+			lost = nil
 			if err := h.heartbeat.Err(); err != nil {
 				h.abort(err)
 			}


Al2Klimov:
Not sure whether I've even correctly conceptionally understood the problem (not to mention reproduction and tests).

pkg/icingadb/ha.go
```diff
@@ -155,10 +156,16 @@ func (h *HA) controller() {
 			}
 
 			shouldLog = false
-		case <-h.heartbeat.Lost():
+
+			if lost == nil {
+				lost = h.heartbeat.Lost()
+			}
```
julianbrost, line 162: This is done too late here, still breaks if at least 60 seconds pass between the heartbeat event and reaching these lines. Moving this right at the top of this case would improve this (receiving from a channel and immediately calling a function probably won't take a minute).

All in all I'd prefer a solution where the heartbeat provided an interface where the obvious way to use it is correct and you don't have to build workarounds in the receiver.

julianbrost (changes requested):
If you rename `cond1`/`cond2` from my issue to `beat`/`lost`, the goroutine simulates a heartbeat component that signals alternating hearbeat received/lost events. In order for the next component (HA) in our case to properly handle this, it has to receive every single one of these events, otherwise it could assume to still be active after a heartbeat loss due to not receiving that event.

Now in our code, there's no loop firing these events as fast as it can, these are only received every few seconds. And there's even a timeout of I think a minute that has to pass between the last beat and the loss event, so I think the only way this can actually happen in our code is when handling of a beat event takes over a minute. But this may happen if the database is gone or very slow.

And while writing this, I figured out that this is actually a way to reproduce the issue:
1. Start Icinga DB with everything else running as normal so that is receives a heartbeat and performs a HA takeover.
2. Stop MySQL.
3. Wait a few seconds (5 should be enough) so that Icinga DB receives another heartbeat and tries to write it to the database (this makes it block within the `case <-h.heartbeat.Beat():` branch)
4. Stop the Icinga 2 instance so that no more heartbeats are written to Redis.
5. Wait for a minute for Icinga DB to loose the heartbeat (it's still trying to write the last one to the database at that point)

You'll then see something like this a the last log message:
```
2021-09-23T09:16:33.902+0200	WARN	icingaredis/heartbeat.go:149	Lost Icinga 2 heartbeat{timeout 8 60000000000  <nil>}
```

Normally, this should immediately be followed by a "Lost heartbeat" message from the HA code, but this doesn't happen as the event was lost (it was still trying to write to the old heartbeat to the database during the time where it would have to listen for loss events to actually receive it). And as it didn't log that message, it won't signal a HA loss as it should have done.

PR #375, opened 2021-09-28
Delete config.yml.example

Al2Klimov:


lippserd (approved):

julianbrost (changes requested):
Why?

Al2Klimov:
https://github.com/Icinga/icingadb/pull/366#discussion_r717414385

julianbrost:
That comment gives no reasoning at all.

lippserd:
> That comment gives no reasoning at all.

That file was only used for packaging but we (should) no longer depend on it. (Same for the systemd service file.)

julianbrost:
The systemd service file got removed in #345 and unfortunately, that PR doesn't provide any reasoning either. I can't remember the reason and can't reconstruct it either. One thing I could have imagined that the service file differs between distributions but not even that's the case right now.

Anyways, I think the example config in the source repo is quite helpful to get people started working on the code. To get Icinga DB running, you need a config file (you can do without a systemd service file, that's why it's valid to make a difference between the two in my opinion), so it's handy to be able to just copy it over.

lippserd:
Let's keep it.

PR #377, opened 2021-09-28
Remove example configuration from docs

lippserd:
We already have config.yml.example as a source to copy and paste
so one can get started easily. In addition,
only one place has to be maintained for the example configuration.

This PR should be merged before #366.


Al2Klimov (changes requested):
Section names are missing in the docs (other construction area), so at least a link to the example config (we decided to keep #375) would be helpful.

lippserd:
> Section names are missing in the docs (other construction area), so at least a link to the example config (we decided to keep #375) would be helpful.

Makes sense.

Al2Klimov (approved):

PR #378, opened 2021-09-28
Systemd journald

lippserd:


pkg/logging/journald_writer.go
```diff
@@ -0,0 +1,122 @@
...
+
+	fields := map[string]string{
+		"SYSLOG_IDENTIFIER": filepath.Base(os.Args[0]),
+	}
+	enc := zapcore.NewMapObjectEncoder()
+	for i := range entry.Fields {
+		entry.Fields[i].AddTo(enc)
+	}
+	for k, v := range enc.Fields {
+		fields[utils.ConvertCamelCase(k, unicode.UpperCase, '_')] = fmt.Sprintf("%v", v)
```
lippserd, line 105: I think we should prefix the fields with `ICINGADB_` in oder to avoid collisions. Any thoughts on that?

julianbrost: Given that it's just whatever you put into `logger.*w()`, sounds like a good idea.

pkg/config/logging.go
```diff
@@ -1,14 +1,38 @@
...
+			// which is the default setting for the Icinga DB service.
+			// This assumes that Icinga DB is running under systemd, so set output to systemd-journald.
+			l.Output = logging.JOURNAL
+		} else {
+			// Otherwise set it to console, i.e. write log messages to stderr.
+			l.Output = logging.CONSOLE
+		}
+	}
+
+	l.Output = logging.JOURNAL
```
lippserd, line 34: That is from debugging. I'll remove it! 😆

julianbrost:
> * Icinga DB assumes that it's running under systemd when the `NOTIFY_SOCKET` environment variable is set. This is done automatically by systemd for `Type=notify` supervised services, which is the default setting for the Icinga DB service. Our packages must take care of that.

The [systemd documentation](https://systemd.io/JOURNAL_NATIVE_PROTOCOL/#automatic-protocol-upgrading) suggests another approach to automatically upgrading stderr logging to journald logging, maybe that's worth to consider implementing.

> * If writing to `systemd-journald` fails, falls back to `console` (stderr).

Is implementing this fallback even useful? I mean if your logging daemon is broken, it's not our fault if logging is broken. Apart from that, if journald is broken, our log messages likely aren't going anywhere as they would eventually be forwarded to journald anyways.

> * Introduces `GopEncoder` for transmitting unmodified log entries to writers, i.e. w/o JSON overhead.

I still don't see why it's necessary to serialize the entries into a byte buffer at all. I've looked a bit at the zap API now and see two options for implementing journald support without that intermediate serialization:
* Implement a [`zapcore.Encoder`](https://pkg.go.dev/go.uber.org/zap/zapcore#Encoder) that encodes the log entries into a message that's understood by the [journald native protocol](https://systemd.io/JOURNAL_NATIVE_PROTOCOL/) and a [`zapcore.WriteSyncer`](https://pkg.go.dev/go.uber.org/zap/zapcore#WriteSyncer) that just sends them to the journald socket.
* Simply implement a [`zapcore.Core`](https://pkg.go.dev/go.uber.org/zap/zapcore#Core) that simply takes the in-memory representation of a log entry and writes it to journald using the package also currently in use by the PR.

lippserd:
Now ready for a full review. The following changes are new since the opening of the PR:

> * Uses the app name (name of the executable that started the program) as the default logger name and as a prefix for the structured logging context that is sent as journal fields.
> * Passes the logger name as SYSLOG_IDENTIFIER so that component logs can also be separated in the systemd journal.

lippserd:
> > * Icinga DB assumes that it's running under systemd when the `NOTIFY_SOCKET` environment variable is set. This is done automatically by systemd for `Type=notify` supervised services, which is the default setting for the Icinga DB service. Our packages must take care of that.
> 
> The [systemd documentation](https://systemd.io/JOURNAL_NATIVE_PROTOCOL/#automatic-protocol-upgrading) suggests another approach to automatically upgrading stderr logging to journald logging, maybe that's worth to consider implementing.

If I'm not mistaken, we would have to set `StandardError=journal`. However, since I would like to have `Type=notify`, I don't see any benefit here.

pkg/logging/journald_core.go
```diff
@@ -0,0 +1,122 @@
...
+	all = append(all, fields...)
+
+	if ent.Level == zapcore.FatalLevel {
+		// Sleep between fatal log and os.Exit() to overcome a race condition that
+		// results in systemd no longer attributing messages from processes
+		// that have exited to their cgroup.
+		// See: https://github.com/systemd/systemd/issues/2913.
+		defer time.Sleep(time.Millisecond * 1100)
+
+		j.loggingFatal.Store(true)
```
julianbrost, line 100: What do you think of additionally having it cancel a root context to also tear down as much as possible and not only hide further log messages?

pkg/logging/journald_core.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	pri, ok := priorities[ent.Level]
+	if !ok {
+		return errors.Errorf("unknown log level %q", ent.Level)
+	}
+
+	journalFields := make(map[string]string, len(all)+1)
+	journalFields["SYSLOG_IDENTIFIER"] = ent.LoggerName
```
julianbrost, line 109: This includes the child logger name resulting in messages like `icingadb.high-availability[4129715]: Removed 0 old instances`. Typically that identifier is just the program name. This also made me miss most log messages at first as I used `journalctl -t icingadb`, which still finds some messages, so that filter looks good at first, but is confusing as messages are missing that you'd expect to be there.

lippserd: True. I wanted to include the child logger name and thought that it's also nice to be able to filter for certain components directly via `-t`. You would have `-u` for all the logs. Anyway, should it be `icingadb [4129715]: high-availability: Removed ...`?

pkg/logging/journald_core.go
```diff
@@ -0,0 +1,122 @@
...
+// For messages under our control,
+// this is actually not an option,
+// as we only want to log to systemd-journald.
+//
+// • Sleep before exiting helps systemd to attribute the message correctly,
+// regardless of whether it is written to systemd-journald or stderr.
+func (j *journaldCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
+	all := make([]zapcore.Field, 0, len(fields)+len(j.context))
+	all = append(all, j.context...)
+	all = append(all, fields...)
```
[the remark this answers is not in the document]
lippserd: > (Also adds quite some dependencies for just encoding messages to a simple format and sending them to a UNIX datagram socket, but I guess that's how software works nowadays)

I know 😢

pkg/logging/journald_core.go
lippserd: Since we wait for more than a second, why not.

PR #382, opened 2021-09-30
Sync state runtime updates ASAP

Al2Klimov:
I.e. don't wait for the complete initial sync first.


Al2Klimov:
## Test protocol

### Machine

* 6 cores
* 8G
* Icinga/icinga2#9018 (not really blocking this one, but w/o that this one is pretty useless)

### Config

```
object Host 1 {
	check_command = "dummy"
	check_interval = 0.5s
	max_check_attempts = 1

	var that = this
	vars.dummy_state = function() use(that) {
		return that.last_check_result && that.last_check_result.state ? 0 : 2
	}
}

for (i in range(1000000)) {
	object User i { }
}
```

### Inspector

```go
package main

import (
	"fmt"
	_ "github.com/go-sql-driver/mysql"
	"github.com/icinga/icingadb/pkg/utils"
	"github.com/jmoiron/sqlx"
	"github.com/jmoiron/sqlx/reflectx"
	"reflect"
	"sync/atomic"
	"time"
)

type hostState struct {
	HardState, Amount uint8
}

func main() {
	db, err := sqlx.Open("mysql", "idb:idb@tcp(10.27.3.164)/idb")
	if err != nil {
		panic(err)
	}

	db.Mapper = reflectx.NewMapperFunc("db", func(s string) string {
		return utils.Key(s, '_')
	})

	lastChange := &atomic.Value{}
	lastChange.Store(&time.Time{})

	go func() {
		var lastStates []hostState
		every01s := time.NewTicker(time.Second / 10)

		for {
			now := <-every01s.C
			var states []hostState

			err := db.Select(
				&states,
				"SELECT hard_state, COUNT(hard_state) amount FROM host_state GROUP BY hard_state ORDER BY hard_state",
			)
			if err != nil {
				panic(err)
			}

			if !reflect.DeepEqual(states, lastStates) {
				lastStates = states
				lastChange.Store(&now)
			}
		}
	}()

	every1s := time.NewTicker(time.Second)
	for {
		if now := <-every1s.C; lastChange.Load().(*time.Time).After(now.Add(-time.Second)) {
			fmt.Print("+")
		} else {
			fmt.Print(".")
		}
	}
}
```

Every character stands for one second:

* .: no state changes
* +: some state changes

### Test event

Icinga 2 reload

### Before

```
+......................................+...........................+
```

### After

```
+....................................+...+
```

Al2Klimov:
Anything more-or-less obvious the changeset could have broken (not necessarily just in the tests)?

julianbrost (commented):
Just took a look at the code in general, looks like it should be fine in general. Haven't done any manual testing so far.

At the moment, the tests manage to crash this PR with the current `icinga2:master`:

```
2021-10-05T13:03:07.877Z	FATAL	no object type for redis key icinga:service:state found
github.com/icinga/icingadb/pkg/icingadb.(*RuntimeUpdates).xRead.func1
	/home/runner/work/icingadb/icingadb/pkg/icingadb/runtime_updates.go:130
	/home/runner/go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57
runtime.goexit
	/opt/hostedtoolcache/go/1.17.1/x64/src/runtime/asm_amd64.s:1581
```

Could be that Icinga 2 still writes some state updates into the `icinga:runtime` stream instead of `icinga:runtime:state` as is should, so that might be something to search over there.

> Anything more-or-less obvious the changeset could have broken (not necessarily just in the tests)?

Don't think so, should be fine if Icinga 2 properly separates update between the two streams then it should be fine to handle them independently.

Is there any more review you'd like for now?

Al2Klimov:
Sure, make a complete review (ex. that 🐘-in-the-room of course).

PR #385, opened 2021-10-05
Schema: Prefix command_id with command type (check, event, notification)

N-o-X:
This PR changes the command reference IDs to `{check,notification,event}command_id` to have unified reference IDs (`{tablename}_id`) across our schema where possible.

- PR Icinga 2: https://github.com/Icinga/icinga2/pull/9058
- PR Icinga DB Web: https://github.com/Icinga/icingadb-web/pull/406

TODO:
- [x] Increase Redis schema version when https://github.com/Icinga/icingadb/pull/391 has been merged

### DIFF
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -26,7 +26,7 @@ import (
 const (
 	ExitSuccess                = 0
 	ExitFailure                = 1
-	expectedRedisSchemaVersion = "3"
+	expectedRedisSchemaVersion = "4"
 	expectedDbSchemaVersion    = 2
 )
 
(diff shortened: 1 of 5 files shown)


julianbrost:
Looks like an incompatible change, so maybe this should also increment the Redis schema version, what do you think? But if yes, this PR has to be coordinated with #391.

N-o-X:
> Looks like an incompatible change, so maybe this should also increment the Redis schema version, what do you think? But if yes, this PR has to be coordinated with #391.

Yeah, it should. I'd wait for #391 and change the Redis schema version after that is merged.

lippserd:
Please rebase and test.

lippserd (approved):

PR #388, opened 2021-10-13
Replace Apache 2.0 licensed gopkg.in/yaml.v3 with MIT licensed github.com/goccy/go-yaml

### LINKED ISSUE #381: Our license isn't compatible with our deps' ones
## Describe the bug

As https://github.com/Icinga/icingadb/pull/379/checks?check_run_id=3745669016 reports, right now we're violating our own license by combining the GPLv2 app with a partially Apache2.0 lib.

## To Reproduce

Run `go mod vendor` and read all the licenses in vendor/ (or just trust the linked check).

### Either

Explicitly upgrade our README.md and license headers (which license headers? **ba dumm tss**) GPLv2 -> GPLv2+. (Luckily as https://github.com/Icinga/icingadb/pull/380#event-5381382376 reports **all** contributors have signed a CLA which allows sublicensing.)

### Or

Get rid of some deps (at least `gopkg.in/yaml.v3`).

CC @N-o-X @julianbrost

Al2Klimov:
... not to have GPLv2<->Apache 2.0 (app<->deps) license conflicts.

fixes #381


Al2Klimov:
## TODO

* goccy/go-yaml#260

lippserd:
Please rebase.

lippserd (approved):

julianbrost:
@lippserd Does your review include the changes done in the `github.com/goccy/go-yaml` fork?

lippserd:
> @lippserd Does your review include the changes done in the `github.com/goccy/go-yaml` fork?

Yes. The only thing I noticed is that the detection of the initial value is a little different than in the defaults pkg we are using:

`reflect.Value.IsValid()` vs `reflect.DeepEqual(reflect.Zero(field.Type()).Interface(), field.Interface())`

But I'm not quite sure what the difference is.

By the way, can't we just set the defaults after the YAML is decoded? 😆

Al2Klimov:
What for?

lippserd:
> What for?

We would not have to patch the lib I guess.

Al2Klimov:
We've already patched our Redis lib, so I see no (greater) problem.

lippserd:
> We've already patched our Redis lib, so I see no (greater) problem.

If I'm right, we would only have to move 3 lines of code. What is the problem with that?

julianbrost:
> By the way, can't we just set the defaults after the YAML is decoded? laughing

Do we set defaults for anything where the zero value would be value but not what we want as default?

lippserd:
> > By the way, can't we just set the defaults after the YAML is decoded? laughing
> 
> Do we set defaults for anything where the zero value would be value but not what we want as default?

Both Alex PR and the defaults package rely on zero value detection for applying defaults.

julianbrost:
Yes, but there would be a slight change in behavior: right now, if `foo` has a default and you set `foo: 0` in the config, it will continue using 0 as value whereas with your suggested change, it would use the default.

lippserd:
> Yes, but there would be a slight change in behavior: right now, if `foo` has a default and you set `foo: 0` in the config, it will continue using 0 as value whereas with your suggested change, it would use the default.

That is true!

PR #389, opened 2021-10-13
schema/mysql/upgrades/1.0.0-rc2.sql: don't enforce foreign keys while adding new foreign keys

### LINKED ISSUE #386: Schema migration: remove rows violating the constaints before adding the constraints
The foreign key constraints added by #364 (bab2d80) may be violated by rows that already exist in the database (as the old code did not care about those constraints when inserting). 

The schema upgrade should be able to handle this, either by removing the rows in violation or setting the foreign keys to `NULL`. It's probably appropriate to do the first as the amount of information still left is of limited usefulness (something like "there was a state change, but we don't know anything about old state, new state, or output").

Al2Klimov:
... not to fail with ERROR 1452 (23000) while adding foreign keys
some already present data doesn't comply to, yet.

More radical alternative: NULL the misreferencing columns in not complying rows first.
Even more radical alternative: DELETE not complying rows first.
Neither of these gives the daemon any chance to sync yet missing data after upgrade.

closes #386

### DIFF
--- a/schema/mysql/upgrades/1.0.0-rc2.sql
+++ b/schema/mysql/upgrades/1.0.0-rc2.sql
@@ -148,6 +148,8 @@ ALTER TABLE service_state
 ALTER TABLE user_notification_history
     MODIFY id binary(20) NOT NULL COMMENT 'sha1(notification_history_id + user_id)';
 
+SET FOREIGN_KEY_CHECKS = 0;
+
 ALTER TABLE history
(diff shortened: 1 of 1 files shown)


lippserd:
I suggest removing the violating rows as the information provided is of little to no use.
Entries that were left in the stream before the upgrade are also unproblematic, as Icinga DB ensures that these are inserted into all relevant tables after the upgrade.

PR #394, opened 2021-10-20
RuntimeUpdates#Sync(): force FIFO for config updates

Al2Klimov:
so they don't interfer.


julianbrost (changes requested):
Oh, and as I've just noticed, this needs a rebase now.

Al2Klimov:
## Speed up by

* #400

julianbrost:
> ## Speed up by

What's that supposed to mean?

Al2Klimov:
That it's not tech.ly blocking, but it would make my life a lot easier here.

pkg/icingadb/runtime_updates.go
```diff
@@ -107,13 +202,26 @@ func (r *RuntimeUpdates) Sync(ctx context.Context, factoryFuncs []contracts.Enti
...
 
-		deleted := make(chan interface{})
 		g.Go(func() error {
-			defer close(deleted)
+			defer func(ch chan interface{}) {
+				select {
+				case <-ch:
+				default:
+					close(ch)
+				}
```
julianbrost, line 211: This operation is only safe if there's no concurrent write/close action on the same channel.

pkg/icingadb/runtime_updates.go
```diff
@@ -107,13 +202,26 @@ func (r *RuntimeUpdates) Sync(ctx context.Context, factoryFuncs []contracts.Enti
...
-			return r.db.DeleteStreamed(ctx, s.Entity(), deleteIds, deleted)
+			return r.db.DeleteStreamed(ctx, s.Entity(), deleteIds, deleted, threshold)
 		})
 		g.Go(func() error {
+			defer func(ch chan interface{}) {
+				select {
+				case <-ch:
+				default:
+					close(ch)
+				}
```
julianbrost, line 222: But here is the same operation again.

pkg/icingadb/runtime_updates.go
```diff
@@ -70,22 +73,114 @@ func (r *RuntimeUpdates) Sync(ctx context.Context, factoryFuncs []contracts.Enti
...
+					case <-ctx.Done():
+						return ctx.Err()
+					}
+				}
+			})
+		} else {
+			upserted = make(chan contracts.Entity)
+			deleted = make(chan interface{})
+			upsertedLog = upserted
+			deletedLog = deleted
```
julianbrost, line 146: And if `allowParallel`, then `upsertedLog = upserted`, so both might observe that the channel is not yet closed and attempt to close it resulting in a panic.

julianbrost (changes requested):
In general, I think the code would be simpler if the goroutine that does `counter.Inc()` would simply forward the entity to a second channel if and only if `!allowParallel` so that that second channel can then be used to synchronize the insertion order. Then it would also be clear when which channel has to be closed where.

julianbrost (changes requested):
I think https://github.com/Icinga/icingadb/pull/394/commits/a90dc7a4b28f3cdc38097ae99bc785f1a19188de shouldn't be in this PR (for now this is the only change request, the inline comment is only a question).

pkg/com/bulker.go
```diff
@@ -93,5 +93,33 @@ func (b *Bulker) run(ch <-chan interface{}, count int) {
...
 func Bulk(ctx context.Context, ch <-chan interface{}, count int) <-chan []interface{} {
+	if count <= 1 {
+		return oneBulk(ctx, ch)
+	}
+
 	return NewBulker(ctx, ch, count).Bulk()
 }
+
+// oneBulk operates just as NewBulker(ctx, ch, 1, 0).Bulk(),
+// but without the overhead of handling different values for 1 and 0.
```
julianbrost, line 104: Outdated comment, `NewBulker` only takes 3 arguments.

pkg/com/entity_bulker.go
```diff
@@ -94,5 +94,33 @@ func (b *EntityBulker) run(ch <-chan contracts.Entity, count int) {
...
 func BulkEntities(ctx context.Context, ch <-chan contracts.Entity, count int) <-chan []contracts.Entity {
+	if count <= 1 {
+		return oneEntityBulk(ctx, ch)
+	}
+
 	return NewEntityBulker(ctx, ch, count).Bulk()
 }
+
+// oneEntityBulk operates just as NewEntityBulker(ctx, ch, 1, 0).Bulk(),
+// but without the overhead of handling different values for 1 and 0.
```
julianbrost, line 105: Outdated comment, `NewEntityBulker` only takes 3 arguments.

julianbrost (changes requested):
Apart from the outdated comments: tested and approved.

julianbrost (approved):

PR #395, opened 2021-10-26
Flush logs after config dump

Al2Klimov:
to simplify debugging the exact config dump finished time.


lippserd:
I don't think that's necessary.

PR #397, opened 2021-10-26
Add more MySQL indexes

Yonas-net:
refs https://github.com/Icinga/icingadb-web/issues/404

### DIFF
--- a/schema/mysql/schema.sql
+++ b/schema/mysql/schema.sql
@@ -80,7 +80,9 @@ CREATE TABLE hostgroup (
 
   zone_id binary(20) DEFAULT NULL COMMENT 'zone.id',
 
-  PRIMARY KEY (id)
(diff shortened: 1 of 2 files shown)


Yonas-net:
**Blocked by** https://github.com/Icinga/icinga2/issues/9049

Yonas-net:
Test `history`, `downtime` and `service_state` indices whether they really improve the performance of Icinga DB Web.

julianbrost:
What's the expected benefit for queries? Like is it more like unbearably slow -> acceptable or acceptable -> instant? Was write performance also considered when adding these?

julianbrost:
I've done my standard schema upgrade diff from RC1 and it was fine. So if someone has a look that the actual changes are fine, this could be merged.

N-o-X (approved):

PR #393, opened 2021-10-15
Sync history independent of HA

Al2Klimov:
## Blocked by

* #392

PR #392, opened 2021-10-15
Make {NotificationHistory,StateHistory,History*}#Id UUID -> SHA1

Al2Klimov:
## Blocked by

* Icinga/icinga2#9029

PR #399, opened 2021-10-29
Integration tests for history sync in HA setups

julianbrost:
This PR extends the existing history sync integration tests so that these are also run with an icinga2 HA cluster to check that even in this case, no duplicates are written to MySQL.

This served as my tests for the following pull requests:
* #392 (merged into this PR for testing)
* #393 (merged into this PR for testing)
* https://github.com/Icinga/icinga2/pull/9029 (the `icinga2:git-v2.13.0-77-ga17592fe7` Docker image is a build of that PR merged with the current master (https://github.com/Icinga/icinga2/commit/b5132a59ef72737fd7c3415cf101c5c1ff1c32dc)

### Blocked by
* https://github.com/Icinga/icinga-testing/pull/7

### DIFF
--- a/tests/go.mod
+++ b/tests/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/containerd/containerd v1.5.6 // indirect
 	github.com/go-redis/redis/v8 v8.11.3
 	github.com/google/uuid v1.3.0
-	github.com/icinga/icinga-testing v0.0.0-20211111113205-9914911dc718
+	github.com/icinga/icinga-testing v0.0.0-20211112112017-64c69fdac3ca
 	github.com/jmoiron/sqlx v1.3.4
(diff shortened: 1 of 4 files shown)


julianbrost:
I will probably move `tests/internal/pki/` to the icinga-testing repo and then mark it ready for review.

julianbrost:
Still draft as the test for HA + Notifications fails sometimes for me locally, so I want to have a look at this first.

julianbrost:
Just rebasing on the current master seems to fix that again. At least tests now passed 10 times in a row for me (was more like 50% failure rate before).

julianbrost:
Looks like this was fixed by #408. Reverting 7f3621e980c3d0d7171db5ff2cd2f307a10727d4 makes the problem appear again.

lippserd (approved):

PR #406, opened 2021-11-11
Add a regression test that ensures that runtime updates are executed in order

julianbrost:
The test is supposed to fail here, but once merged together with #394, it should succeed.

### Blocked by
* https://github.com/Icinga/icinga-testing/pull/6 (this PR already updates the `go.mod` file to include this PR here, but it should be merged over there first)


lippserd:
#394 has been merged (which includes this PR). I'll close here.

PR #416, opened 2021-11-23
Docs: Fix foreground command config parameter

N-o-X:
Adds a missing `-` to the config parameter of the foreground command.


lippserd (approved):

lippserd:
We could also just remove it because `/etc/icingadb/config.yml` is the default. What do you think?

PR #418, opened 2021-11-23
Port integration tests to PostgreSQL

julianbrost:
WIP to run integration tests for #221 on PostgreSQL.

### Blocked by
* #221
* https://github.com/Icinga/icinga-testing/pull/9

### Todo
* [x] After #221 was merged, rebase this PR to no longer include the commits from there
* [x] After https://github.com/Icinga/icinga-testing/pull/9 was merged, update this PR to use the merge commit there


.github/workflows/integration-tests.yml
```diff
@@ -28,8 +39,10 @@ jobs:
       - name: Run Integration Tests
         run: ./icingadb-test -icingatesting.debuglog debug.log -test.v
         env:
+          ICINGADB_TESTS_DATABASE_TYPE: ${{ matrix.database.name }}
           ICINGA_TESTING_ICINGADB_BINARY: ${{ github.workspace }}/icingadb
-          ICINGA_TESTING_ICINGADB_SCHEMA: ${{ github.workspace }}/schema/mysql/schema.sql
+          ICINGA_TESTING_ICINGADB_SCHEMA_MYSQL: ${{ github.workspace }}/schema/mysql/schema.sql
+          ICINGA_TESTING_ICINGADB_SCHEMA_PGSQL: ${{ github.workspace }}/schema/pgsql/schema.sql
```
Al2Klimov, line 45: Errm... `schema/${{ matrix.database.name }}`?

tests/environment_test.go
```diff
@@ -14,13 +15,12 @@ import (
 )
 
 func TestMultipleEnvironments(t *testing.T) {
-	m := it.MysqlDatabaseT(t)
-	m.ImportIcingaDbSchema()
+	rdb := getDatabase(t)
```
Al2Klimov, line 18: rdb implies Redis!

Al2Klimov (changes requested):

tests/environment_test.go
julianbrost: What?

.github/workflows/integration-tests.yml
julianbrost: There seems to be no function for converting to uppercase in these expressions. Also I don't know if expressions are even supported in dictionary keys/variable names.

tests/environment_test.go
Al2Klimov: If you grep in the icingadb repo... OK, forget it.

.github/workflows/integration-tests.yml
julianbrost: For the environment variable name. The code in icinga-testing allows creating both database types within the same test case, so these have to be separate variables.

Al2Klimov: Do we need that?

julianbrost: Using the same variable for both the MySQL and PostgreSQL schema import functions wouldn't make code simpler but would break functionality (that we don't need at the moment, to answer you question), just to save one line here.

Al2Klimov (approved):
No more obvious complaints from my side.

julianbrost:
Oh no, looks like there's another missing event to debug: https://github.com/Icinga/icingadb/runs/5555147501?check_suite_focus=true

Let's see how often this happens. But shouldn't be related to PostgreSQL at all.

julianbrost:
6 passes in a row, so doesn't look to happen too often. So please go ahead with the review.

Al2Klimov:
It seems you just rebased, i.e. didn’t change anything substantial, so...

PR #420, opened 2021-11-29
Schema: make notification_history#text longtext

Al2Klimov:


Al2Klimov:
## Blocked by

* https://github.com/Icinga/icingadb/pull/421

schema/mysql/upgrades/1.0.0.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE notification_history
+    MODIFY `text` longtext NOT NULL;
```
julianbrost, line 2: This file should also insert a new row into `icingadb_schema` and update the version check in the code.

Not yet relevant for this PR but for the next one touching the schama: we have to decide if we want to do a version bump per release or per change in the development phase, i.e. should we increase the version with each PR that updates the schema?

julianbrost (changes requested):

schema/mysql/upgrades/1.0.0.sql
Al2Klimov: +1 per release (look at IDO)

lippserd: I would see the `1.0.0.sql` as the baseline and then start with incremental schema upgrades. Conceptually, something like https://flywaydb.org/documentation/concepts/migrations#sql-based-migrations.

lippserd:
Please rebase.

lippserd (changes requested):
The version increment is missing in `schema/mysql/schema.sql`.

lippserd (approved):

Al2Klimov:
PING @julianbrost

julianbrost (approved):

PR #421, opened 2021-11-30
Integration tests: skip intermittently failing test

julianbrost:
The test fails depending on the order in which both Icinga DB instances process history events and is due to an issue in Icinga 2 documented at https://github.com/Icinga/icinga2/issues/9101.


Al2Klimov (commented):
I opt for fixing the problem.

julianbrost:
Sure. But this PR exists to keep the tests more useful in the meantime.

Al2Klimov:
OK, to be more precise: I opt for keeping _the meantime_ as short as possible.

julianbrost:
No longer needed as that issue should be resolved now.

PR #422, opened 2021-12-02
Integration tests: test that both icinga2 instances write identical history streams

julianbrost:
Tests for the following two PRs that extend the history integration tests so that they check if both icinga2 instances in a HA setup write identical history streams. Should only be merged after these are merged to prevent failing tests (d'oh!).

### Blocked by
* https://github.com/Icinga/icinga2/pull/9112
* https://github.com/Icinga/icinga2/pull/9122

refs https://github.com/Icinga/icinga2/issues/9101

### DIFF
--- a/tests/go.mod
+++ b/tests/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/containerd/containerd v1.5.6 // indirect
 	github.com/go-redis/redis/v8 v8.11.3
 	github.com/google/uuid v1.3.0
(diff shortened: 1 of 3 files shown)


lippserd:
We can continue here right?

julianbrost:
[Yes](https://github.com/Icinga/icingadb/pull/422#event-5858414762)

lippserd (approved):

PR #423, opened 2021-12-02
GHA: go mod tidy

Al2Klimov:
If this changes go.*, complain about the non-perfect file(s).

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -63,3 +63,19 @@ jobs:
             fmtvar="$(gofmt -d .)"
             echo "$fmtvar"
             test -z "$fmtvar"
+
+  modtidy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-go@v2
+        with:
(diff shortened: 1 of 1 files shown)


lippserd:
Please add a description.

lippserd:
If I'm not mistaken, dependabot automatically runs `go mod tidy`. So is that really necessary?

Al2Klimov:
Actually I'm not sure. But if **I'm** not mistaken, dependabot isn’t our only contributor.

lippserd:
> But if **I'm** not mistaken, dependabot isn’t our only contributor.

Sure, but dependabot will fix any _mistakes_. So I'd be ok if this isn't perfect until a dependabot PR fixes it. Because why bother a contributor when it will be fixed automatically anyway.

Al2Klimov:
We already bother them w/ fmt, lint and even vet. The go.sum cryptographically secures that we have no surprises. I'd like it to be always-perfect and not to wait for dependabot's mercy. If we can squash external commits, we can also run go mod tidy.

lippserd (approved):

PR #428, opened 2022-01-05
Auto-update honnef.co/go/tools/staticcheck via dependabot

Al2Klimov:
Instead of cloning a particular hardcoded commit,
    put it to the Go mods, so dependabot updates it, too.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -35,7 +35,10 @@ jobs:
 
       - name: Run staticcheck ./...
         run: |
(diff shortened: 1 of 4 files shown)


lippserd:
Please add a description.

lippserd:
Isn't it possible to use https://github.com/marketplace/actions/staticcheck and don't define `version` as it defaults to `latest`?

Al2Klimov:
Every of our deps is pinned to $COMMIT which we actively choose whether to update by merging Dependabot PRs. This way we don’t get any sudden surprises. That 3rd party tool shall behave the same.

lippserd:
But this one is only a dep for github actions. I don't see why it has to be in our `go.mod`.

PR #431, opened 2022-01-11
Run daily integration tests

julianbrost:
As those also implicitly test code in the icinga2 repo, it makes sense to run them periodically to notice if we break something over there.

(Some time at night was chosen by a fair dice roll.)


lippserd (approved):

julianbrost:
Works, did run tonight: https://github.com/Icinga/icingadb/actions/runs/1685923093
5:19 UTC instead of 3:57 UTC, quite some delay but fine.

PR #434, opened 2022-01-28
GHA: use latest staticcheck version

Al2Klimov:
not a hardcoded one.

This action takes care of that.

closes #428


Al2Klimov:
This actually refers to that version.

Al2Klimov:
Just yet another _just to be sure in case the sky falls me on the head_ thing...

lippserd (approved):

PR #435, opened 2022-01-31
History retention

### LINKED ISSUE #163: History cleanup

lippserd:
This PR implements history retention by running a periodic task for each configured history table whose data should only be retained for a specific number of days.

By default, checks for data to be deleted every hour and then deletes a maximum of 5000 rows per round. If the query execution time exceeds 1 second, the count is halved and the query is rerun. The count resets once the query is run within the 1 second grace period.

A user can configure the retention days for each table with a single option, but also enable/override history retention for specific tables. It's also possible configure the count, interval and grace period, but this is undocumented as it is set to sensible defaults like our other configuration options (DB and Redis).

fixes #163
closes #249


pkg/icingadb/cleanup.go
```diff
@@ -0,0 +1,124 @@
...
+
+	affectedRows := make(chan int64, 1)
+	defer close(affectedRows)
+
+	for limit, stop := count, false; !stop; {
+		deleteCtx, cancelDeleteCtx := context.WithCancel(ctx)
+		g, ctx := errgroup.WithContext(ctx)
+
+		g.Go(func() error {
+			stmt := stmt.Build(db.DriverName(), limit)
```
lippserd, line 62: Note to myself before I forget it again: Call `Rebind()`.

pkg/icingadb/cleanup.go
```diff
@@ -0,0 +1,124 @@
...
+				db.logger.Warnf(
+					"Query %q could not be executed within %s. Limit is halved and the query is executed again",
+					stmt.Build(db.DriverName(), limit),
+					gracePeriod,
+				)
+				rounds.Inc()
+				counter.Add(uint64(<-affectedRows))
+
+				limit = limit / 2
+				cancelDeleteCtx()
```
julianbrost, line 106: This looks the wrong way around. If I didn't miss something, cancelling the context doesn't have an effect because the other goroutine only uses it before sending to the channel.

pkg/icingadb/cleanup.go
```diff
@@ -0,0 +1,124 @@
...
+	default:
+		panic("invalid database type")
+	}
+}
+
+// CleanupOlderThan deletes all rows with the specified statement that are older than the given time.
+// Deletes a maximum of as many rows per round as defined in count.
+// If the query execution time exceeds the duration defined in the grace period,
+// the count is halved and the query is rerun.
+// The count is reset once the query has run within the grace period.
```
julianbrost, line 48: Wouldn't this, if the database is never able to delete more than `limit` rows per `grace`, result in always trying a useless query in between?

cmd/icingadb/main.go
```diff
@@ -279,6 +288,21 @@ func run() int {
...
 
+						g.Go(func() error {
+							// Wait for config and state sync to avoid putting additional pressure on the database.
+							configInitSync.Wait()
+							stateInitSync.Wait()
+
+							select {
+							case <-synctx.Done():
+								return synctx.Err()
+							default:
```
julianbrost, line 299: Could be made a simple if. `Err() != nil` iff `Done()` is closed: https://pkg.go.dev/context#Context

pkg/icingadb/history/retention.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// Start starts the retention.
+func (r *Retention) Start(ctx context.Context) error {
+	tasks := make([]periodic.Stopper, 0, len(statements))
+	defer func() {
+		for i := range tasks {
+			tasks[i].Stop()
+		}
+	}()
```
julianbrost, line 60: Could the same also be achieved by cancelling a context passed to `periodic`?

pkg/icingadb/history/retention.go
```diff
@@ -0,0 +1,133 @@
...
+	"downtime": {
+		Table: "downtime_history",
+		PK:    "downtime_id",
+		Where: "start_time < :time AND (end_time IS NOT NULL AND end_time < :time)",
+	},
+	"flapping": {
+		Table: "flapping_history",
+		PK:    "id",
+		Where: "start_time < :time AND (end_time IS NOT NULL AND end_time < :time)",
+	},
```
julianbrost, line 122: Isn't the first part of the checks somewhat redundant as `set_time`/`entry_time`/`start_time` should never be greater than `clear_time`/`remove_time`/`end_time`? This might make a difference on the index used, but I haven't looked at this in any more detail, so the current state might be perfectly fine. Have you considered this?

pkg/periodic/periodic.go
```diff
@@ -25,15 +25,35 @@ type Tick struct {
...
 // OnStop configures a callback that is executed when a periodic task is stopped or canceled.
 func OnStop(f func(Tick)) Option {
 	return optionFunc(func(p *periodic) {
 		p.onStop = f
 	})
 }
 
+// Stable puts the periodic task into stable mode,
+// which guarantees to always wait for the specified interval
+// between task executions.
```
julianbrost, line 44: Wouldn't the overall PR work just as well without this additional flag? Also I find naming this `stable` confusing, is that terminology used anywhere else?

schema/mysql/schema.sql
```diff
@@ -1013,7 +1015,9 @@ CREATE TABLE downtime_history (
   trigger_time bigint unsigned NOT NULL,
   cancel_time bigint unsigned DEFAULT NULL,
 
-  PRIMARY KEY (downtime_id)
+  PRIMARY KEY (downtime_id),
+
+  INDEX idx_downtime_history_start_time_end_time (start_time, end_time) COMMENT 'Filter for history retention'
```
julianbrost, line 1020: As already stated in a previous comment, isn't the check for `start_time` redundant and `end_time` could be used as the only value for filtering and thus would be better suited as the first (or only) value in the index? Same for the folliwing indices.

julianbrost (changes requested):
I'm not fully understanding the 1 second limit. I first read the PR description as if the query doesn't finish in that time, it should be cancelled and retried, but it doesn't seem to be implemented this way. What's the intended behavior there?

A few thoughts in general: 1s sounds somewhat low to me, especially if it's a hard limit. Sure, we probably don't want to have queries running for minutes. And if it's indeed supposed to be a hard limit: aborting the query probably means wasting resources, so I think it might be a good idea to avoid this as best as we can. Maybe something like the following could be implemented without getting to complex: Try do dynamically adjust the number of rows (up and down) within the range 1..limit to aim for a target execution time per query, maybe with a hard deadline being an order of magnitude larger than the target time.

pkg/driver/driver.go
```diff
@@ -75,9 +79,9 @@ func (d Driver) OpenConnector(name string) (driver.Connector, error) {
 	}, nil
 }
 
-// Register makes our database Driver available under the name "icingadb-mysql".
+// Register makes our database Driver available under the name stored in MYSQL.
```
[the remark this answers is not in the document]
lippserd: Reverted the change.

pkg/driver/driver.go
```diff
@@ -14,6 +14,10 @@ import (
 	"time"
 )
 
+const MYSQL = "icingadb-mysql"
+
+const PGSQL = "icingadb-pgsql"
```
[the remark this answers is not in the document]
lippserd: Done.

pkg/icingadb/cleanup.go
```diff
@@ -0,0 +1,124 @@
...
+			case n := <-affectedRows:
+				rounds.Inc()
+				counter.Add(uint64(n))
+
+				if n < int64(limit) {
+					stop = true
+				} else {
+					limit = count
+				}
+			case <-time.After(gracePeriod):
```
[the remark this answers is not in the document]
lippserd: Decided to remove the grace period entirely. We don't have anything like that for config sync either and I really don't think it's necessary.

pkg/icingadb/cleanup.go
lippserd: Removed the grace period (see above).

pkg/icingadb/cleanup.go
lippserd: Removed the grace period (see above).

cmd/icingadb/main.go
lippserd: Done.

pkg/icingadb/history/retention.go
lippserd: Yes, but I would have to call the cancel function in a defer anyway and it shouldn't be done in the for loop. Or do I misunderstand you?

pkg/icingadb/history/retention.go
lippserd: Good catch. Fixed.

pkg/periodic/periodic.go
lippserd: Would work without it too. But the lower the interval, the higher the chance that you'll have consecutive runs that are under the wait time. And I wanted to keep the interval constant. I just chose `stable` like that. We could also go with `constant` or something like that. But it's also fine to just remove it.

schema/mysql/schema.sql
lippserd: Fixed.

pkg/icingadb/history/retention.go
julianbrost: If I'm not missing something,
```go
ctx, cancel := context.WithCancel(ctx)
defer cancel()
```
should do basically the same here (with that new `ctx` passed to `periodic.Start()` below) as is will stop all tasks with one cancel operation. Also looks like you won't even need the `tasks` variable anymore to keep track of them then.

lippserd: Nice, that looks good. I will check it.

julianbrost (commented):
Just two small comments for now. Haven't done a full pass of the new version so far, but I'll continue working on something else for now, so before I keep them as a draft in the meantime, here they are.

pkg/periodic/periodic.go
julianbrost: When the feature is activated initially, this counts as a single run anyways, doesn't it?

So I think the exact details here only affect continuous operation anyways. Averaged over a longer time, you should end up with the same fraction of time the feature is running with both variants. Let's say that fraction is 50% (probably highly exaggerated) and the interval is 1 hours, this would either make it run for 30 minutes every hour or for 1 hour every 2 hours.

I don't see a meaningful difference, so I'd go with the simpler option (that would be using the existing periodically implementation as-is).

Things might look a bit different if there was more control on when exactly the cleanup is running, so that you could move it to the night or weekends. (Please don't take this as a request to implement this, it would probably be overkill.)

lippserd: Removed.

pkg/icingadb/cleanup.go
```diff
@@ -0,0 +1,81 @@
...
+// CleanupOlderThan deletes all rows with the specified statement that are older than the given time.
+// Deletes a maximum of as many rows per round as defined in count.
+func (db *DB) CleanupOlderThan(ctx context.Context, stmt CleanupStmt, count uint64, olderThan time.Time) (CleanupResult, error) {
+	var counter com.Counter
+	defer db.log(ctx, stmt.Build(db.DriverName(), 0), &counter).Stop()
+
+	var rounds uint64
+
+	affectedRows := make(chan int64, 1)
+	defer close(affectedRows)
```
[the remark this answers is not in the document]
lippserd: Removed. Ty!

julianbrost (changes requested):
Just some small things in the actual implementation and a few suggestions for the new test case. I'll probably start testing this soon (at least I don't expect major changes to the implementation anymore).

cmd/icingadb/main.go
```diff
@@ -129,6 +129,14 @@ func run() int {
 	hs := history.NewSync(db, rc, logs.GetChildLogger("history-sync"))
 	rt := icingadb.NewRuntimeUpdates(db, rc, logs.GetChildLogger("runtime-updates"))
 	ods := overdue.NewSync(db, rc, logs.GetChildLogger("overdue-sync"))
+	ret := history.NewRetention(
+		db,
+		cmd.Config.HistoryRetention.Days,
+		cmd.Config.HistoryRetention.Interval,
+		cmd.Config.HistoryRetention.Count,
+		cmd.Config.HistoryRetention.Options,
+		logs.GetChildLogger("history-retention"),
```
julianbrost, line 138: New logger should also be added to the example config.

julianbrost (changes requested):
Things still look fine in general after the first round of testing, just 2 more minor things I noticed.

pkg/icingadb/cleanup.go
```diff
@@ -0,0 +1,78 @@
...
+	PK     string
+	Column string
+}
+
+// Build assembles the cleanup statement for the specified database driver with the given limit.
+func (stmt *CleanupStmt) Build(driverName string, limit uint64) string {
+	switch driverName {
+	case driver.MYSQL, "mysql":
+		return fmt.Sprintf(`DELETE FROM %[1]s WHERE %[2]s < :time ORDER BY %[2]s LIMIT %[3]d`, stmt.Table, stmt.Column, limit)
+	case driver.PGSQL:
```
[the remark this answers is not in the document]
lippserd: This is needed for tests. We don't load our drivers there.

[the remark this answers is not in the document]
julianbrost: Okay, then this will probably also need `"postgres"` soon.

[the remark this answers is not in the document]
lippserd: Yes, should I add it right away?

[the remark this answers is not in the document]
julianbrost: My current plan would be to merge #136, #418 and then this PR. So this one will probably require a quick rebase anyways. Can be done then or also now if you do other changes anyways.

tests/cleanup_and_retention_test.go
```diff
@@ -0,0 +1,200 @@
...
+			_, err := db.Exec(fmt.Sprintf(`CREATE TEMPORARY TABLE %s (event_time bigint(20) NOT NULL, PRIMARY KEY (event_time))`, table))
+			require.NoError(t, err)
+
+			// One row per millisecond.
+			total := scenario.rowsToDelete + scenario.rowsToSpare
+			values := make([]struct{ Time int64 }, 0, MaxPlaceholders)
+			for j := uint64(0); j < total; j++ {
+				values = append(values, struct{ Time int64 }{startMilli + int64(j)})
+
+				if j > 0 && j%(MaxPlaceholders-1) == 0 || j == total-1 {
```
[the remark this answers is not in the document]
lippserd: Yes! 🤣

tests/cleanup_and_retention_test.go
```diff
@@ -0,0 +1,200 @@
...
+			table := fmt.Sprintf("retention_test_limit_%d_delete_%d", scenario.limit, scenario.rowsToDelete)
+
+			db := database(t)
+
+			_, err := db.Exec(fmt.Sprintf(`CREATE TEMPORARY TABLE %s (event_time bigint(20) NOT NULL, PRIMARY KEY (event_time))`, table))
+			require.NoError(t, err)
+
+			// One row per millisecond.
+			total := scenario.rowsToDelete + scenario.rowsToSpare
+			values := make([]struct{ Time int64 }, 0, MaxPlaceholders)
```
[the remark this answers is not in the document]
lippserd: Only with `NamedExec` you can do batch/bulk queries.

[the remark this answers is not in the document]
lippserd: (without constructing the necessary placeholders yourself of course)

tests/go.mod
```diff
@@ -2,14 +2,18 @@ module github.com/icinga/icingadb/tests
...
+replace github.com/icinga/icingadb => ../
+
 require (
 	github.com/containerd/containerd v1.5.6 // indirect
-	github.com/go-redis/redis/v8 v8.11.3
+	github.com/go-redis/redis/v8 v8.11.4
 	github.com/google/uuid v1.3.0
-	github.com/icinga/icinga-testing v0.0.0-20211203084126-748428acf86d
+	github.com/icinga/icinga-testing v0.0.0-20220204140727-68d4825f41cd
+	github.com/icinga/icingadb v1.0.0-rc2
```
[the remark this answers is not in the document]
lippserd: `latest` does not work. go complains that go mod requires an update.

pkg/icingadb/cleanup.go
```diff
@@ -0,0 +1,78 @@
...
+		return fmt.Sprintf(`WITH rows AS (
+SELECT %[1]s FROM %[2]s WHERE %[3]s < :time ORDER BY %[3]s LIMIT %[4]d
+)
+DELETE FROM %[2]s WHERE %[1]s IN (SELECT %[1]s FROM rows)`, stmt.PK, stmt.Table, stmt.Column, limit)
```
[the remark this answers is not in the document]
lippserd: I haven't found a tabbed version that looks better, so I'd just leave it as is.

tests/cleanup_and_retention_test.go
```diff
@@ -0,0 +1,186 @@
...
+			))
+			require.NoError(t, err)
+
+			total := scenario.rowsToDelete + scenario.rowsToSpare
+			values := make([]struct{ Time int64 }, 0, MaxPlaceholders)
+			start := time.Now().AddDate(0, 0, -retentionDays).Add(-1 * time.Millisecond * time.Duration(scenario.rowsToDelete))
+			startMilli := utils.UnixMilli(start)
+			for j := uint64(0); j < total; j++ {
+				if j == scenario.rowsToDelete {
+					startMilli += int64(time.Minute)
```
julianbrost, line 110: The test would fail if inserting all test data would take longer than this, wouldn't it? In my one-off test, inserting took 15s, so there's some safety margin there, but would there be any downside in using a larger value here?

julianbrost (changes requested):

tests/go.mod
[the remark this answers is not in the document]
julianbrost: I think I saw this somewhere before, but couldn't find any reference. Maybe I was hallucinating? :fearful:

tests/cleanup_and_retention_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"time"
+)
+
+const MaxPlaceholders = 1 << 13
+
+func TestCleanupAndRetention(t *testing.T) {
+	database := func(t *testing.T) *icingadb.DB {
+		rdb := getEmptyDatabase(t)
+		db, err := sqlx.Connect(rdb.Driver(), rdb.DSN())
+		require.NoError(t, err, "connecting to MySQL")
```
julianbrost, line 25: This message should now also be database-agnostic.

julianbrost (changes requested):

julianbrost (approved):

julianbrost:
@cla-bot check

julianbrost:
Looks like the bot doesn't want to report a status for "verification/cla-signed"...

PR #436, opened 2022-02-03
Fix currently not able to enable TLS

lippserd:
When embedding `TLS`, the field name is implicitly specified as `TLS`, which conflicts with the YAML key `tls`, which currently does not allow TLS to be enabled. I'm not entirely sure if this is a bug in goccy/go-yaml, but if a struct field name other than TLS is specified, the configuration works again.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -83,7 +83,7 @@ func ParseFlags() (*Flags, error) {
 
 // TLS provides TLS configuration options for Redis and Database.
 type TLS struct {
-	Tls      bool   `yaml:"tls"`
+	Enable   bool   `yaml:"tls"`
 	Cert     string `yaml:"cert"`
 	Key      string `yaml:"key"`
 	Ca       string `yaml:"ca"`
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -83,7 +83,7 @@ func ParseFlags() (*Flags, error) {
 
 // TLS provides TLS configuration options for Redis and Database.
 type TLS struct {
-	Tls      bool   `yaml:"tls"`
+	Enable   bool   `yaml:"tls"`
```
Al2Klimov, line 86: Go field <> YAML field

pkg/config/database.go
```diff
@@ -19,13 +19,13 @@ var registerDriverOnce sync.Once
...
-	User     string `yaml:"user"`
-	Password string `yaml:"password"`
-	TLS      `yaml:",inline"`
-	Options  icingadb.Options `yaml:"options"`
+	Host       string           `yaml:"host"`
+	Port       int              `yaml:"port"`
+	Database   string           `yaml:"database"`
+	User       string           `yaml:"user"`
+	Password   string           `yaml:"password"`
+	TlsOptions TLS              `yaml:",inline"`
```
Al2Klimov, line 27: Inlined, but not embedded

Al2Klimov (approved):
Not exactly as I'd do it, but it is fine.

PR #439, opened 2022-02-14
Add previous_soft_state to host_state and service_state

### LINKED ISSUE #437: Add `previous_soft_state` to `host_state` and `service_state`
Web would like to show the user the previous soft state in an object's detail header. `previous_hard_state` is already available, though `previous_soft_state` is not.

Al2Klimov:
fixes #437


julianbrost (changes requested):
1.0.0rc2 is already released, so don't change that schema upgrade file.

julianbrost (approved):

PR #447, opened 2022-03-08
Bump icinga-testing version to fix intermittent test failures

### LINKED ISSUE #433: Tests may fail due to out of sync conf.d
The [scheduled integration test run from tonight](https://github.com/Icinga/icingadb/runs/4932033202?check_suite_focus=true) failed because the state history streams were inconsistent between the two masters. The inconsistencies were in checkables that come from the default contents of `conf.d/` which is out of sync between the two masters in the tests and should be removed there: https://github.com/Icinga/icingadb/blob/68a9f7faba496184203e612931a19e2bcb765ea6/tests/history_test.go#L87-L113

This might need additional functionality in the icinga-testing repo, for example to remove files matching a glob (could also be implemented in other ways though).

julianbrost:
New icinga-testing will include a commit that fixes #433

### Blocked by
* https://github.com/Icinga/icinga-testing/pull/15

### Todo
* [ ] After merging above PR, update this PR to use the merge commit

fixes #433

### DIFF
--- a/tests/go.mod
+++ b/tests/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/containerd/containerd v1.5.6 // indirect
 	github.com/go-redis/redis/v8 v8.11.3
 	github.com/google/uuid v1.3.0
-	github.com/icinga/icinga-testing v0.0.0-20211203084126-748428acf86d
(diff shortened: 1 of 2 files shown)


julianbrost:
Superseded by #449.

PR #448, opened 2022-03-09
Remove parellelism from history test

julianbrost:
This is done to prevent intermittent test failures caused by another sub-test still running that submits check results while the state history stream is checked for consistency.


julianbrost:
Superseded by #449.

PR #449, opened 2022-03-09
Fix intermittent test failures

### LINKED ISSUE #433: Tests may fail due to out of sync conf.d

julianbrost:
This PR fixes two causes of intermittent test failures:

1. In the TestHistory/HA test case, both nodes were started with the default `/etc/icinga2/conf.d/*.conf` still in place. This leads to an inconsistent configuration on both nodes which is doomed to write inconsistent history at some point. This will be fixed by https://github.com/Icinga/icinga-testing/pull/15 and this PR bumps the version of icinga-testing.
2. Running the sub-tests of TestHistory/HA in parallel can result in a situation where some history stream is checked for consistency while another sub-test is still running and generating events that also affect that stream (for example, the downtime test may generate notification, the notification test may generate state changes, ...) that are only processed by one of the nodes so far. Therefore remove the parallelism altogether from that test case. The runtime is dominated by the downtime test anyways.

### Blocked by
* https://github.com/Icinga/icinga-testing/pull/15

### Todo
* [x] After merging above PR, update this PR to use the merge commit

fixes #433

### DIFF
--- a/tests/go.mod
+++ b/tests/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/containerd/containerd v1.5.6 // indirect
 	github.com/go-redis/redis/v8 v8.11.3
 	github.com/google/uuid v1.3.0
-	github.com/icinga/icinga-testing v0.0.0-20211203084126-748428acf86d
+	github.com/icinga/icinga-testing v0.0.0-20220309154907-12cf833da4f3
 	github.com/jmoiron/sqlx v1.3.4
 	github.com/stretchr/testify v1.7.0
 	go.uber.org/zap v1.19.1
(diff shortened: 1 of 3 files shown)


julianbrost:
Individual PRs weren't really feasible as the fix for (2) makes the issue from (1) more likely to appear as the state consistency check is performed later than before so there was more time for the inconsistent configuration to actually cause inconsistent state history.

julianbrost:
I've clicked retry two times now and things look good for me. Feel free to retry more often.

If you're happy, please do me a favor: merge this PR and rebase #221 (should make thinks easier for me in #418).

Al2Klimov (approved):

PR #456, opened 2022-03-24
Schema: Make custom variable names case insensitive

### LINKED ISSUE #443: Make custom variable searches case-insensitive by default
The columns `flatname` and `flatvalue` in the table `customvar_flat` should use `COLLATE utf8mb4_unicode_ci`.

lippserd:
To make custom variable searches case-insensitive by default.

fixes #443

### DIFF
--- a/schema/mysql/schema.sql
+++ b/schema/mysql/schema.sql
@@ -809,7 +809,7 @@ CREATE TABLE customvar (
   environment_id binary(20) NOT NULL COMMENT 'environment.id',
   name_checksum binary(20) NOT NULL COMMENT 'sha1(name)',
 
-  name varchar(255) NOT NULL,
+  name varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL,
   value text NOT NULL,
 
   PRIMARY KEY (id)
@@ -821,7 +821,7 @@ CREATE TABLE customvar_flat (
   customvar_id binary(20) NOT NULL COMMENT 'sha1(customvar.id)',
   flatname_checksum binary(20) NOT NULL COMMENT 'sha1(flatname after conversion)',
 
-  flatname varchar(512) NOT NULL COMMENT 'Path converted with `.` and `[ ]`',
+  flatname varchar(512) COLLATE utf8mb4_unicode_ci NOT NULL COMMENT 'Path converted with `.` and `[ ]`',
   flatvalue text NOT NULL,
 
   PRIMARY KEY (id),
--- a/schema/mysql/upgrades/1.0.0.sql
+++ b/schema/mysql/upgrades/1.0.0.sql
@@ -41,5 +41,11 @@ ALTER TABLE notes_url
   MODIFY id binary(20) NOT NULL COMMENT 'sha1(environment.id + notes_url)',
   ADD PRIMARY KEY (id);
 
+ALTER TABLE customvar
+  MODIFY name varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL;
+
+ALTER TABLE customvar_flat
+  MODIFY flatname varchar(512) COLLATE utf8mb4_unicode_ci NOT NULL COMMENT 'Path converted with `.` and `[ ]`';
+
 INSERT INTO icingadb_schema (version, TIMESTAMP)
   VALUES (3, CURRENT_TIMESTAMP() * 1000);
--- a/schema/pgsql/schema.sql
+++ b/schema/pgsql/schema.sql
@@ -1370,7 +1370,7 @@ CREATE TABLE customvar (
   environment_id bytea20 NOT NULL,
   name_checksum bytea20 NOT NULL,
 
-  name varchar(255) NOT NULL,
+  name citext NOT NULL,
   value text NOT NULL,
 
   CONSTRAINT pk_customvar PRIMARY KEY (id)
@@ -1390,7 +1390,7 @@ CREATE TABLE customvar_flat (
   customvar_id bytea20 NOT NULL,
   flatname_checksum bytea20 NOT NULL,
 
-  flatname varchar(512) NOT NULL,
+  flatname citext NOT NULL,
   flatvalue text NOT NULL,
 
   CONSTRAINT pk_customvar_flat PRIMARY KEY (id)


julianbrost (changes requested):
Merge conflict has to be resolved.

julianbrost:
By the way, what will happen in icingadb-web if I set
```
vars.foo = "bar"
vars.FOO = "bar"
```
and then search for that?

Will there be additional changes in other repos? (I'm going to test this! :D)

julianbrost:
Oh and yet another thought: If I set
```
vars.foo = "x"
vars.FOO = "y"
```
what will a search for `foo == "x" && foo != "x"` return?

nilmerg:
Try it, I'd say. There will be no changes regarding this in any web repository. Icinga Cube however is the best example to look at (with Icinga DB support enabled) as without this the handling of these variables will differ from how they're handled with the IDO backend.

julianbrost:
General questions:
1. Why is only the variable name supposed to be case-insensitive but not the value?
2. What's the intention for having the variable name case-insensitive? Just allowing people to be lazy and not bother pressing the shift key when typing their search queries? Or do you expect the same variable to be present in different objects with different capitalization?

Also, results of my tests with the following object:

```
object Host "icingadb-456" {
	check_command = "dummy"
	vars.foo = "foo"
	vars.FOO = "foo"
	vars.bar = "x"
	vars.BAR = "y"
}
```

* `host.vars.foo=foo` returns just one object (I suspected it could return the same object twice because the same variable matches twice, but that doesn't happen)
* `host.vars.bar=x&host.vars.bar!=x` returns no objects (From looking at the SQL query, the semantics of `!=` seems to be to exclude objects that have a matching value, rather than searching for variables with different values)
* `host.vars.bar=!y` also returns no values, maybe a bit strange since that variable is `x` (but also `y` at the same time)
* `host.vars.bar=x&host.vars.bar=y` returns one object, also a bit strange as this would imply `x=y`

nilmerg:
General answers:

1. It has been introduced due to compatibility with Icinga 1.x. (https://github.com/Icinga/icingaweb2/issues/1461) Now? Because users are used to it.
2. In some environments it's not just one user who configures Icinga. Consider one user to write `vars.community`, another one writing `vars.Community` and then one who defines a restriction so that Web hides vars which contain a SNMP community. (That's the default by the way and we don't have both casings in the default.)

> `host.vars.foo=foo` returns just one object (I suspected it could return the same object twice because the same variable matches twice, but that doesn't happen)

Luckily, that'd be bad.

> `host.vars.bar=x&host.vars.bar!=x` returns no objects (From looking at the SQL query, the semantics of != seems to be to exclude objects that have a matching value, rather than searching for variables with different values)

Indeed. That's also the expectation of users who want to exclude hosts or services inside specific groups, no matter which other groups they're part of.

> `host.vars.bar=!y` also returns no values, maybe a bit strange since that variable is x (but also y at the same time)

Is that really a `=!`? Then the exclamation mark is part of the value. With a `!=` it should show the object, since the other variable is indeed not `y`.

> `host.vars.bar=x&host.vars.bar=y` returns one object, also a bit strange as this would imply x=y

No, not strange. If columns are case-insensitive this is to be expected.

julianbrost:
> > `host.vars.bar=!y` also returns no values, maybe a bit strange since that variable is x (but also y at the same time)
> 
> Is that really a `=!`? Then the exclamation mark is part of the value. With a `!=` it should show the object, since the other variable is indeed not `y`.

Nope, typo. Not sure exactly which one I tested yesterday, but I just tested again and `host.vars.bar!=y` returns no results (while `host.vars.bar=x` does return something).

> > `host.vars.bar=x&host.vars.bar=y` returns one object, also a bit strange as this would imply x=y
> 
> No, not strange. If columns are case-insensitive this is to be expected.

When saying strange, I mean from a logical point of view, not that the behavior can't be explained knowing the implementation.

nilmerg:
> With a != it should show the object, since the other variable is indeed not y.

That's bullshit. I've already said why `host.vars.bar=x&host.vars.bar!=x` leads to the result it shows. Why should `host.vars.bar!=y` be different..

julianbrost:
But in summary: To get strange results, you have to put yourself in the strange situation of defining multiple variable differing only in case *on the same* object. And even if you do, results are maybe strange, but not terribly broken.

In general, it would be nice if this was consistent throughout the whole stack. Like with this change, imagine some host template that sets `vars.something` and then someone sets `vars.Something = "else"`, then it will look like they overwrote the value when they search for `host.vars.something`, but then also look like they didn't overwrite it when they open the host and look at the variables because it shows up there twice. But retroactively making variables case-insensitive in icinga2 doesn't sound feasible either.

So what do we do with this? One can probably argue for and against any combination of making name and value case-(in)sensitive. So if you deem this necessary to keep users happy, go ahead. On the other hand I wouldn't mind not doing things because the IDO/monitoring module did something as a workaround because icinga1 did something.

julianbrost:
> it's nice that I don't have to worry about capitalization. Especially if I don't know the complete configuration and have to look up how the variable is written exactly.

Doesn't icingadb-web suggest variable names that would hint the correct capitalization? Okay, turns out, to suggest capitalization, you have to search case-insensitive in the first place (oh no). But as these suggestions use `LIKE '%...%' anyways, I don't think an index would be used anyways, so throwing in another `LOWER(...)` would be possible. I don't know how feasible implementing this in icingadb-web would be though.

> Also, existing filters and restrictions moved to Icinga DB Web may no longer work without these changes.

But of course that suggestion wouldn't address this. I don't know how important this is. Like are restrictions automatically migrated to from monitoring to icingadb-web? Or if both are enabled, do they maybe even use the same restrictions?

> I also think it would be great if Icinga enforced something like `unique lower(trim($key))` but unfortunately that's asking too much.

I mean we could try to start with a warning and maybe consider deprecating this in the sense of making the warning an error in a later version. Really strange things only happen if you use different capitalization for a variable in the same object and I doubt there are many cases where you would want to do this intentionally.

julianbrost (approved):
Change does what it advertises, I have no clear answer to what's the best way moving forward, so merge at your own discretion.

julianbrost:
@cla-bot check

PR #452, opened 2022-03-16
Enable generics

Al2Klimov:
Because we can.

closes #457

PR #457, opened 2022-03-30
Remove obsolete comment

Al2Klimov:
That Go version doesn’t differ from the mainly used one.


julianbrost (approved):
Don't you think a separate PR is overkill when I mention this in [another PR](https://github.com/Icinga/icingadb/pull/452#discussion_r838373478)?

Al2Klimov:
GHA says: no. :)

PR #458, opened 2022-04-07
Support *nix sockets as documented

Al2Klimov:
and document as supported.

### DIFF
--- a/pkg/config/database.go
+++ b/pkg/config/database.go
@@ -13,6 +13,7 @@ import (
 	"net"
 	"net/url"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


pkg/config/database.go
```diff
@@ -68,7 +76,6 @@ func (d *Database) Open(logger *logging.Logger) (*icingadb.DB, error) {
 		uri := &url.URL{
 			Scheme: "postgres",
 			User:   url.UserPassword(d.User, d.Password),
-			Host:   net.JoinHostPort(d.Host, strconv.FormatInt(int64(d.Port), 10)),
 			Path:   "/" + url.PathEscape(d.Database),
 		}
```
julianbrost, line 80: Just after this block, `uri.Host` is used. (Also, what happens if you enable TLS and use UNIX sockets?)

julianbrost (changes requested):
Additionally, please rebase so that GH Actions can pass again.

pkg/config/database.go
Al2Klimov: (Nobody does this.)

julianbrost (approved):

PR #459, opened 2022-04-07
Ensure icingadb_instance#responsible=n

Al2Klimov:
for non-responsible instances.

### DIFF
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -14,6 +14,7 @@ import (
 	"github.com/icinga/icingadb/pkg/logging"
 	"github.com/icinga/icingadb/pkg/retry"
 	"github.com/icinga/icingadb/pkg/types"
+	"github.com/icinga/icingadb/pkg/utils"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"sync"
@@ -145,6 +146,7 @@ func (h *HA) controller() {
 				if tt.Before(now.Add(-1 * timeout)) {
 					h.logger.Errorw("Received heartbeat from the past", zap.Time("time", tt))
 					h.signalHandover()
+					h.realizeLostHeartbeat()
 					continue
 				}
 				s, err := m.Stats().IcingaStatus()
@@ -211,6 +213,7 @@ func (h *HA) controller() {
 			} else {
 				h.logger.Error("Lost heartbeat")
 				h.signalHandover()
+				h.realizeLostHeartbeat()
 			}
 		case <-h.heartbeat.Done():
 			if err := h.heartbeat.Err(); err != nil {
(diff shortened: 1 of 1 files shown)


Al2Klimov:
@julianbrost My icingadb_instance stays empty, any ideas?

pkg/icingadb/ha.go
```diff
@@ -286,6 +288,13 @@ func (h *HA) realize(ctx context.Context, s *icingaredisv1.IcingaStatus, t *type
 				return internal.CantPerformQuery(err, stmt)
 			}
 
+			if takeover {
+				stmt := h.db.Rebind("UPDATE icingadb_instance SET responsible = ? WHERE environment_id = ? AND id <> ?")
+				_, err := tx.ExecContext(ctx, stmt, "n", envId, h.instanceId)
+
+				return internal.CantPerformQuery(err, stmt)
```
julianbrost, line 295: Too eager error handling ...

pkg/icingadb/ha.go
```diff
@@ -286,6 +288,13 @@ func (h *HA) realize(ctx context.Context, s *icingaredisv1.IcingaStatus, t *type
...
+			}
+
 			if err := tx.Commit(); err != nil {
```
julianbrost, line 298: ... so it won't even attempt to commit this transaction (if `takeover == true`).

pkg/icingadb/ha.go
Al2Klimov: Why... exactly?

julianbrost: You're always returning here, even if there was no error?

pkg/icingadb/ha.go
```diff
@@ -324,6 +335,32 @@ func (h *HA) realize(ctx context.Context, s *icingaredisv1.IcingaStatus, t *type
...
+					if attempt > 2 {
+						log = h.logger.Infow
+					}
+
+					log("Can't update instance. Retrying", zap.Error(err), zap.Uint64("retry count", attempt))
+				}
+			},
+		},
+	)
+}
```
julianbrost, line 362: `realize()` can also write `responsible = 'n'`:
https://github.com/Icinga/icingadb/blob/af41705f115bda2cd4870b16af6a56c0545a379c/pkg/icingadb/ha.go#L271

So I'm not sure if this new function is needed at all. Why is it?

Al2Klimov: I am sure that in this case a transaction is not needed at all.

julianbrost (changes requested):

pkg/icingadb/ha.go
Al2Klimov: `realize()` has already enough purpose(s), complexity and LOC.

pkg/icingadb/ha.go
```diff
@@ -323,6 +334,32 @@ func (h *HA) realize(ctx context.Context, s *icingaredisv1.IcingaStatus, t *type
 	return nil
 }
 
+func (h *HA) realizeLostHeartbeat() {
+	_ = retry.WithBackoff(
```
julianbrost, line 338: Okay, so if you insist on keeping this function, let's consider it in more detail: what's the reason for using `retry.WithBackoff()` here? There are multiple other functions in this file doing SQL queries and the only other one using it is `realize()`, but as far as I can tell, it's used there because it's using a transaction with isolation level serializable that may fail to commit and this isn't the case here.

Also, this is some kind of "let's try to be nice" feature. The worst that would happen if that query fails would be that the next instance to take over due to an expired heartbeat would reset the `responsible` column for the old instance (similar to when it just crashed instead).

Changing this even removes most of the duplication with `realize()` so I wouldn't even be that unhappy about keeping it then I guess.

Al2Klimov: The reason is... a copy&🐘 error.

(Btw. I don’t _insist_. I was just reasoning.)

julianbrost (approved):

PR #460, opened 2022-04-08
Schema: Adjust primary key irregularities of icon_image, notes_url & …

### LINKED ISSUE #450: Primary key of icon_image, action_url, notes_url should not explicitly include the environment_id column
These three tables have `PRIMARY KEY (environment_id, id)` and the comment for `id` says that the environment is not included in the hash:
https://github.com/Icinga/icingadb/blob/dece98e3e9e12bd592e2ea5d59ac6cb6f6a3ed23/schema/mysql/schema.sql#L722-L747

However, the code that generates the IDs actually includes the environment ID:
https://github.com/Icinga/icinga2/blob/fb2a58f177eddde909261e3da06894e28c889d3f/lib/icingadb/icingadb-objects.cpp#L682-L729

So we can just remove this irregularity from the schema without other code changes.

yhabteab:
…action_url tables

fixes #450

### DIFF
--- a/schema/mysql/upgrades/1.0.0.sql
+++ b/schema/mysql/upgrades/1.0.0.sql
@@ -26,6 +26,20 @@ ALTER TABLE flapping_history
 ALTER TABLE state_history
   ADD index idx_state_history_event_time (event_time) COMMENT 'Filter for history retention';
 
+ALTER TABLE icon_image
+  DROP PRIMARY KEY,
+  MODIFY id binary(20) NOT NULL COMMENT 'sha1(environment.id + icon_image)',
+  ADD PRIMARY KEY (id);
+
+ALTER TABLE action_url
+  DROP PRIMARY KEY,
+  MODIFY id binary(20) NOT NULL COMMENT 'sha1(environment.id + action_url)',
+  ADD PRIMARY KEY (id);
+
+ALTER TABLE notes_url
+  DROP PRIMARY KEY,
+  MODIFY id binary(20) NOT NULL COMMENT 'sha1(environment.id + notes_url)',
+  ADD PRIMARY KEY (id);
 
 INSERT INTO icingadb_schema (version, TIMESTAMP)
   VALUES (3, CURRENT_TIMESTAMP() * 1000);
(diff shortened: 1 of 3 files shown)


julianbrost (changes requested):
All the other comments use `environment.id` instead of `environment_id`, so this should stay consistent.

Also, for MySQL, this has to be added to the upgrade file (but not for PostgreSQL as there's no released version of the schema yet).

yhabteab:
Should the pk always have to be created with a constraint? Like `add constraint pk_xy pk ..`

schema/mysql/upgrades/1.0.0.sql
```diff
@@ -26,6 +26,23 @@ ALTER TABLE flapping_history
...
+  DROP COLUMN id,
+  DROP PRIMARY KEY,
+  ADD COLUMN id binary(20) NOT NULL COMMENT 'sha1(environment.id + action_url)' FIRST,
+  ADD PRIMARY KEY (id);
+
+ALTER TABLE notes_url
+  DROP COLUMN id,
+  DROP PRIMARY KEY,
+  ADD COLUMN id binary(20) NOT NULL COMMENT 'sha1(environment.id + notes_url)' FIRST,
+  ADD PRIMARY KEY (id);
```
julianbrost, line 45: Will this work if there is data in the tables?

julianbrost:
> Should the pk always have to be created with a constraint? Like `add constraint pk_xy pk ..`

I'm not sure what this question is about. Given that you mention explicitly named constraints, I think you're referring to PostgreSQL but that change seems to be fine. In case your question is if we always do these named explicitly for PostgreSQL, then yes, we need to refer to these constraints in https://github.com/Icinga/icingadb/blob/89bcc99071fa23667efc45aaf2b981fa3c3d5cac/pkg/icingadb/db.go#L198-L200

schema/mysql/upgrades/1.0.0.sql
yhabteab: Nope, they will lose their hash IDs. Fixed!

yhabteab:
> I'm not sure what this question is about.

My question is referring to whether I should explicitly set the pk name in the upgrade script or not!

julianbrost:
For MySQL? We don't name them in the full schema, so no need to do so in the upgrade. (Is that even something MySQL would support?)

The current version of the PR seems fine (i.e. I don't have any more complaints right now, I just have to test it).

julianbrost (approved):

julianbrost:
@yhabteab Can this be merged immediately or should we rather wait for https://github.com/Icinga/icingadb-web/pull/525 to also be ready?

yhabteab:
It's already approved, so if you want to merge this, feel free.

PR #461, opened 2022-04-11
PostgreSQL: Support LIKE for integer/enum columns

lippserd:
At the moment Icinga DB Web doesn't know the column types, so it sends
SQL queries with LIKE operators for all suggestions in the search bar,
which fails for numeric and enum types on PostgreSQL. To support this,
the LIKE operator (internally translated to ~~) is overloaded. Note
that this is only a temporary solution until Icinga DB Web provides
column type support.

### DIFF
--- a/schema/pgsql/schema.sql
+++ b/schema/pgsql/schema.sql
@@ -161,6 +161,24 @@ BEGIN
 END;
 $$;
 
+/* At the moment Icinga DB Web doesn't know the column types,
+   so it sends SQL queries with LIKE operators for all suggestions in the search bar,
+   which fails for numeric and enum types on PostgreSQL.
+   To support this, the LIKE operator (internally translated to ~~) is overloaded.
+   Note that this is only a temporary solution until Icinga DB Web provides column type support.
+ */
+CREATE OR REPLACE FUNCTION anynonarrayliketext(anynonarray, text)
+  RETURNS bool
+  LANGUAGE plpgsql
+  IMMUTABLE
+  PARALLEL SAFE
+  AS $$
+    BEGIN
+      RETURN $1::TEXT LIKE $2;
+    END;
+  $$;
+CREATE OPERATOR ~~ (LEFTARG=anynonarray, RIGHTARG=text, PROCEDURE=anynonarrayliketext);
+
 CREATE TABLE host (
   id bytea20 NOT NULL,
   environment_id bytea20 NOT NULL,


yhabteab (commented):
As it is, pgsql throws an error that the uint operator isn't unique, as the two domains (`uint`, `biguint`) have the same underlying data type. Hence just using three operators (`smallint`, `int`, `bigint`) are sufficient. With this I tested it and it works great.

yhabteab (commented):
Requesting for a `Host/Service Check Interval=5.5` produces the following result:

Bildschirmfoto 2022-04-12 um 16 06 44

julianbrost:
Can't web just always use `x::text LIKE y` instead of `x LIKE y` in the queries?

julianbrost:
For completeness as we discussed this somewhere else: `x::text` wouldn't work properly for columns of type `citext`.

schema/pgsql/schema.sql
```diff
@@ -21,6 +21,55 @@ CREATE TYPE comment_type AS ENUM ( 'comment', 'ack' );
 CREATE TYPE notification_type AS ENUM ( 'downtime_start', 'downtime_end', 'downtime_removed', 'custom', 'acknowledgement', 'problem', 'recovery', 'flapping_start', 'flapping_end' );
 CREATE TYPE history_type AS ENUM ( 'notification', 'state_change', 'downtime_start', 'downtime_end', 'comment_add', 'comment_remove', 'flapping_start', 'flapping_end', 'ack_set', 'ack_clear' );
 
+/* For the search bar suggestions, Icinga DB Web always issues SQL queries with text values,
+   even if the value should be a number, such as for check_interval.
+   This fails for PostgreSQL as it cannot match numbers to text. To support this,
```
nilmerg, line 26: This is misleading. DB Web doesn't just pass numbers as text, it's rather passing numbers encapsulated by wildcards (e.g. `%4%`) which is only possible by passing text. The solution in DB Web also is not by offering different operators but by disallowing wildcards for integer columns.

lippserd:
> Requesting for a `Host/Service Check Interval=5.5` produces the following result:
> 
> Bildschirmfoto 2022-04-12 um 16 06 44

This cannot be fixed by operator overloading, but requires column type support in Icinga DB Web.

julianbrost:
Is PostgreSQL support in icingadb-web in a state where it's meaningful for me to use that for testing this?

Btw (before I tested anything): what happens if you filter for `flapping_threshold_low` (on of the few `float` columns we have)?

lippserd:
> Is PostgreSQL support in icingadb-web in a state where it's meaningful for me to use that for testing this?

Yes. You can also test this by running `SELECT 25::biguint LIKE '%5%';` or `SELECT 25::float LIKE '%5%';` and so on.

> Btw (before I tested anything): what happens if you filter for `flapping_threshold_low` (on of the few `float` columns we have)?

Good point. Fixed.

julianbrost:
Trying LIKE for the different types we use in the schema not yet mentioned in this PR:
* `text`, `citext`, `varchar`: works out of the box as you'd expect.
* `bytea` (and `byteaN`): works out of the box, probably of limited usefulness, but not an error (users shouldn't typically filter on these raw values anyways but rather on the referenced objects).
* `enum` (including the custom types like `boolenum`): haven't tested with icingadb-web + PostgreSQL, but with MySQL it offers me to filter on these columns and doing this manually in PostgreSQL with LIKE yields an error, so same problem there as well?
    ```
    icingadb=# SELECT * FROM host WHERE event_handler_enabled LIKE '%42%';
    ERROR:  operator does not exist: boolenum ~~ unknown
    LINE 1: SELECT * FROM host WHERE event_handler_enabled LIKE '%42%';
    ```

julianbrost:
I can't follow. Doing the very same operator dance seems to work for enum types as well:
```
icingadb=# CREATE FUNCTION boolenumliketext(boolenum, text) RETURNS boolean
  LANGUAGE SQL
  IMMUTABLE
  PARALLEL SAFE
  RETURN $1::TEXT LIKE $2;
CREATE OPERATOR ~~ (LEFTARG=boolenum, RIGHTARG=text, PROCEDURE=boolenumliketext);
CREATE FUNCTION
CREATE OPERATOR
icingadb=# SELECT * FROM host WHERE event_handler_enabled LIKE '%42%';
(0 rows)
```

lippserd:
> I can't follow. Doing the very same operator dance seems to work for enum types as well:

Strange. I will do the tests again.

lippserd:
> > I can't follow. Doing the very same operator dance seems to work for enum types as well:
> 
> Strange. I will do the tests again.

Found a catch-all solution.

julianbrost (commented):
Nice, so now this depends on https://github.com/Icinga/icinga-testing/pull/21 (in particular https://github.com/Icinga/icinga-testing/pull/21/commits/01aa6d63dfa0c863b21fb3d0618692e861caa05f) for the tests.

Side note: my personal preference would be putting `AS $$...$$` last (but you don't have to change that).

julianbrost (changes requested):
Version bump for icinga-testing was merged now, so a rebase should make the tests work. Also needed to resolve merge conflicts now.

lippserd:
> Side note: my personal preference would be putting `AS $$...$$` last (but you don't have to change that).

Mine too, but that leads to syntax errors.

julianbrost:
Strange, it shouldn't. I mean the SLA function does just that (you can even see the `$$;` in the diff context).

lippserd:
> Strange, it shouldn't. I mean the SLA function does just that (you can even see the `$$;` in the diff context).

Maybe I was just missing a semicolon then. I'll give it another try!

lippserd:
> > Strange, it shouldn't. I mean the SLA function does just that (you can even see the `$$;` in the diff context).
> 
> Maybe I was just missing a semicolon then. I'll give it another try!

Works with your layout!

julianbrost (approved):
Let's see for how long this workaround will live on. (I'd say pretty long, but please do prove me wrong!)

PR #465, opened 2022-04-13
Do x.UnixMilli(), not utils.UnixMilli(x)

Al2Klimov:
and remove redundant utils.UnixMilli().

## TODO

* #452
* [x] re-run checks


julianbrost (approved):
Looks fine, but is the first commit still supposed to be in the PR? Changing a function just to remove it in the next commit seems a bit pointless to me. (Feel free the either merge or remove that commit.)

PR #467, opened 2022-04-27
Use Icinga DB docker instance for retention tests

lippserd:
This change eliminates the dependency on Icinga DB sources for the
retention tests by launching an Icinga DB docker instance. Furthermore,
different retention configurations are now being tested.


tests/cleanup_and_retention_test.go
```diff
@@ -1,191 +1,154 @@
...
-			cleanup icingadb.CleanupResult
+		r := it.RedisServerT(t)
+		i := it.Icinga2NodeT(t, "master")
+		i.EnableIcingaDb(r)
+		i.Reload()
+		waitForDumpDoneSignal(t, r, 20*time.Second, 100*time.Millisecond)
+		it.IcingaDbInstanceT(t, r, rdb, services.WithIcingaDbConfig(fmt.Sprintf(`
+history-retention:
+  days: %d
+`, retentionDays)))
```
julianbrost, line 81: Would be nice to specify explicit retention policies for two types, one just using a different value, and disabling cleanup for another one.

julianbrost (changes requested):
The more I look at this dropping and recreating the table, the less I like it.

Alternative suggestion: Add some `IgnoreNotNullCols []string` to `CleanupStmt` and then do `ALTER TABLE t DROP COLUMN c` for each.

PostgreSQL would also support `ALTER COLUMN c DROP NOT NULL` without having to specify the full column type again, MySQL can only do `ALTER COLUMN c SET DEFAULT d` (but maybe there's a default good enough for all column types, like `0` or `''`)?

tests/cleanup_and_retention_test.go
```diff
@@ -1,191 +1,154 @@
...
+
+	var binaryColumn string
+	switch db.DriverName() {
+	case "mysql":
+		binaryColumn = `binary(20)`
+	case "postgres":
+		binaryColumn = `bytea20`
 	}
 
 	t.Run("History", func(t *testing.T) {
```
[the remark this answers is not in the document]
lippserd: Nope, fixed.

tests/cleanup_and_retention_test.go
```diff
@@ -1,191 +1,154 @@
...
+`, retentionDays)))
+
+		time.Sleep(time.Minute)
+
+		for _, stmt := range RetentionStatements {
+			var count int
+			err := db.QueryRow(fmt.Sprintf(`SELECT COUNT(*) FROM %s`, stmt.Table)).Scan(&count)
+			assert.NoError(t, err)
+			assert.Equal(t, rowsToSpare, count, "rows left in table")
 		}
```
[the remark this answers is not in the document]
lippserd: Nice, fixed.

tests/cleanup_and_retention_test.go
lippserd: Done.

lippserd:
> The more I look at this dropping and recreating the table, the less I like it.
> 
> Alternative suggestion: Add some `IgnoreNotNullCols []string` to `CleanupStmt` and then do `ALTER TABLE t DROP COLUMN c` for each.
> 
> PostgreSQL would also support `ALTER COLUMN c DROP NOT NULL` without having to specify the full column type again, MySQL can only do `ALTER COLUMN c SET DEFAULT d` (but maybe there's a default good enough for all column types, like `0` or `''`)?

Fixed by selecting the NOT NULL columns from the information schema.

tests/cleanup_and_retention_test.go
```diff
@@ -1,191 +1,166 @@
...
+	rowsToSpare := 1000
+
+	for category, stmt := range retentionStatements {
+		rows, err := db.Query(db.Rebind(fmt.Sprintf(`
+SELECT column_name
+FROM information_schema.columns
+WHERE table_schema = %s AND table_name = ? AND column_name NOT IN (?, ?) AND is_nullable = ?`,
+			schema)),
+			stmt.Table, stmt.PK, stmt.Column, "NO")
+		if err != nil {
```
julianbrost, line 52: I'd move everything that's related to fixing up the schema for the test into a helper function. Either it gets a pretty descriptive name or a comment explaining why that's done would also be nice.

(Comment affect more code than selected but I didn't want to select half the diff because that would also be pointless.)

tests/cleanup_and_retention_test.go
```diff
@@ -1,191 +1,166 @@
...
-					require.NoError(t, err)
-					values = values[:0]
-				}
+			if _, err := db.Exec(fmt.Sprintf(`ALTER TABLE %s DROP COLUMN %s`, stmt.Table, col)); err != nil {
+				require.NoError(t, err)
 			}
+		}
+		if err := rows.Err(); err != nil {
+			require.NoError(t, err)
+		}
```
julianbrost, line 66: I'd go for `Select()` from sqlx here which can directly put the result in a `[]string` and only needs error handling once.

tests/cleanup_and_retention_test.go
```diff
@@ -1,191 +1,166 @@
...
+	config, err := yaml.Marshal(struct {
+		Retention retention `yaml:"history-retention"`
+	}{reten})
+	assert.NoError(t, err)
+	it.IcingaDbInstanceT(t, r, rdb, services.WithIcingaDbConfig(string(config)))
+
+	eventually.Assert(t, func(t require.TestingT) {
+		for category, stmt := range retentionStatements {
+			var count int
+			err := db.QueryRow(fmt.Sprintf(`SELECT COUNT(*) FROM %s`, stmt.Table)).Scan(&count)
```
julianbrost, line 113: Some check that the oldest rows were indeed the ones deleted would be nice. So basically doing that query twice, once with `WHERE time_column < threshold` and once with the inverse.

julianbrost (changes requested):

tests/cleanup_and_retention_test.go
lippserd: Done.

tests/cleanup_and_retention_test.go
lippserd: Done.

tests/cleanup_and_retention_test.go
lippserd: Done.

tests/cleanup_and_retention_test.go
```diff
@@ -1,191 +1,191 @@
...
-
-			_, err = db.Exec(fmt.Sprintf(`CREATE INDEX idx_%[1]s ON %[1]s (%[2]s)`, stmt.Table, stmt.Column))
-			require.NoError(t, err)
-
-			total := scenario.rowsToDelete + scenario.rowsToSpare
-			type row struct {
-				Id   int64
-				Time int64
+			if retentionDays == 0 {
+				// No retention.
```
julianbrost, line 108: I'd say cleanup instead of retention here. No retention could be misinterpreted as "no history should be retained for this type".

tests/cleanup_and_retention_test.go
```diff
@@ -1,191 +1,191 @@
...
+				assert.Equal(t, rowsToDelete, rowsLeft, "rows left in retention period")
+				assert.Equal(t, rowsToSpare, rowsSpared, "rows spared")
```
julianbrost, line 110: I'd just compare the sum of both in a single assert. `thresholdMilli` isn't really that much of a useful value to separate rows in this scenario. Also, if one of these asserts fails, I think it would be more confusing than just a single assert with a message like "all rows should still be present".

julianbrost (changes requested):

tests/go.mod
```diff
@@ -2,16 +2,15 @@ module github.com/icinga/icingadb/tests
 
 go 1.16
 
-replace github.com/icinga/icingadb => ../
+replace github.com/icinga/icinga-testing => github.com/icinga/icinga-testing v0.0.0-20220428124602-54e951e968d8
```
julianbrost, line 5: Dang it! Missed to check this again before doing my final ready to merge check now: The PR over at icinga-testing was merged, so this should just update the version to the latest commit on the main branch in the normal require directive below.

julianbrost (changes requested):

tests/go.mod
lippserd: Well, I could have noticed that too 😆

julianbrost (approved):

PR #470, opened 2022-05-04
Update dependabot reviewers

lippserd:


julianbrost (approved):
Let's see how long it'll take until dependabot annoys me!

PR #471, opened 2022-05-04
go:embed LUA script

Al2Klimov:
for the sake of syntax highlighting.

### DIFF
--- a/pkg/icingadb/overdue/sync.go
+++ b/pkg/icingadb/overdue/sync.go
@@ -2,6 +2,7 @@ package overdue
 
 import (
 	"context"
+	_ "embed"
 	"fmt"
 	"github.com/go-redis/redis/v8"
 	"github.com/google/uuid"
@@ -16,7 +17,9 @@ import (
 	"github.com/icinga/icingadb/pkg/periodic"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
+	"regexp"
 	"strconv"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/icingadb/overdue/sync.go
```diff
@@ -129,39 +132,12 @@ func (s Sync) log(ctx context.Context, objectType string, counter *com.Counter)
...
-
-local res = {redis.call('SDIFF', tempOverdue, icingadbOverdue), redis.call('SDIFF', icingadbOverdue, tempOverdue)}
-
-redis.call('DEL', tempOverdue)
-
-return res
-
-`)
+var luaGetOverdues = redis.NewScript(strings.TrimSpace(
+	regexp.MustCompile(`(?m)^--.*?$`).ReplaceAllString(getOverduesLua, ""),
```
julianbrost, line 139: What's the `?` supposed to do in the regex? Also, if `EVALSHA` is used (from a quick look look at the lib I'm not entirely sure if this happens automatically or only if you call `s.Load()` on the script first), it wouldn't really matter if a few extra comments are sent over the wire once.

Al2Klimov: 1. Explicitly includes only up to the end of the current line.
2. No such automation here.
3. Less (data) is more.

julianbrost: 1. TIL: `*` longest possible match, `*?` shortest possible match.
2. (intentionally left blank, just there to make the list numbering work out)
3. Well, then `EVALSHA` might be a good idea, but time for another PR.

Al2Klimov: 1. -
2. -
3. No, I meant no automation like my regex.

julianbrost (approved):

pkg/icingadb/overdue/sync.go
julianbrost: 3. From a quick look at the code, `script.Run()` attempts to do `EVALSHA` but if that fails, it falls back to `EVAL` instead of loading the script, so seems like we're sending the script every time at the moment (and waste one round trip trying `EVALSHA`)

Al2Klimov: > we're sending the script every time at the moment

Huh?

julianbrost: Ah, I wasn't aware that `EVAL` implicitly caches the script, I thought you'd need an explicit `SCRIPT LOAD` for that. So never mind what I've said (would probably be too easy if https://redis.io/commands/eval/ mentioned any of the words "cache", "hash", "sha1").

PR #474, opened 2022-05-10
Retention: take environment_id into account

### LINKED ISSUE #468: History retention/cleanup ignores environment_id
The history retention/cleanup code ignores the `environment_id` column in its delete statements:
https://github.com/Icinga/icingadb/blob/41d5df3f34f0ef19d89d6bdc5a3fecd5f0eb1245/pkg/icingadb/cleanup.go#L27-L40

I.e. one Icinga DB instance will also delete history from other environments.

Hint: DB indices should probably be updated as well.

julianbrost:
This PR ensures that the retention feature only deletes history from the current environment. Previously, the deletion queries where missing an appropriate WHERE clause.

Also adds missing indices to the PostgreSQL schema. With this PR:
```console
$ diff -us <(grep -o 'idx_\w*' schema/mysql/schema.sql | sort -u) <(grep -o 'idx_\w*' schema/pgsql/schema.sql | sort -u)
Files /proc/self/fd/11 and /proc/self/fd/12 are identical
```

Indices are kept without the environment ID for now as icingadb-web currently does not support multiple environments so that feature is of limited use at the moment anyways and there will almost certainly be only one environment in the database. So adding the environment ID would only make the indices bigger until then and in one case would even require duplicating the index as icingadb-web needs it without the environment ID at the moment. It's quite possible that when icingadb-web gets multi-env support, it might need the index with the environment ID then as well. Once it gets there, one has to reevaluate all these indices anyways.

## Tests

See changes to the included tests.

## Blocked by
* #247: That PR also touches retention, so changes here would only introduce merge conflicts anyways. As SLA reporting should be mostly done now, I'd want to avoid having to delay it due to having to resolve merge conflicts.

closes #468

### DIFF
--- a/pkg/icingadb/cleanup.go
+++ b/pkg/icingadb/cleanup.go
@@ -21,10 +21,11 @@ type CleanupStmt struct {
 func (stmt *CleanupStmt) Build(driverName string, limit uint64) string {
 	switch driverName {
 	case driver.MySQL, "mysql":
-		return fmt.Sprintf(`DELETE FROM %[1]s WHERE %[2]s < :time ORDER BY %[2]s LIMIT %[3]d`, stmt.Table, stmt.Column, limit)
+		return fmt.Sprintf(`DELETE FROM %[1]s WHERE environment_id = :environment_id AND %[2]s < :time
+ORDER BY %[2]s LIMIT %[3]d`, stmt.Table, stmt.Column, limit)
 	case driver.PostgreSQL, "postgres":
 		return fmt.Sprintf(`WITH rows AS (
-SELECT %[1]s FROM %[2]s WHERE %[3]s < :time ORDER BY %[3]s LIMIT %[4]d
+SELECT %[1]s FROM %[2]s WHERE environment_id = :environment_id AND %[3]s < :time ORDER BY %[3]s LIMIT %[4]d
 )
 DELETE FROM %[2]s WHERE %[1]s IN (SELECT %[1]s FROM rows)`, stmt.PK, stmt.Table, stmt.Column, limit)
 	default:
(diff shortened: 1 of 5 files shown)


julianbrost:
To remove old rows per environment, adding `environment_id` (as first row) to the following indices should help with query times:

```
  INDEX idx_notification_history_send_time (send_time DESC) COMMENT 'Notification list filtered/ordered by send_time'
  INDEX idx_acknowledgement_history_clear_time (clear_time) COMMENT 'Filter for history retention'
  INDEX idx_comment_history_remove_time (remove_time) COMMENT 'Filter for history retention'
  INDEX idx_downtime_history_end_time (end_time) COMMENT 'Filter for history retention'
  INDEX idx_flapping_history_end_time (end_time) COMMENT 'Filter for history retention'
  INDEX idx_state_history_event_time (event_time) COMMENT 'Filter for history retention'
```

So for `notification_history` probably a new index should be added, unless we can confirm that `environment_id, send_time (DESC)` (not sure why DESC is there but shouldn't really make a difference) is also fine for icingadb-web.

For the others, I'd just add the `environment_id`. According to the comments, the purpose of these indices is use in the retention feature, unless of course there are objections that some of these happen to be also useful for icingadb-web but only without `environment_id` in it.

@nilmerg Can you please comment on the need of these indices in icingadb-web?

nilmerg:
Only `idx_notification_history_send_time` should be of interest for Icinga DB Web. The others were not there before the retention feature (I don't have them yet), so they're exclusive to it I guess. Though, for `idx_notification_history_send_time` to have an effect in Icinga DB Web, the latter needs to also incorporate the environment_id into the sort rule if it's part of the index, right? I'd do that only at a last resort.

julianbrost:
Yes, for the indices I propose to be useful, queries need some extra `WHERE environment_id = ?`. This is something that might happen anyways when icingadb-web gets full multi-env support. Until then, letting multiple environments write to the same database it probably of limited use. So an alternative we could consider would be not updating any indices for now but only once there's proper multi-env support everywhere.

For the index change to be of any use in the retention feature, there has to be another environment with longer retention (then the updated indices would save filtering out all older entries from that other environment in each iteration), if there's only one environment, it basically just makes the index larger.

lippserd (approved):

PR #480, opened 2022-05-11
Retention: periodically log number of items removed

julianbrost:
Other components log this information periodically at level info, whereas the retention feature was completely quiet at that level. It's logged at the interval the retention cleanup is run rather than the global configured logging interval since retention cleanup is done less often.

### Before

Nothing logged at level info.

### After

```
2022-05-11T09:30:08.862+0200	INFO	history-retention	Removed 5 old state history items
```

### DIFF
--- a/pkg/icingadb/history/retention.go
+++ b/pkg/icingadb/history/retention.go
@@ -129,6 +129,10 @@ func (r *Retention) StartWithCallback(ctx context.Context, c func(table string, 
 				return
 			}
 
+			if rs.Count > 0 {
+				r.logger.Infof("Removed %d old %s history items", rs.Count, category)
+			}
+
 			if c != nil {
 				c(stmt.Table, rs)
 			}


julianbrost:
@cla-bot check

pkg/icingadb/history/retention.go
```diff
@@ -129,6 +129,10 @@ func (r *Retention) StartWithCallback(ctx context.Context, c func(table string,
 				return
 			}
 
+			if rs.Count > 0 {
+				r.logger.Infof("Removed %d old %s history items", rs.Count, category)
```
lippserd, line 133: I think there's no harm in logging `olderThan` and `rs.Rounds` (via `Infow()`) as well. What do you think?

julianbrost: I left it out to look more consistent with the other messages like

```
INFO	runtime-updates	Upserted 170 ServiceState items
```

Initially, I had it in there as part of the message and it looked too cluttered to me. I'll try if I like `Infow()` better. `rs.Rounds` sounds like too much details for info logging to me though.

lippserd: I just noticed that we don't use `Rounds` anywhere, and now that we have this log message I thought it might be good to include it as an additional argument, even if it's INFO.

julianbrost: Wasn't `Rounds` primarily added to check it in tests which it is no longer since #467? I'd say this is more of an internal value (for other syncs, we also don't log like in how many chunks things were synced) and the value will always be `ceil(items / count)` anyways. So another option would be to also remove this in #481 (i.e. just return the number of rows deleted instead of the struct).

lippserd: > Wasn't `Rounds` primarily added to check it in tests which it is no longer since #467? I'd say this is more of an internal value (for other syncs, we also don't log like in how many chunks things were synced) and the value will always be `ceil(items / count)` anyways.

True.

> Wasn't `Rounds` primarily added to check it in tests which it is no longer since #467? I'd say this is more of an internal value (for other syncs, we also don't log like in how many chunks things were synced) and the value will always be `ceil(items / count)` anyways. So another option would be to also remove this in #481 (i.e. just return the number of rows deleted instead of the struct).

Sounds good to me.

lippserd (approved):

PR #482, opened 2022-05-11
Add CommandArgument#Separator

Al2Klimov:
## Blocked by

(not really)

* [ ] Icinga/icinga2#7919

### DIFF
--- a/pkg/icingadb/db.go
+++ b/pkg/icingadb/db.go
@@ -114,9 +114,9 @@ func (db *DB) BuildInsertStmt(into interface{}) (string, int) {
 	columns := db.BuildColumns(into)
 
 	return fmt.Sprintf(
-		`INSERT INTO "%s" (%s) VALUES (%s)`,
+		`INSERT INTO "%s" ("%s") VALUES (%s)`,
 		utils.TableName(into),
-		strings.Join(columns, ", "),
+		strings.Join(columns, `", "`),
 		fmt.Sprintf(":%s", strings.Join(columns, ", :")),
 	), len(columns)
 }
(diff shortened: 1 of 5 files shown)


julianbrost (changes requested):
Misses the MySQL schema upgrade.

julianbrost (changes requested):
`separator` is a [reserved word](https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-S) in MySQL. I'd just put it in <code>``</code> (rather than choosing a different name), that's already done in another place anyways (`notification_history.text`).

julianbrost:
```
2022-05-17T08:32:32.133Z	FATAL	icingadb	Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'separator, description, argument_key_override, repeat_key, set_if, environment_i' at line 1
```

Do we have any special handling for the `text` column name? Or can you use `text` in a column list without quoting but for `separator`, you'd have to quote?

Al2Klimov:
1. No.
2. Yes:

```
MariaDB [idb]> insert into lolcat(text) values(1);
ERROR 1146 (42S02): Table 'idb.lolcat' doesn't exist
MariaDB [idb]> insert into lolcat(separator) values(1);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'separator) values(1)' at line 1
MariaDB [idb]>
```

"sep" ftw?

Al2Klimov:
Or quoting ftw?

lippserd:
> (how does it handle table names for PostgreSQL at the moment?)

We have a hack in order to quote it. Reserved words as column names are not a problem as we're always selecting fully qualified, i.e. `table_alias.column_name`.

lippserd:
> We have a hack in order to quote it.

https://github.com/Icinga/icingadb-web/pull/508/commits/f251891f0f78b8c998e6880fb3cabda314d6d319

julianbrost:
I have a slight preference towards `separator`, but only if it's possible with reasonable effort. So if nobody objects, I'd say try to add quoting, but get back to the discussion if this needs more complex changes than maybe writing some `JoinColumnNames` function to replace calls to `strings.Join(..., ", ")`.

lippserd (approved):

julianbrost (approved):

PR #484, opened 2022-05-13
Split Redis address into host and port

lippserd:
TBD

Split Redis address into host and port

Now also supports Unix domain sockets and the port is set to 6379 by
default, but the example configuration explicitly specifies 6380 since
that is the default port of the icingadb-redis package.

Expect serverName only in TLS.MakeConfig()

With recent changes to Unix domain socket support, port and address
handling in SQL and Redis configuration, the function can now be reduced
to just expecting the server name.


julianbrost:
> Now also supports Unix domain sockets

That was actually already possible before, go-redis auto-detects this internally.

> port is set to 6379 by default, but the example configuration explicitly specifies 6380 since that is the default port of the icingadb-redis package.

Why this instead of making 6380 the default?

lippserd:
> > Now also supports Unix domain sockets
> 
> That was actually already possible before, go-redis auto-detects this internally.

I see. But had to write the code anyway.

> > port is set to 6379 by default, but the example configuration explicitly specifies 6380 since that is the default port of the icingadb-redis package.
> 
> Why this instead of making 6380 the default?

I thought that's just consistent as with the default ports for MySQL and PostgreSQL and works ootb with Docker.

config.yml.example
```diff
@@ -8,8 +8,17 @@ database:
...
+# This is the same connection as configured in the 'icingadb' feature of the primary Icinga node.
+# High availability setups require a dedicated Icinga DB instance per primary node,
+# each connected to the Redis that the respective Icinga node writes to.
 redis:
-  address: localhost:6380
+  # Redis host or absolute Unix socket path.
+  host: localhost
+
+  # Redis port. Defaults to '6379' which is the standard Redis port,
+  # but explicitly set to '6380' here as that is the port of the icingadb-redis package.
```
julianbrost, line 20: Suggestion: "Redis port. The Redis server provided by the icingadb-redis package listens on port 6380. If not specified, the Redis standard port 6379 is used instead."

doc/03-Configuration.md
```diff
@@ -11,7 +11,8 @@ Configuration of the Redis that Icinga writes to.
 
 Option                   | Description
 -------------------------|-----------------------------------------------
-address                  | **Required.** Redis host:port address or absolute Unix socket path.
+host                     | **Required.** Redis host or absolute Unix socket path.
+port                     | **Optional.** Redis port. Defaults to `6379`. Specify `6380` when using the `icingadb-redis` package.
```
julianbrost, line 15: Should be "if" instead of "when", [I guess](https://dictionary.cambridge.org/grammar/british-grammar/if-or-when).

tests/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/icinga/icingadb/tests
 
 go 1.16
 
+replace github.com/icinga/icinga-testing => github.com/icinga/icinga-testing v0.0.0-20220513144038-4aef51dce82a
```
julianbrost, line 5: I just approved and merged https://github.com/Icinga/icinga-testing/pull/24, so just go ahead and just do a regular update instead.

julianbrost (approved):

PR #489, opened 2022-05-24
Streamline IsRetryable functions and retry DNS errors

lippserd:
This PR aligns the behavior of the IsRetryable functions for Redis and database. It now also checks if the error implements timeout, even if the error implements temporary, where previously only the temporary bool was returned and timeout was ignored. The idea here is that an error could implement both, returning false for temporary and true for timeout. Also, all DNS errors are now considered retryable.

### DIFF
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"github.com/icinga/icingadb/pkg/backoff"
 	"github.com/pkg/errors"
+	"net"
+	"syscall"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/retry/retry.go
```diff
@@ -81,3 +83,31 @@ func WithBackoff(
 		}
 	}
 }
+
+// Opinionated considers timeouts, temporary, DNS and connection refused errors as retryable.
+func Opinionated(err error) bool {
```
julianbrost, line 88: I don't like the name, doesn't really say what it does. Why not simply `IsRetryable`, maybe adding something like "common" or "generic" (to state that it catches the most common cases and others may choose to extend it)?

pkg/retry/retry.go
```diff
@@ -81,3 +83,31 @@ func WithBackoff(
...
+	if t := temporary(nil); errors.As(err, &t) {
+		return t.Temporary()
+	}
+
+	type timeout interface {
+		Timeout() bool
+	}
+	if t := timeout(nil); errors.As(err, &t) {
+		return t.Timeout()
+	}
```
julianbrost, line 110: I'd consistently define variable before or in the if here. (Hint: you can also do `var temporary interface { ... }`, then you don't even need to explicitly do `temporary(nil)`.

julianbrost (changes requested):

julianbrost (approved):

PR #498, opened 2022-06-10
Rename config.yml.example to config.example.yml

julianbrost:
This makes more tools recognize this as a YAML file and enable syntax highlighting for example.


lippserd (changes requested):
`./doc/03-Configuration.md:4` needs an update as well.

PR #499, opened 2022-06-15
Add environment_id to indices for history retention

### LINKED ISSUE #485: Crash due to MySQL deadlock in retention query
While testing #484 (I was using 2532d06e0ae92547f1019a49e3fc47558a9f53c3 to be precise), I was randomly greeted by the following crash. As that PR does nothing with retention at all, I presume the same may happen on the current master as well.

```
2022-05-16T16:25:52.784+0200	FATAL	icingadb	Error 1213: Deadlock found when trying to get lock; try restarting transaction
can't perform "DELETE FROM sla_history_state WHERE environment_id = :environment_id AND event_time < :time\nORDER BY event_time LIMIT 5000"
	/home/jbrost/dev/icingadb/internal/internal.go:30
github.com/icinga/icingadb/pkg/icingadb.(*DB).CleanupOlderThan
	/home/jbrost/dev/icingadb/pkg/icingadb/cleanup.go:51
github.com/icinga/icingadb/pkg/icingadb/history.(*Retention).Start.func1
	/home/jbrost/dev/icingadb/pkg/icingadb/history/retention.go:188
github.com/icinga/icingadb/pkg/periodic.Start.func1
	/home/jbrost/dev/icingadb/pkg/periodic/periodic.go:78
runtime.goexit
	/usr/lib/go/src/runtime/asm_amd64.s:1571
exit status 1
```

Not sure why that happened or how to reproduce it, but that shouldn't bring down the process for sure.

yhabteab:
When UPSERT and DELETE statements are executed at the same time, a
deadlock can occur if both want to get an exclusive lock on one of the
PRIMARY KEY index pages. This happens with DELETE statements when there
is no suitable index for the columns used in the WHERE clause, which is
true for our history retention queries since commit https://github.com/Icinga/icingadb/commit/eccac78ff31186430a33d65d286eb91ad7d20d05. This PR
fixes the problem by adding a suitable index for the columns used in
these queries.

fixes #485

### DIFF
--- a/schema/mysql/schema.sql
+++ b/schema/mysql/schema.sql
@@ -1113,7 +1113,8 @@ CREATE TABLE notification_history (
 
   PRIMARY KEY (id),
 
-  INDEX idx_notification_history_send_time (send_time DESC) COMMENT 'Notification list filtered/ordered by send_time'
+  INDEX idx_notification_history_send_time (send_time DESC) COMMENT 'Notification list filtered/ordered by send_time',
+  INDEX idx_notification_history_env_send_time (environment_id, send_time) COMMENT 'Filter for history retention'
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
 
 CREATE TABLE user_notification_history (
@@ -1150,7 +1151,7 @@ CREATE TABLE state_history (
 
   PRIMARY KEY (id),
 
-  INDEX idx_state_history_event_time (event_time) COMMENT 'Filter for history retention'
+  INDEX idx_state_history_env_event_time (environment_id, event_time) COMMENT 'Filter for history retention'
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
 
 CREATE TABLE downtime_history (
@@ -1180,7 +1181,7 @@ CREATE TABLE downtime_history (
 
   PRIMARY KEY (downtime_id),
 
-  INDEX idx_downtime_history_end_time (end_time) COMMENT 'Filter for history retention'
+  INDEX idx_downtime_history_env_end_time (environment_id, end_time) COMMENT 'Filter for history retention'
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
 
 CREATE TABLE comment_history (
@@ -1204,7 +1205,7 @@ CREATE TABLE comment_history (
 
   PRIMARY KEY (comment_id),
 
-  INDEX idx_comment_history_remove_time (remove_time) COMMENT 'Filter for history retention'
+  INDEX idx_comment_history_env_remove_time (environment_id, remove_time) COMMENT 'Filter for history retention'
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
 
 CREATE TABLE flapping_history (
@@ -1224,7 +1225,7 @@ CREATE TABLE flapping_history (
 
   PRIMARY KEY (id),
 
-  INDEX idx_flapping_history_end_time (end_time) COMMENT 'Filter for history retention'
+  INDEX idx_flapping_history_env_end_time (environment_id, end_time) COMMENT 'Filter for history retention'
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
 
 CREATE TABLE acknowledgement_history (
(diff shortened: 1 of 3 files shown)


julianbrost (changes requested):
Looks like this was somewhat inconsistent before already, so not the fault of this PR, but probably each of the following tables should have an index on `(environment_id, $Column)`:
https://github.com/Icinga/icingadb/blob/8f31ac143875498797725adb9bfacf3d4f64ede2/pkg/icingadb/history/retention.go#L28-L93

Looks like those are missing for `notification` and `sla_history_downtime`, see:
* `grep 'INDEX.*history.*environment_id' schema/mysql/schema.sql`
* `grep 'CREATE INDEX.*history.*environment_id' schema/pgsql/schema.sql`

schema/pgsql/schema.sql
```diff
@@ -1787,6 +1788,7 @@ COMMENT ON COLUMN notification_history.service_id IS 'service.id';
 COMMENT ON COLUMN notification_history.notification_id IS 'notification.id';
 
 COMMENT ON INDEX idx_notification_history_send_time IS 'Notification list filtered/ordered by send_time';
+COMMENT ON INDEX idx_notification_history_environment_id IS 'Filter for sla history retention';
```
julianbrost, line 1791: This table is not related to SLA.

julianbrost (changes requested):
Regarding naming: looks like we don't really follow a consistent pattern there, using (just) the first column in the name seems a common choice, so probably fine.

julianbrost (changes requested):
That's not what I tried to say with that pattern. They should be named `idx_notification_history_env_send_time` for example.

schema/mysql/schema.sql
```diff
@@ -1180,7 +1182,8 @@ CREATE TABLE downtime_history (
 
   PRIMARY KEY (downtime_id),
 
-  INDEX idx_downtime_history_end_time (end_time) COMMENT 'Filter for history retention'
+  INDEX idx_downtime_history_end_time (end_time) COMMENT 'Downtime history filtered/ordered by end_time',
```
lippserd, line 1185: This index is free to remove. Web does not provide such a view.

schema/mysql/schema.sql
```diff
@@ -1150,7 +1151,8 @@ CREATE TABLE state_history (
 
   PRIMARY KEY (id),
 
-  INDEX idx_state_history_event_time (event_time) COMMENT 'Filter for history retention'
+  INDEX idx_state_history_event_time (event_time) COMMENT 'State history filtered/ordered by event_time',
```
lippserd, line 1154: This index is free to remove. Web does not provide such a view.

schema/mysql/schema.sql
```diff
@@ -1204,7 +1207,8 @@ CREATE TABLE comment_history (
 
   PRIMARY KEY (comment_id),
 
-  INDEX idx_comment_history_remove_time (remove_time) COMMENT 'Filter for history retention'
+  INDEX idx_comment_history_remove_time (remove_time) COMMENT 'Comment history filtered/ordered by remove_time',
```
lippserd, line 1210: This index is free to remove. Web does not provide such a view.

schema/mysql/schema.sql
```diff
@@ -1224,7 +1228,8 @@ CREATE TABLE flapping_history (
 
   PRIMARY KEY (id),
 
-  INDEX idx_flapping_history_end_time (end_time) COMMENT 'Filter for history retention'
+  INDEX idx_flapping_history_end_time (end_time) COMMENT 'Flapping history filtered/ordered by end_time',
```
lippserd, line 1231: This index is free to remove. Web does not provide such a view.

schema/mysql/schema.sql
```diff
@@ -1246,7 +1251,8 @@ CREATE TABLE acknowledgement_history (
 
   PRIMARY KEY (id),
 
-  INDEX idx_acknowledgement_history_clear_time (clear_time) COMMENT 'Filter for history retention'
+  INDEX idx_acknowledgement_history_clear_time (clear_time) COMMENT 'Acknowledgement history filtered/ordered by clear_time',
```
lippserd, line 1254: This index is free to remove. Web does not provide such a view.

lippserd (changes requested):

lippserd (approved):

julianbrost (changes requested):
Now has a merge conflict that has to be resolved.

julianbrost (changes requested):
Schema upgrade does not produce the same schema as the full schema file:

```diff
@@ -1583,7 +1584,7 @@
   `downtime_end` bigint unsigned NOT NULL COMMENT 'end time of the downtime',
   PRIMARY KEY (`downtime_id`),
   KEY `idx_sla_history_downtime_event` (`host_id`,`service_id`,`downtime_start`,`downtime_end`) COMMENT 'Filter for calculating the sla reports',
-  KEY `idx_sla_history_downtime_env_downtime_end` (`environment_id`,`downtime_end`) COMMENT 'Filter for sla history retention'
+  KEY `idx_sla_history_downtime_env_event_time` (`environment_id`,`downtime_end`) COMMENT 'Filter for sla history retention'
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
 /*!40101 SET character_set_client = @saved_cs_client */;
```

julianbrost (approved):

PR #501, opened 2022-06-20
Make Database/Redis schema mismatch error more user friendly

### LINKED ISSUE #500: Make early error messages more helpful
The current error messages are hard to understand for users and should give advice on what to do.

For example:
* Unexpected database schema version: *please make sure you have applied all database migrations after upgrading Icinga DB*
* Unexpected Redis schema version: *please make sure you are running compatible versions of Icinga 2 and Icinga DB*

Also, these are more cases where spitting out a stack trace gives no additional information and makes it look more like a bug than a misconfiguration.

yhabteab:
fixes #500

### DIFF
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"github.com/go-redis/redis/v8"
 	"github.com/icinga/icingadb/internal/command"
 	"github.com/icinga/icingadb/pkg/common"
@@ -17,6 +18,7 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
+	"net"
 	"os"
 	"os/signal"
 	"sync"
@@ -63,7 +65,7 @@ func run() int {
 	}
 	defer db.Close()
 	{
-		logger.Info("Connecting to database")
+		logger.Infof("Connecting to database at '%s'", net.JoinHostPort(cmd.Config.Database.Host, fmt.Sprint(cmd.Config.Database.Port)))
 		err := db.Ping()
 		if err != nil {
 			logger.Fatalf("%+v", errors.Wrap(err, "can't connect to database"))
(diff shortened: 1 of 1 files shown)


cmd/icingadb/main.go
```diff
@@ -358,7 +358,10 @@ func checkDbSchema(ctx context.Context, db *icingadb.DB) error {
 	}
 
 	if version != expectedDbSchemaVersion {
-		return errors.Errorf("expected database schema v%d, got v%d", expectedDbSchemaVersion, version)
+		return errors.Errorf(
+			"Unexpected database schema version: v%d (expected v%d). Please make sure you have applied all database"+
+				"migrations after upgrading Icinga DB", version, expectedDbSchemaVersion,
```
julianbrost, line 363: The resulting message misses a space (same thing with the other message below).

julianbrost (changes requested):
Also, looks like this PR changes nothing about the second thing I suggested in #500:
> Also, these are more cases where spitting out a stack trace gives no additional information and makes it look more like a bug than a misconfiguration.

cmd/icingadb/main.go
```diff
@@ -63,7 +66,9 @@ func run() int {
 	}
 	defer db.Close()
 	{
-		logger.Info("Connecting to database")
+		logger.Infof("Connecting to database at host '%s'", net.JoinHostPort(
+			cmd.Config.Database.Host, strconv.FormatInt(int64(cmd.Config.Database.Port), 10)),
```
julianbrost, line 70: `fmt.Sprint(cmd.Config.Database.Port)` would do the same here but is shorter/easier to read.

cmd/icingadb/main.go
```diff
@@ -79,7 +84,9 @@ func run() int {
 		logger.Fatalf("%+v", errors.Wrap(err, "can't create Redis client from config"))
 	}
 	{
-		logger.Info("Connecting to Redis")
+		logger.Infof("Connecting to Redis at host '%s'", net.JoinHostPort(
+			cmd.Config.Redis.Host, strconv.FormatInt(int64(cmd.Config.Redis.Port), 10)),
+		)
```
julianbrost, line 89: Same two comments apply here just as well.

cmd/icingadb/main.go
```diff
@@ -379,9 +392,11 @@ func monitorRedisSchema(logger *logging.Logger, rc *icingaredis.Client, pos stri
 // checkRedisSchema verifies rc's icinga:schema version.
 func checkRedisSchema(logger *logging.Logger, rc *icingaredis.Client, pos string) (newPos string, err error) {
 	if pos == "0-0" {
-		defer time.AfterFunc(3*time.Second, func() { logger.Info("Waiting for current Redis schema version") }).Stop()
+		defer time.AfterFunc(3*time.Second, func() {
+			logger.Info("Waiting for Icinga 2 to write into Redis. Please make sure you have started Icinga 2 and Icinga DB feature is enabled")
```
julianbrost, line 396: "[...] and **the** Icinga DB feature [...]"

Additionally, I'd either end both sentences with a period here, or just do the same ", please" as with the other messages.

julianbrost (approved):

PR #503, opened 2022-06-22
parseString(): Don't Repeat Yourself

Al2Klimov:


pkg/structify/structify.go
```diff
@@ -118,66 +119,61 @@ func parseString(src string, dest interface{}) error {
...
+	if err == nil {
+		*dest = T(f)
+	}
+
+	return err
+}
+
+func bitSizeOf[T any]() int {
+	var x T
+	return int(unsafe.Sizeof(x) * 8)
```
julianbrost, line 178: My hope was that we could get rid of `unsafe` in this file and thus the whole project :(

Al2Klimov: unsafe.Sizeof isn’t actually _unsafe_.

julianbrost: Yes, but still means that the project doesn't pass the `git grep unsafe` test immediately by returning no results but rather returning something you have to inspect.

julianbrost (approved):
Not 100% happy with this but I think I can live with it and hope that Go brings a nicer function like `ParseNumber[T]` in the future, replacing these cumbersome functions in `strconv`.

I tried to find some alternatives but the only thing I could find in the standard library was `json.Unmarshal`, which is a magnitude slower than `strconv` unfortunately.

PR #506, opened 2022-06-24
Icinga DB Check: Proposed changes to #464

julianbrost:
My proposed changes to #464 (corresponding Icinga 2 PR: https://github.com/Icinga/icinga2/pull/9416)

### DIFF
--- a/pkg/icingaredis/telemetry/heartbeat.go
+++ b/pkg/icingaredis/telemetry/heartbeat.go
@@ -16,6 +16,7 @@ import (
 	"runtime/metrics"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 )
@@ -30,13 +31,51 @@ type SuccessfulSync struct {
 	DurationMilli int64
 }
 
-type DbConnErr struct {
-	Message    string
-	SinceMilli int64
+// currentDbConnErr stores ongoing errors from database connections.
+var currentDbConnErr struct {
+	mu         sync.Mutex
+	message    string
+	sinceMilli int64
 }
 
-// CurrentDbConnErr is to be updated by the DB connector.
-var CurrentDbConnErr com.Atomic[DbConnErr]
+// UpdateCurrentDbConnErr updates the current error information stored in currentDbConnErr.
+func UpdateCurrentDbConnErr(err error) {
+	now := time.Now().UnixMilli()
+
+	currentDbConnErr.mu.Lock()
+	defer currentDbConnErr.mu.Unlock()
+
+	if currentDbConnErr.sinceMilli >= now {
+		// Already updated with a more recent error, ignore this one.
+		return
+	}
+
+	message := ""
+	if err != nil {
+		message = err.Error()
+	}
+
+	if currentDbConnErr.message == message {
+		// Error stayed the same, no update needed, keeping the old timestamp.
+		return
+	}
+
+	if currentDbConnErr.message == "" || message == "" {
+		// Either first error or recovery from an error, update timestamp.
+		currentDbConnErr.sinceMilli = now
+	}
+
+	currentDbConnErr.message = message
+}
+
+// GetCurrentDbConnErr returns the last error message (or the empty string if not in an error state) and a timestamp in
+// milliseconds of the last change from OK to error or from error to OK.
+func GetCurrentDbConnErr() (string, int64) {
+	currentDbConnErr.mu.Lock()
+	defer currentDbConnErr.mu.Unlock()
+
+	return currentDbConnErr.message, currentDbConnErr.sinceMilli
+}
 
 // OngoingSyncStartMilli is to be updated by the main() function.
 var OngoingSyncStartMilli int64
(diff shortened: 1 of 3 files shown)


pkg/icingaredis/telemetry/heartbeat.go
```diff
@@ -146,5 +109,57 @@ func StartHeartbeat(
...
+		}
+	}
+
+	return m
+}
+
+func (g *GoMetrics) PerformanceData() string {
+	metrics.Read(g.samples)
+
+	buf := bytes.NewBuffer(nil)
```
Al2Klimov, line 149: _May_ be just the zero-value of the struct.

Al2Klimov, line 149: _Should_ be a strings.Builder...

pkg/icingaredis/telemetry/heartbeat.go
```diff
@@ -146,5 +109,57 @@ func StartHeartbeat(
...
+
+		switch value.Value.Kind() {
+		case metrics.KindUint64:
+			_, _ = fmt.Fprintf(buf, "%s=%d%s", g.names[i], value.Value.Uint64(), g.units[i])
+		case metrics.KindFloat64:
+			_, _ = fmt.Fprintf(buf, "%s=%f%s", g.names[i], value.Value.Float64(), g.units[i])
+		}
+	}
+
+	return buf.String()
```
Al2Klimov, line 164: ... due to this line.

pkg/icingaredis/telemetry/heartbeat.go
```diff
@@ -146,5 +109,57 @@ func StartHeartbeat(
...
+
+func NewGoMetrics() *GoMetrics {
+	m := &GoMetrics{}
+
+	forbiddenRe := regexp.MustCompile(`\W`)
+
+	for _, d := range metrics.All() {
+		switch d.Kind {
+		case metrics.KindUint64, metrics.KindFloat64:
+			name := "go_" + strings.TrimPrefix(forbiddenRe.ReplaceAllString(d.Name, "_"), "_")
```
Al2Klimov, line 126: This is for the case there are <1 _ characters at the beginning, right? Then you could also handle >1 ones by using another trim function.

pkg/icingaredis/telemetry/heartbeat.go
```diff
@@ -146,5 +109,57 @@ func StartHeartbeat(
...
+}
+
+func (g *GoMetrics) PerformanceData() string {
+	metrics.Read(g.samples)
+
+	buf := bytes.NewBuffer(nil)
+
+	for i, value := range g.samples {
+		if i > 0 {
+			_, _ = fmt.Fprint(buf, " ")
```
Al2Klimov, line 153: _Could_ be speed up and simplified by WriteByte.

pkg/icingaredis/telemetry/heartbeat.go
```diff
@@ -146,5 +109,57 @@ func StartHeartbeat(
...
+		}
+
+		switch value.Value.Kind() {
+		case metrics.KindUint64:
+			_, _ = fmt.Fprintf(buf, "%s=%d%s", g.names[i], value.Value.Uint64(), g.units[i])
+		case metrics.KindFloat64:
+			_, _ = fmt.Fprintf(buf, "%s=%f%s", g.names[i], value.Value.Float64(), g.units[i])
```
Al2Klimov, line 160: _Should_ be deduplicated a bit.

pkg/icingaredis/telemetry/heartbeat.go
```diff
@@ -146,5 +109,57 @@ func StartHeartbeat(
...
+
+func (g *GoMetrics) PerformanceData() string {
+	metrics.Read(g.samples)
+
+	buf := bytes.NewBuffer(nil)
+
+	for i, value := range g.samples {
+		if i > 0 {
+			_, _ = fmt.Fprint(buf, " ")
+		}
```
Al2Klimov, line 154: And if it's a histogram?

pkg/icingaredis/telemetry/heartbeat.go
```diff
@@ -146,5 +109,57 @@ func StartHeartbeat(
 	})
 }
 
-// Assert interface compliance.
-var _ json.Marshaler = jsonMetricValue{}
+type GoMetrics struct {
```
Al2Klimov, line 112: _Could_ be unexported.

pkg/icingaredis/telemetry/heartbeat.go
julianbrost: Do you see much potential for deduplication here that wouldn't break static checks for the format strings?

Al2Klimov: Oh. 🐘

pkg/icingaredis/telemetry/heartbeat.go
Al2Klimov: Oh. 🐘

pkg/icingaredis/telemetry/heartbeat.go
julianbrost: You mean the use of `TrimPrefix()`?

I expect this code to always to always have the same effect as `name := "go_" + forbiddenRe.ReplaceAllString(d.Name[1:], "_")` but using a `strings.Trim*()` function is more robust (in case there's an empty string or one that starts with another character than expected, but both shouldn't happen with the `metrics` package anyways).

But sure, `TrimLeft()` will do the same on all inputs expected from the `metrics` package but would handle more strange inputs correctly (as in not producing `go__something`).

PR #508, opened 2022-06-29
icinga:*:state: rename state to soft_state

Al2Klimov:


Al2Klimov:
## Blocked by

* [ ] Icinga/icinga2#9420

PR #510, opened 2022-06-29
MySQL schema: Drop ROW_FORMAT=DYNAMIC

lippserd:


julianbrost:
(Translated) quote:

> doesn't work

-- @lippserd

PR #525, opened 2022-09-13
Save memory during config sync via SyncSubject#FactoryForDelta()

Al2Klimov:
Code comment TL;DR: Allocate the same amount of smaller data structures

## Peek of reserved memory

* Before: 3.5G
* After: 2.8G

### DIFF
--- a/pkg/common/sync_subject.go
+++ b/pkg/common/sync_subject.go
@@ -2,6 +2,7 @@ package common
 
 import (
 	"github.com/icinga/icingadb/pkg/contracts"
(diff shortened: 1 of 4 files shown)


julianbrost (approved):

julianbrost:
@Al2Klimov IIRC, at some point you said you were looking into making this a plain map from ID to checksum (i.e. no `Entity`, just the sole checksum value). What happened to that plan?

Al2Klimov:
IDK

PR #526, opened 2022-09-15
Free memory ASAP

Al2Klimov:


Al2Klimov:
Didn’t notice any improvements compared to #525.

julianbrost (commented):
Had a quick look at it and looks like some might help to allow the GC to reclaim some memory earlier. But

> Didn’t notice any improvements compared to #525.

is there a point in doing it then? Maybe this is because when the sync begins, all of that is loaded at once, so the peak will stay the same, even if you free something a bit earlier later on (but this is pure speculation)?

Al2Klimov:
> is there a point in doing it then?

Feel free to close it. IMAO yet another of my abject tries to reduce mem spikes.

PR #536, opened 2022-11-02
Improve documentation for migration tool

julianbrost:
1. Moves the example config to a separate file one can just download.
2. Establishes some kind of standard workflow for the migration.

## TODO
* [x] ~Figure out if we want to keep the endpoint setting or just use NULL ([context](https://github.com/Icinga/icingadb/pull/253#issuecomment-1296932224)), then either write documentation for it or remove it altogether.~
* [x] Remove `endpoint` option in other PR (#541)

### DIFF
--- a/doc/06-Migration.md
+++ b/doc/06-Migration.md
@@ -1,90 +1,102 @@
 # Migration from IDO
 
-The Icinga DB Migration commandline tool migrates history data from [IDO] to
-Icinga DB. Or, more precisely: from the IDO SQL database to the Icinga DB one.
+Migrating from the IDO feature to Icinga DB starts by setting up Icinga DB. To
+do so, please follow the [installation instructions]. The Icinga DB feature can
+be enabled in parallel to the IDO, allowing you to perform the migration while
+the IDO is still running.
 
-!!! info
+After setting up Icinga DB, all Icinga objects and their current state should
+already show up in Icinga DB Web as this information is synced from Icinga 2.
+At this point, the old host and service history is missing in Icinga DB. If it
+is desired to keep it, this information has to be migrated explicitly from the
+old IDO database. To do so, follow the instructions below.
 
-    Everything else is already populated by Icinga DB itself.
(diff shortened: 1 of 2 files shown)


doc/06-Migration.md
```diff
@@ -1,90 +1,102 @@
...
+`env` option in the section `icinga2`.
+
+#### Database Connection
+
+The migration tool needs to access both the IDO and the Icinga DB databases.
+Please specify the connection details in the corresponding `ido` and `icingadb`
+sections of the configuration.
+
+Both the IDO and Icinga DB support MySQL and PostgreSQL. You can migrate from
+and to both types, including from one type to the other.
```
Al2Klimov, line 42: I still like the idea to get this fact underlined...

doc/icingadb-migration.example.yml
```diff
@@ -0,0 +1,25 @@
+icinga2:
+  # Content of /var/lib/icinga2/icingadb.env
+  env: "da39a3e..."
+
+# IDO database
+ido:
+  type: mysql # or "pgsql" for PostgreSQL
```
Al2Klimov, line 7: ... by switching the DB types here. :)

(After all this one's migrating from IPv4 to IPv6 as well.)

Al2Klimov (approved):

doc/icingadb-migration.example.yml
julianbrost: As we support both databases, I don't think we should incentivize a switch here. So if a user has the desire to switch, they will probably look for it, others may be confused why the example migrates from MySQL to PostgreSQL.

Al2Klimov: Oh, **I think** we should incentivize a switch – from PostgreSQL **to MySQL**.

julianbrost: No. It would make no sense for Icinga DB to support PostgreSQL if we would recommend to migrate away from it.

Issue #553, opened 2022-12-19
icingadb-migrate: can't insert because key constraint fails "fk_history_flapping_history"


xeiss:
## Describe the bug
I wanted to migrate my IDO from 1. Januar 2021 until I started the icingadb (maybe more after that). But the icingadb-migrate tool stops with an Insert issue and so I can not migrate the important downtime and state infos. Here is the output from my last start with using the already created cache file.

```
# icingadb-migrate -c icingadb-migration.yml -t ~/icingadb-migration.cache
2022-12-19T10:42:24.965+0100    INFO    icingadb-migrate/main.go:78     Starting IDO to Icinga DB history migration
2022-12-19T10:42:24.965+0100    INFO    icingadb-migrate/main.go:175    Connecting to databases
2022-12-19T10:42:24.971+0100    INFO    icingadb-migrate/main.go:141    Preparing cache
2022-12-19T10:42:24.975+0100    INFO    icingadb-migrate/main.go:93     Computing progress
2022-12-19T10:42:26.232+0100    INFO    icingadb-migrate/main.go:330    Counted migrated IDO events     {"type": "notification", "migrated": 878, "total": 878}
2022-12-19T10:42:26.245+0100    INFO    icingadb-migrate/main.go:330    Counted migrated IDO events     {"type": "ack & comment", "migrated": 3306, "total": 3306}
2022-12-19T10:42:26.290+0100    INFO    icingadb-migrate/main.go:330    Counted migrated IDO events     {"type": "flapping", "migrated": 0, "total": 32270}
2022-12-19T10:42:26.447+0100    INFO    icingadb-migrate/main.go:330    Counted migrated IDO events     {"type": "downtime", "migrated": 0, "total": 108008}
2022-12-19T10:42:26.691+0100    INFO    icingadb-migrate/main.go:330    Counted migrated IDO events     {"type": "state", "migrated": 0, "total": 739982}
2022-12-19T10:42:26.691+0100    INFO    icingadb-migrate/main.go:103    Filling cache
flapping 100 %    0s  0/s
notification 100 %    0s  0/s
state 100 %    0s  0/s
2022-12-19T10:42:28.539+0100    INFO    icingadb-migrate/main.go:106    Actually migrating
ack & comment 100 %    0s  0/s
downtime   0 % [---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------]   0s  0/s
[...]

PR #554, opened 2022-12-19
convertFlappingRows(): fix foreign key error history -> flapping_history

Al2Klimov:
Don't INSERT IGNORE everything, then UPSERT everything. Instead INSERT IGNORE    flapping_history, UPSERT flapping_history and finally INSERT IGNORE history.

### DIFF
--- a/cmd/icingadb-migrate/main.go
+++ b/cmd/icingadb-migrate/main.go
@@ -368,7 +368,7 @@ func migrateOneType[IdoRow any](
 	c *Config, idb *icingadb.DB, envId []byte, ht *historyType,
 	convertRows func(env string, envId icingadbTypes.Binary,
 		selectCache func(dest interface{}, query string, args ...interface{}), ido *sqlx.Tx,
-		idoRows []IdoRow) (icingaDbInserts, icingaDbUpserts [][]contracts.Entity, checkpoint any),
+		idoRows []IdoRow) (stages []icingaDbOutputStage, checkpoint any),
 ) {
 	var lastQuery string
 	var lastStmt *sqlx.Stmt
@@ -423,30 +423,35 @@ func migrateOneType[IdoRow any](
 		ht, ht.migrationQuery, args, ht.lastId,
 		func(idoRows []IdoRow) (checkpoint interface{}) {
 			// ... convert them, ...
-			inserts, upserts, lastIdoId := convertRows(c.Icinga2.Env, envId, selectCache, ht.snapshot, idoRows)
+			stages, lastIdoId := convertRows(c.Icinga2.Env, envId, selectCache, ht.snapshot, idoRows)
 
 			// ... and insert them:
 
-			for _, op := range []struct {
-				kind     string
-				data     [][]contracts.Entity
-				streamer func(context.Context, <-chan contracts.Entity, ...icingadb.OnSuccess[contracts.Entity]) error
-			}{{"INSERT IGNORE", inserts, idb.CreateIgnoreStreamed}, {"UPSERT", upserts, idb.UpsertStreamed}} {
-				for _, table := range op.data {
-					if len(table) < 1 {
-						continue
-					}
+			for _, stage := range stages {
+				for _, op := range []struct {
+					kind     string
+					data     [][]contracts.Entity
+					streamer func(context.Context, <-chan contracts.Entity, ...icingadb.OnSuccess[contracts.Entity]) error
+				}{
+					{"INSERT IGNORE", stage.inserts, idb.CreateIgnoreStreamed},
+					{"UPSERT", stage.upserts, idb.UpsertStreamed},
+				} {
+					for _, table := range op.data {
+						if len(table) < 1 {
+							continue
+						}
 
-					ch := make(chan contracts.Entity, len(table))
-					for _, row := range table {
-						ch <- row
-					}
+						ch := make(chan contracts.Entity, len(table))
+						for _, row := range table {
+							ch <- row
+						}
 
-					close(ch)
+						close(ch)
 
-					if err := op.streamer(context.Background(), ch); err != nil {
-						log.With("backend", "Icinga DB", "op", op.kind, "table", utils.TableName(table[0])).
-							Fatalf("%+v", errors.Wrap(err, "can't perform DML"))
+						if err := op.streamer(context.Background(), ch); err != nil {
+							log.With("backend", "Icinga DB", "op", op.kind, "table", utils.TableName(table[0])).
+								Fatalf("%+v", errors.Wrap(err, "can't perform DML"))
+						}
 					}
 				}
 			}
--- a/cmd/icingadb-migrate/main.go
+++ b/cmd/icingadb-migrate/main.go
@@ -9,7 +9,6 @@ import (
 	"github.com/creasty/defaults"
 	"github.com/goccy/go-yaml"
 	"github.com/icinga/icingadb/pkg/config"
-	"github.com/icinga/icingadb/pkg/contracts"
 	"github.com/icinga/icingadb/pkg/icingadb"
 	"github.com/icinga/icingadb/pkg/logging"
 	icingadbTypes "github.com/icinga/icingadb/pkg/types"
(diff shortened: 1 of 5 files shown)


julianbrost:
How can this be reproduced/tested?

julianbrost:
Didn't trigger the bug with the following flapping history in the IDO for me (1000 should be start and 1001 be end, so the sequence is end, start, end, end on the same object):
```
MariaDB [icingaido]> select * from icinga_flappinghistory order by event_time;
+--------------------+-------------+---------------------+-----------------+------------+-------------+---------------+-----------+----------------------+---------------+----------------+--------------+---------------------+--------------------+
| flappinghistory_id | instance_id | event_time          | event_time_usec | event_type | reason_type | flapping_type | object_id | percent_state_change | low_threshold | high_threshold | comment_time | internal_comment_id | endpoint_object_id |
+--------------------+-------------+---------------------+-----------------+------------+-------------+---------------+-----------+----------------------+---------------+----------------+--------------+---------------------+--------------------+
|                  3 |           1 | 2021-10-26 14:36:48 |          865855 |       1001 |           1 |             1 |       408 |                   21 |            25 |             30 | NULL         |                   0 |                239 |
|                  4 |           1 | 2022-05-30 08:30:43 |          746691 |       1000 |           0 |             1 |       408 |                 32.5 |            25 |             30 | NULL         |                   0 |                239 |
|                  5 |           1 | 2022-06-07 09:06:13 |          434294 |       1001 |           1 |             1 |       408 |                   21 |            25 |             30 | NULL         |                   0 |                239 |
|                  7 |           1 | 2022-11-14 10:09:03 |          573515 |       1001 |           1 |             1 |       408 |                 22.3 |            25 |             30 | NULL         |                   0 |                239 |
+--------------------+-------------+---------------------+-----------------+------------+-------------+---------------+-----------+----------------------+---------------+----------------+--------------+---------------------+--------------------+
```

This generated a single row in `flapping_history` (and the corresponding two rows in `history`), where the timestamps are 2022-05-30 08:30:43 and 2022-06-07 09:06:13, i.e. the flapping period from `flappinghistory_id` 4 to 5, so it looks like the ones without a start have just been ignored.

```
idb_migration=# select * from flapping_history;
-[ RECORD 1 ]--------------+-------------------------------------------
id                         | \xdceb3d4282ee4f5b72bc66547570db556237ee8c
environment_id             | \x6cc4013ece4ffdbb275a4a763460507d17203146
endpoint_id                | 
object_type                | service
host_id                    | \x2bea57085609682477bde443bff3e4081558cd15
service_id                 | \x52cb639f147b0f429b3c3a18fb19529d50c81816
start_time                 | 1653899443746
end_time                   | 1654592773434
percent_state_change_start | 32.5
percent_state_change_end   | 21
flapping_threshold_low     | 25
flapping_threshold_high    | 30
```

Al2Klimov:
With or without my changes here? If without, what happens _with_ and vice-versa? Did you compare with latest master or the merge base?

julianbrost:
I tested the current master branch.

Al2Klimov:
What you described above IS the bug. The foreign key error does this due to INSERT IGNORE.

julianbrost:
What exactly is the bug? #553 mentions an explicit foreign key error being shown, which I didn't see.

(I just noticed I tested PostgreSQL while the bug report mentions MariaDB and shows a corresponding error message, so I'll retry with that, maybe that makes a difference, but I doubt it.)

julianbrost:
Key to reproducing the issue: set `ido.from` to a time while a checkable was flapping (see https://github.com/Icinga/icingadb/issues/553#issuecomment-1651686651 for more details).

The error goes away with this PR (I didn't check much more so far though).

cmd/icingadb-migrate/main.go
```diff
@@ -423,30 +423,35 @@ func migrateOneType[IdoRow any](
...
+
+						close(ch)
+
+						if err := op.streamer(context.Background(), ch); err != nil {
+							log.With("backend", "Icinga DB", "op", op.kind, "table", utils.TableName(table[0])).
+								Fatalf("%+v", errors.Wrap(err, "can't perform DML"))
+						}
 					}
 				}
 			}
```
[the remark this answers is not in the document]
Al2Klimov: Not bad on first sight, feel free to make a PR into this one (to be rebase-and-merged to avoid additional merge commits).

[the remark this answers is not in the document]
Al2Klimov: ... or into master w/ this one included. I'd prefer this if your complaint doesn't actually block anything here.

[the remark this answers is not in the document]
julianbrost: My preferred way would be that if you approve of that change, you just manually push my commit to this PR (so more similar to the typical "include proposed changes from a review" workflow).

> Not bad on first sight, feel free to make a PR into this one (to be rebase-and-merged to avoid additional merge commits).

That would leave a weird merged PR that wasn't actually merged into any "productive" branch, i.e. show up in "merged PRs without milestone" searches and so on.

> ... or into master w/ this one included. I'd prefer this if your complaint doesn't actually block anything here.

That would mean some merge/approval with "but actually I didn't look at this code in detail because (hopefully) it will be replaced anyways".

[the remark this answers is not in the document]
Al2Klimov: > My preferred way would be that if you approve of that change, you just manually push my commit to this PR (so more similar to the typical "include proposed changes from a review" workflow).

Yeees... but... this isn't just a typo, you know?

> That would leave a weird merged PR that wasn't actually merged into any "productive" branch, i.e. show up in "merged PRs without milestone" searches and so on.

No, I think it wouldn't, e.g.

* Icinga/icinga2#9523

is on the 2.14 milestone, but not in the changelog. I think such and buildfix PRs can also be in the changelog in the future, possibly starting with

* Icinga/icinga2#9830

which is admittedly a bad example as it touches an out-of-changelog file, but you get the point.

> That would mean some merge/approval with "but actually I didn't look at this code in detail because (hopefully) it will be replaced anyways".

Huh? You didn't look at this code in detail, and/but replaced it?

## Anyway.

If you really don't want an additional PR, feel free to push on top of my branch, I have just a few non-critical things about the unit tests.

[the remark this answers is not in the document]
julianbrost: > Yeees... but... this isn't just a typo, you know?

Reviews often include things way more complex than typos.

> > That would leave a weird merged PR that wasn't actually merged into any "productive" branch, i.e. show up in "merged PRs without milestone" searches and so on.
> 
> No, I think it wouldn't, e.g.
> 
> * [RedisConnection: Don't log queries that are going to be discarded icinga2#9523](https://github.com/Icinga/icinga2/pull/9523)
> 
> is on the 2.14 milestone, but not in the changelog.

Well, I already didn't like that PR at that time but you merged it faster than I could say please don't. The main reason I don't like this is that typically, if you see that a PR is merged, you expect that the change was included in the software. But that's only the case if you merge it into `master`/`support/*`. If it's merged into another branch of some PR, and that PR is discarded, the other PR would still show as merged even though the change never made it into the software.

> Huh? You didn't look at this code in detail, and/but replaced it?

I replaced it with simpler code that does what I'd expect to happen there and what I hope your code does. If my code does something different than your code, please tell me.

> ## Anyway.
> If you really don't want an additional PR, feel free to push on top of my branch, I have just a few non-critical things about the unit tests.

Is that anything complex that would need the review tools from GH PRs? I mean I can create a PR to discuss it, I'd just want to avoid having it show up as merged in the end.

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,54 @@
...
+	requireClosedEmpty := func(t *testing.T, ch <-chan int) {
+		t.Helper()
+
+		select {
+		case _, ok := <-ch:
+			require.False(t, ok, "receiving from channel should not return anything")
+		default:
+			require.Fail(t, "receiving should not block")
+		}
+	}
```
Al2Klimov, line 32: These are closures, not regular functions. But why? What are they capturing?

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	t.Run("Nil", func(t *testing.T) {
+		ch := ChanFromSlice[int](nil)
+		require.NotNil(t, ch)
+		requireClosedEmpty(t, ch)
+	})
+
+	t.Run("Empty", func(t *testing.T) {
+		ch := ChanFromSlice([]int{})
```
Al2Klimov, line 41: _In practice_ `[]T{}!=nil`. But could this change _in theory_ so we should use `make([]T,0,1)` as in TestBinary_MarshalJSON, just to be sure to cover both nil and empty?

Wait! _I think_ we should use `make([]T,0,1)` anyway as it's the absolute corner case of an empty slice.

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,54 @@
...
+		require.NotNil(t, ch)
+		requireClosedEmpty(t, ch)
+	})
+
+	t.Run("NonEmpty", func(t *testing.T) {
+		ch := ChanFromSlice([]int{42, 23, 1337})
+		require.NotNil(t, ch)
+		requireReceive(t, ch, 42)
+		requireReceive(t, ch, 23)
+		requireReceive(t, ch, 1337)
```
Al2Klimov, line 51: Ah, yes, the 2.0 version of 23 :)

pkg/utils/utils_test.go
julianbrost: > What are they capturing?

Nothing.

> But why?

Initially scoping (i.e. aren't used elsewhere) but probably looks nicer the other way.

pkg/utils/utils_test.go
Al2Klimov: Admittedly (luckily!) you're right.

> Wait! _I think_ we should use `make([]T,0,1)` anyway as it's the absolute corner case of an empty slice.

But I think we should do this anyway as `cap([]int{})==0` in contrast to make(). Just to provoke the tested code a bit (which is one of the purposes of unit tests).

pkg/utils/utils_test.go
Al2Klimov: > probably looks nicer the other way.

And now you _could_ make them generic. Just for fun. :)

Al2Klimov (commented):
Nothing critical left. Change as above or merge if you're happy with it.

julianbrost (approved):

PR #557, opened 2023-01-18
GHA: handle changed interface of Icinga/docker-icingadb

Al2Klimov:
## Blocked by

* Icinga/docker-icingadb#56

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,12 +10,23 @@ on:
       - published
 
 jobs:
-  docker:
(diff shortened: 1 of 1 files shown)


julianbrost:
Are there any other reasons for this being marked as a draft apart from https://github.com/Icinga/docker-icingadb/pull/56 not being merged so far?

julianbrost:
I don't know how we ended up with these two PRs, but this one has the same contents as #562 and changes nothing compared to what's in the main branch right now.

PR #560, opened 2023-01-25
Fix sla report inconsistencies

### LINKED ISSUE #558: SLA: Ignore objects that have not yet been checked in a given time period
See here a fix for the IDO reports.
https://github.com/Icinga/icingaweb2-module-idoreports/pull/51

yhabteab:
This PR only fixes that the host/service reports with empty history event entries are returning `null` rather than `100`. However, the total report results for the same timeframe are still different depending on whether you select host reports for a week at a time or split them into a 7-day period (in reporting module is called this a `breakdown`), i.e. instead of for a week at a time, they are only generated for one day at a time and that 7 times.

fixes #558
refs https://github.com/Icinga/icingadb-web/issues/713

### DIFF
--- a/schema/mysql/schema.sql
+++ b/schema/mysql/schema.sql
@@ -25,6 +25,7 @@ BEGIN
   DECLARE active_downtimes int unsigned;
   DECLARE problem_time bigint unsigned;
   DECLARE total_time bigint unsigned;
+  DECLARE rowCounts int unsigned;
   DECLARE done int;
   DECLARE cur CURSOR FOR
     (
@@ -69,6 +70,8 @@ BEGIN
         AND ((in_service_id IS NULL AND s.service_id IS NULL) OR s.service_id = in_service_id)
         AND s.event_time > in_start_time
         AND s.event_time < in_end_time
+        AND s.hard_state IS NOT NULL
+        AND s.previous_hard_state IS NOT NULL
     ) UNION ALL (
       -- end event to keep loop simple, values are not used
       SELECT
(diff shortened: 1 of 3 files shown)


schema/mysql/schema.sql
```diff
@@ -69,6 +70,8 @@ BEGIN
         AND ((in_service_id IS NULL AND s.service_id IS NULL) OR s.service_id = in_service_id)
         AND s.event_time > in_start_time
         AND s.event_time < in_end_time
+        AND s.hard_state IS NOT NULL
+        AND s.previous_hard_state IS NOT NULL
```
julianbrost, line 74: Don't these columns already have a `NOT NULL` constraint?

schema/mysql/schema.sql
```diff
@@ -156,7 +160,12 @@ BEGIN
   END LOOP;
   CLOSE cur;
 
-  SET result = 100 * (total_time - problem_time) / total_time;
+  -- row count "1" because of the faked ending result used for the
+  -- cursor loop, whose result set is never used.
+  IF rowCounts > 1 THEN
```
julianbrost, line 165: This would also count downtime rows returned by that query which I doubt you want. But also if this would just count the state rows, this makes the assumption that if there's no event in the interval, there's no valid result to report. Problem with this is that if nothing happens on a checkable, there is no history. So for something that's 100% up, had no downtimes, this would now no longer return a value.

julianbrost (commented):
Can you please share the data from the SLA history tables you were using (so probably one of the hosts you had for testing in #558).

yhabteab:
Superseded by https://github.com/Icinga/icingadb-web/pull/710

PR #566, opened 2023-02-23
Write checkable `create` & `delete` sla history events

yhabteab:


yhabteab:
```go
2023-02-27T08:20:03.534Z        INFO    config-sync     Inserting 11 items of type host
2023-02-27T08:20:03.534Z        INFO    config-sync     Inserting 11 items of type host sla history trails
...
2023-02-27T08:20:03.541Z        INFO    config-sync     Inserting 13 items of type notification usergroup
2023-02-27T08:20:03.542Z        INFO    config-sync     Inserting 50 items of type service
2023-02-27T08:20:03.542Z        INFO    config-sync     Inserting 50 items of type service sla history trails
```

schema/mysql/schema.sql
```diff
@@ -1321,6 +1321,19 @@ CREATE TABLE sla_history_downtime (
   INDEX idx_sla_history_downtime_env_downtime_end (environment_id, downtime_end) COMMENT 'Filter for sla history retention'
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
 
+CREATE TABLE sla_history_trail (
+  id binary(20) NOT NULL,
+  environment_id binary(20) NOT NULL COMMENT 'environment.id',
+  object_type enum('host', 'service') NOT NULL,
+  host_id binary(20) NOT NULL COMMENT 'host.id (may reference already deleted hosts)',
+  service_id binary(20) NOT NULL COMMENT 'service.id (may reference already deleted services)',
```
Al2Klimov, line 1329: NOT NULL?

pkg/icingadb/sync.go
```diff
@@ -163,6 +178,18 @@ func (s Sync) ApplyDelta(ctx context.Context, delta *Delta) error {
 		g.Go(func() error {
 			return s.db.Delete(ctx, delta.Subject.Entity(), delta.Delete.IDs(), OnSuccessIncrement[any](stat))
 		})
+
+		switch subjectType.(type) {
+		case *v1.Host, *v1.Service:
+			s.logger.Infof("Inserting %d items of type %s sla history trails of type delete", len(delta.Delete), utils.Key(utils.Name(delta.Subject.Entity()), ' '))
+			var slaTrails <-chan contracts.Entity
```
Al2Klimov, line 185: Isn’t this line redundant?

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+	})
+
+	return entities, com.WaitAsync(g)
+}
+
+// GenerateBinaryId generates a 20 byte length random id
+func generateBinaryId() (types.Binary, error) {
+	id := make([]byte, 20)
+	_, err := rand.Read(id)
```
Al2Klimov, line 77: If(!) your ID is allowed to be random, UUIDv4 is your friend.

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,80 @@
...
+	ObjectType               string          `json:"object_type"`
+	HostId                   types.Binary    `json:"host_id"`
+	ServiceId                types.Binary    `json:"service_id"`
+	EventTime                types.UnixMilli `json:"event_time"`
+	EventType                string          `json:"event_type"`
+}
+
+func CheckableToSlaTrailEntities(ctx context.Context, checkables <-chan contracts.Entity, eventType string) (<-chan contracts.Entity, <-chan error) {
+	entities := make(chan contracts.Entity)
+	g, ctx := errgroup.WithContext(ctx)
```
Al2Klimov, line 28: Admittedly this is an established pattern. To create new groups inside functions. But I like re-using existing ones -as in ErrgroupReceive- better.

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,80 @@
...
+				return ctx.Err()
+			}
+		}
+	})
+
+	return entities, com.WaitAsync(g)
+}
+
+// GenerateBinaryId generates a 20 byte length random id
+func generateBinaryId() (types.Binary, error) {
```
Al2Klimov, line 75: Small function and used only once. IMAO not necessary as such.

Al2Klimov (changes requested):

schema/mysql/schema.sql
yhabteab: Overseen the 🐘

pkg/icingadb/sla.go
Al2Klimov: @julianbrost Your second opinion, please.

schema/mysql/schema.sql
```diff
@@ -1321,6 +1321,18 @@ CREATE TABLE sla_history_downtime (
   INDEX idx_sla_history_downtime_env_downtime_end (environment_id, downtime_end) COMMENT 'Filter for sla history retention'
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
 
+CREATE TABLE sla_history_trail (
+  id int unsigned NOT NULL AUTO_INCREMENT,
```
Al2Klimov, line 1325: IMAO not one of the cases where 32 bits are enough.

Al2Klimov, line 1325: Either make this one signed...

schema/pgsql/schema.sql
```diff
@@ -2147,6 +2148,23 @@ COMMENT ON COLUMN sla_history_downtime.downtime_id IS 'downtime.id (may referenc
 COMMENT ON COLUMN sla_history_downtime.downtime_start IS 'start time of the downtime';
 COMMENT ON COLUMN sla_history_downtime.downtime_end IS 'end time of the downtime';
 
+CREATE TABLE sla_history_trail (
+  id serial NOT NULL,
```
Al2Klimov, line 2152: ... or use CREATE DOMAIN here as I did.

schema/mysql/schema.sql
```diff
@@ -1321,6 +1321,18 @@ CREATE TABLE sla_history_downtime (
...
+  environment_id binary(20) NOT NULL COMMENT 'environment.id',
+  host_id binary(20) NOT NULL COMMENT 'host.id (may reference already deleted hosts)',
+  service_id binary(20) DEFAULT NULL COMMENT 'service.id (may reference already deleted services)',
+
+  PRIMARY KEY (id),
```
Al2Klimov, line 1330: This line comes too early.

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,72 @@
...
+	"time"
+)
+
+type SlaHistoryTrail struct {
+	v1.EntityWithoutChecksum `json:",inline" db:"-"`
+	EnvironmentId            types.Binary    `json:"environment_id"`
+	HostId                   types.Binary    `json:"host_id"`
+	ServiceId                types.Binary    `json:"service_id"`
+	EventTime                types.UnixMilli `json:"event_time"`
+	EventType                string          `json:"event_type"`
```
Al2Klimov, line 18: Should have the same order as in SQL.

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+	"github.com/icinga/icingadb/pkg/contracts"
+	v1 "github.com/icinga/icingadb/pkg/icingadb/v1"
+	"github.com/icinga/icingadb/pkg/types"
+	"golang.org/x/sync/errgroup"
+	"time"
+)
+
+type SlaHistoryTrail struct {
+	v1.EntityWithoutChecksum `json:",inline" db:"-"`
```
Al2Klimov, line 13: Doesn't match the datatype in SQL.

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+type SlaHostHistoryTrailColumns struct {
+	v1.EntityWithoutChecksum `json:",inline"`
+	EnvironmentId            `json:",inline"`
+}
+
+type SlaServiceHistoryTrailColumns struct {
+	v1.EntityWithoutChecksum `json:",inline"`
+	EnvironmentId            `json:",inline"`
```
Al2Klimov, line 28: Don't you already have the env somewhere?

julianbrost:
I just had an idea how we could call that type of SLA history after we didn't really come up with good name for this initially: lifecycle

pkg/types/int.go
```diff
@@ -58,11 +60,16 @@ func (i *Int) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func (i Int) String() string {
+	return fmt.Sprintf("%d", i.Int64)
+}
+
```
Al2Klimov, line 66: Something is missing here due to i is not just an int. Also while you're on it consider FormatUint.

Al2Klimov: And I don't mean the (also) missing doc.

pkg/icingadb/db.go
```diff
@@ -515,7 +517,14 @@ func (db *DB) YieldAll(ctx context.Context, factoryFunc contracts.EntityFactoryF
...
 
-		rows, err := db.NamedQueryContext(ctx, query, scope)
+		var rows *sqlx.Rows
+		var err error
+		if namedQueryParams {
+			rows, err = db.NamedQueryContext(ctx, query, scope)
+		} else {
+			rows, err = db.QueryxContext(ctx, query, scope...)
+		}
+
```
Al2Klimov, line 527: If you reflect whether scope is a slice, could you decide based on that...

pkg/icingadb/db.go
```diff
@@ -506,7 +506,9 @@ func (db *DB) BatchSizeByPlaceholders(n int) int {
 // YieldAll executes the query with the supplied scope,
 // scans each resulting row into an entity returned by the factory function,
 // and streams them into a returned channel.
-func (db *DB) YieldAll(ctx context.Context, factoryFunc contracts.EntityFactoryFunc, query string, scope interface{}) (<-chan contracts.Entity, <-chan error) {
+func (db *DB) YieldAll(
+	ctx context.Context, factoryFunc contracts.EntityFactoryFunc, query string, namedQueryParams bool, scope ...interface{},
+) (<-chan contracts.Entity, <-chan error) {
```
Al2Klimov, line 511: ... and avoid the extra parameter?

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/icinga/icingadb/pkg/contracts"
+	v1 "github.com/icinga/icingadb/pkg/icingadb/v1"
+	"github.com/icinga/icingadb/pkg/types"
+	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
+	"time"
+)
+
+type SlaHistoryTrail struct {
+	Id                 types.Int `json:"id" db:"-"`
```
Al2Klimov, line 14: `db:"-"` is correct. But its reason isn’t obvious for a not involved programmer.

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,137 @@
...
+	v1.EnvironmentMeta `json:",inline"`
+	HostId             types.Binary    `json:"host_id"`
+	ServiceId          types.Binary    `json:"service_id"`
+	EventType          string          `json:"event_type"`
+	EventTime          types.UnixMilli `json:"event_time"`
+}
+
+// Fingerprint implements the contracts.Fingerprinter interface.
+func (sht SlaHistoryTrail) Fingerprint() contracts.Fingerprinter {
+	return sht
```
Al2Klimov, line 24: Why **exactly** do you do it like this?

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,137 @@
...
+func (sht SlaHistoryTrail) ID() contracts.ID {
+	return sht.Id
+}
+
+// SetID implements part of the contracts.IDer interface.
+func (sht *SlaHistoryTrail) SetID(id contracts.ID) {
+	sht.Id = id.(types.Int)
+}
+
+type SlaServiceHistoryTrailColumns struct {
```
Al2Klimov, line 37: Isn’t this basically HostIdByService?

Al2Klimov (changes requested):
Please don’t **force**-push for now.

pkg/icingadb/db.go
yhabteab: NO! Since it's fine to pass slice of named structs as well.

pkg/icingadb/sla.go
Al2Klimov, line 14: Why is the column here at all?

pkg/types/int.go
Al2Klimov, line 66: And why you've added this?

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,137 @@
...
+
+// ID implements part of the contracts.IDer interface.
+func (sht SlaHistoryTrail) ID() contracts.ID {
+	return sht.Id
+}
+
+// SetID implements part of the contracts.IDer interface.
+func (sht *SlaHistoryTrail) SetID(id contracts.ID) {
+	sht.Id = id.(types.Int)
+}
```
Al2Klimov, line 35: Just to implement these two methods? Then what about making them no-ops instead? Just an idea...

pkg/icingadb/db.go
Al2Klimov: As scope?

pkg/icingadb/sync.go
```diff
@@ -160,9 +182,52 @@ func (s Sync) ApplyDelta(ctx context.Context, delta *Delta) error {
 	// Delete
 	if len(delta.Delete) > 0 {
 		s.logger.Infof("Deleting %d items of type %s", len(delta.Delete), utils.Key(utils.Name(delta.Subject.Entity()), ' '))
+		entity := delta.Subject.Entity()
+		if _, ok := entity.(*v1.Service); ok {
```
Al2Klimov, line 186: Can be...

pkg/icingadb/sync.go
```diff
@@ -160,9 +182,52 @@ func (s Sync) ApplyDelta(ctx context.Context, delta *Delta) error {
...
+				com.ErrgroupReceive(g, err)
+
+				return s.db.CreateStreamed(ctx, CheckableToSlaTrailEntities(ctx, g, entities, "delete"))
+			})
+		}
+
+		var hostIds chan any
+		onSuccessHandlers := []OnSuccess[any]{OnSuccessIncrement[any](stat)}
+
+		_, isHost := entity.(*v1.Host)
```
Al2Klimov, line 211: ... a switch/case IMAO.

pkg/types/int.go
yhabteab: > Something is missing here due to i is not just an int

Of course it's just an `Int`. I don't know what this comment is supposed to tell me.

pkg/icingadb/sync.go
```diff
@@ -160,9 +182,52 @@ func (s Sync) ApplyDelta(ctx context.Context, delta *Delta) error {
...
+					query += ` WHERE id = ?`
+				} else {
+					var placeholders []string
+					for i := 0; i < len(delta.Delete); i++ {
+						placeholders = append(placeholders, "?")
+					}
+
+					query += fmt.Sprintf(` WHERE id IN (%s)`, strings.Join(placeholders, `, `))
+				}
+				entities, err := s.db.YieldAll(ctx, delta.Subject.Factory(), query, false, delta.Delete.IDs()...)
```
Al2Klimov, line 201: Wondering if you could just pass a slice of IDs for one ? placeholder...

pkg/icingadb/db.go
yhabteab: Yes!

Al2Klimov: Where do we do this?

yhabteab: Didn't say we were doing that somewhere right now. Just saying that it's possible!

pkg/icingadb/sla.go
yhabteab: 😕?

Al2Klimov: Please see the review https://github.com/Icinga/icingadb/pull/566#pullrequestreview-1323482708 as a whole. The comments belong together.

pkg/icingadb/db.go
Al2Klimov: Do we do this in Icinga DB?

yhabteab: 👇
> Didn't say we were doing that somewhere right now. Just saying that it's possible!

Al2Klimov: Do we do this in Icinga DB? Yes or no? And where?

pkg/icingadb/sync.go
yhabteab: No! You can't!

pkg/icingadb/sla.go
yhabteab: You're questioning why I have this column in this struct in the first place. I don't see any coherence with other comments.

pkg/icingadb/sla.go
Al2Klimov: If these methods would be no-ops, you could drop the methods https://github.com/Icinga/icingadb/pull/566#discussion_r1124277436 altogether along with the column https://github.com/Icinga/icingadb/pull/566#discussion_r1124277147 IMAO.

yhabteab: Shouldn't the struct reflect the database table? You even told me that yesterday.

Al2Klimov: Yes. But in this case I see no sense. See, we even explicitly set db:"-". But on the other hand you’re right. As I said: Just an idea.

pkg/icingadb/sync.go
yhabteab: It won't make any difference as the host IDs goroutine is dependent on the `onSuccess` handler of the Delete goroutine.

yhabteab: NM!

pkg/types/int.go
```diff
@@ -58,11 +59,21 @@ func (i *Int) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// String returns the string representation of i.Int64 in 10 base
```
Al2Klimov, line 62: IMAO this doc could be more like the others in the file.

pkg/icingadb/v1/entity.go
```diff
@@ -26,3 +29,13 @@ func (e EntityWithChecksum) Fingerprint() contracts.Fingerprinter {
 func NewEntityWithChecksum() contracts.Entity {
 	return &EntityWithChecksum{}
 }
+
+// EntityWitNoneBinaryId represents entities with auto incremented integer id
+type EntityWitNoneBinaryId struct {
+	Id types.Int `json:"id" db:"-"` // Ignore db tag as it's auto incremented
```
Al2Klimov, line 35: 💡 Instead of the comment...

pkg/icingadb/v1/entity.go
```diff
@@ -26,3 +29,13 @@ func (e EntityWithChecksum) Fingerprint() contracts.Fingerprinter {
 func NewEntityWithChecksum() contracts.Entity {
 	return &EntityWithChecksum{}
 }
+
+// EntityWitNoneBinaryId represents entities with auto incremented integer id
+type EntityWitNoneBinaryId struct {
```
Al2Klimov, line 34: ... you could reflect it in the name making it btw. more speaking.

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,132 @@
...
```
Al2Klimov, line 30: Apropos the mentioned struct w/ autoinc. IDs: these methods could be part of it.

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Get environment id from the given context
+func getEnvironmentId(ctx context.Context) (types.Binary, error) {
+	env, ok := v1.EnvironmentFromContext(ctx)
+	if !ok {
+		return nil, errors.New("can't get environment from context")
+	}
+
+	return env.Id, nil
+}
```
julianbrost, line 115: That function looks like it should already exist somewhere else.

julianbrost:
Did you test that for runtime updates, the normal objects tables are still updated properly? The integration tests suggest that there might be something wrong with that given that most `TestObjectSync/RuntimeUpdates/*` tests fail.

yhabteab:
Since the normal runtime updates are waiting now for the sla-related goroutines to finish, it won't sync all updates that fast. I think the tests just need a higher sleep time.

julianbrost:
They wait for up to 20 seconds. Shouldn't that be more than enough?

julianbrost:
The integration tests manage to crash the Icinga DB process:

```
2023-03-14T16:38:44.331Z        FATAL   icingadb        Error 1048: Column 'host_id' cannot be null
can't perform "INSERT INTO \"sla_history_trail\" (\"host_id\", \"service_id\", \"create_time\", \"delete_time\", \"environment_id\", \"id\") VALUES (:host_id, :service_id, :create_time, :delete_time, :environment_id, :id) ON DUPLICATE KEY UPDATE \"id\" = \"id\""
        /home/runner/work/icingadb/icingadb/internal/internal.go:30
        /home/runner/work/icingadb/icingadb/pkg/icingadb/db.go:401
        /home/runner/work/icingadb/icingadb/pkg/retry/retry.go:45
        /home/runner/work/icingadb/icingadb/pkg/icingadb/db.go:396
        /home/runner/go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57
runtime.goexit
        /opt/hostedtoolcache/go/1.20.1/x64/src/runtime/asm_amd64.s:1598
can't retry
        /home/runner/work/icingadb/icingadb/pkg/retry/retry.go:64
        /home/runner/work/icingadb/icingadb/pkg/icingadb/db.go:396
        /home/runner/go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57
runtime.goexit
        /opt/hostedtoolcache/go/1.20.1/x64/src/runtime/asm_amd64.s:1598
```

yhabteab:
```go
--- PASS: TestObjectSync (17.77s)
    --- PASS: TestObjectSync/RuntimeUpdates (20.00s)
        --- PASS: TestObjectSync/RuntimeUpdates/User (0.01s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-Base (2.13s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-Types-4 (2.14s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-States-5 (2.14s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-States-4 (2.14s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-States-3 (2.16s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-States-2 (2.14s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-States-1 (2.16s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-Types-9 (2.13s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-Types-8 (2.15s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-Types-7 (2.14s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-Types-6 (2.12s)
            --- PASS: TestObjectSync/RuntimeUpdates/User/CreateAndDelete-Types-5 (2.13s)
                ...
                --- PASS: TestObjectSync/RuntimeUpdates/Service/Update/DisplayName-2 (14.20s)
                --- PASS: TestObjectSync/RuntimeUpdates/Service/Update/CheckCommandName-2 (14.18s)
                --- PASS: TestObjectSync/RuntimeUpdates/Service/Update/CheckCommandName-1 (13.19s)
                --- PASS: TestObjectSync/RuntimeUpdates/Service/Update/HostName-2 (13.17s)
                --- PASS: TestObjectSync/RuntimeUpdates/Service/Update/DisplayName-1 (14.19s)
                --- PASS: TestObjectSync/RuntimeUpdates/Service/Update/HostName-1 (14.18s)
PASS
```

pkg/icingadb/runtime_updates.go
```diff
@@ -95,57 +96,198 @@ func (r *RuntimeUpdates) Sync(
...
-			}).Stop()
 
-			sem := r.db.GetSemaphoreForTable(utils.TableName(s.Entity()))
+				stmt, _ := r.db.BuildInsertIgnoreStmt(first)
 
-			onSuccess := []OnSuccess[any]{OnSuccessIncrement[any](&counter), OnSuccessIncrement[any](stat)}
-			if !allowParallel {
-				onSuccess = append(onSuccess, OnSuccessSendTo(deletedFifo))
-			}
+				// We can only bulk exec one query at a time, otherwise, the regression test won't pass
```
julianbrost, line 168: > the regression test won't pass

Bonus points for telling why apart form the tests. I mean, in theory, there could also be a problem in the tests.

pkg/icingadb/sync.go
```diff
@@ -160,9 +180,61 @@ func (s Sync) ApplyDelta(ctx context.Context, delta *Delta) error {
...
+				if first == nil {
+					return errors.Wrap(err, "can't copy first entity")
+				}
+				sem := s.db.GetSemaphoreForTable(utils.TableName(first))
+				stmt, _ := s.db.BuildUpdateStmt(first, []string{"delete_time"}, false)
+				if _, ok := entity.(*v1.Host); ok {
+					stmt += ` WHERE "id" = :id AND "service_id" IS NULL AND "delete_time" = 0`
+				} else {
+					stmt += ` WHERE "host_id" = :host_id AND "service_id" = :service_id AND "delete_time" = 0`
+				}
```
julianbrost, line 201: Why is only one using `id`? The second one should probably also use `id` as you don't have an index on the other columns.

pkg/icingadb/sync.go
```diff
@@ -160,9 +180,61 @@ func (s Sync) ApplyDelta(ctx context.Context, delta *Delta) error {
...
+			})
+
+			s.logger.Infof("Inserting %d %s sla lifecycles of type delete", len(delta.Delete), delta.Subject.Name())
+
+			entities := make(chan contracts.Entity, len(delta.Delete))
+			g.Go(func() error {
+				defer close(entities)
+
+				return s.db.CreateIgnoreStreamed(ctx, updatedCheckables, OnSuccessApplyAndSendTo(entities, GetCheckableFromSlaLifecycle))
+			})
```
julianbrost, line 213: There should be an explanation what this is doing here.

pkg/icingadb/sync.go
```diff
@@ -160,9 +180,61 @@ func (s Sync) ApplyDelta(ctx context.Context, delta *Delta) error {
...
+
+					select {
+					case deleteIds <- entity.ID():
+					case <-ctx.Done():
+						return ctx.Err()
+					}
+				}
+
+				return s.db.DeleteStreamed(ctx, delta.Subject.Entity(), deleteIds, OnSuccessIncrement[any](stat))
+			})
```
julianbrost, line 232: That looks like it should be two goroutines. Apart from that, can't what the for loop does also be done by `OnSuccessApplyAndSendTo`?

julianbrost (changes requested):
Please also write a PR description that describes the change itself and the motivation for doing that change.

schema/mysql/schema.sql
```diff
@@ -1321,6 +1321,18 @@ CREATE TABLE sla_history_downtime (
...
+CREATE TABLE sla_lifecycle (
+  id binary(20) NOT NULL COMMENT 'sha1(environment.id, host.id, service.id)',
+  environment_id binary(20) NOT NULL COMMENT 'environment.id',
+  host_id binary(20) NOT NULL COMMENT 'host.id (may reference already deleted hosts)',
+  service_id binary(20) DEFAULT NULL COMMENT 'service.id (may reference already deleted services)',
+
+  create_time bigint unsigned DEFAULT NULL COMMENT 'unix timestamp the event occurred',
+  delete_time bigint unsigned DEFAULT NULL COMMENT 'unix timestamp the delete event occurred',
+
+  CONSTRAINT `pk_sla_lifecycle` UNIQUE KEY (id, delete_time)
```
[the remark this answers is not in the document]
yhabteab: The Github actions says I can't use nullable column as part of the primary key.

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,144 @@
...
+	v1.EntityWithoutChecksum `json:",inline"`
+	v1.EnvironmentMeta       `json:",inline"`
+	HostId                   types.Binary    `json:"host_id"`
+	ServiceId                types.Binary    `json:"service_id"`
+	CreateTime               types.UnixMilli `json:"create_time"`
+	DeleteTime               types.UnixMilli `json:"delete_time"`
+
+	// The original checkable entity from which this sla lifecycle were transformed
+	OrgEntity contracts.Entity `json:"-" db:"-"`
+}
```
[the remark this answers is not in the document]
yhabteab: I was also thinking the same at the beginning, but then icingadb wouldn't build, since there will be some dependence cycles.

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,144 @@
...
+
+		env, ok := v1.EnvironmentFromContext(ctx)
+		if !ok {
+			return errors.New(environmentError)
+		}
+
+		for {
+			// There is no other way to use the same date time for runtime updates, as this
+			// goroutine won't complete until Icinga DB stops or crashes once it has been started.
+			// Otherwise, "now" will not be updated once it is initialized.
```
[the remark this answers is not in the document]
yhabteab: Well, if you delete 10 hosts at the same time they should theoretically have the same event time, as it was e.g. with initial config sync.

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,144 @@
...
+					sl.Id = utils.Checksum(objectpacker.MustPackSlice(sl.EnvironmentId, sl.HostId))
+				case *v1.EntityWithChecksum:
+					sl.HostId = checkable.Id
+					sl.Id = utils.Checksum(objectpacker.MustPackSlice(sl.EnvironmentId, sl.HostId))
+				}
+
+				if isDeleteEvent {
+					sl.DeleteTime = types.UnixMilli(now)
+					sl.CreateTime = types.UnixMilli(time.Unix(0, 0))
+				}
```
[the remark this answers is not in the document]
yhabteab: No. Delete event updates/inserts only the `delete_time` column.

schema/mysql/schema.sql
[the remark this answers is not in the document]
julianbrost: Are you even inserting NULL?

pkg/icingadb/sla.go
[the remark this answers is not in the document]
julianbrost: What did you put there? Just the struct definition or this whole file?

pkg/icingadb/sla.go
```diff
@@ -0,0 +1,144 @@
...
+			select {
+			case entity, ok := <-entities:
+				if !ok {
+					return nil
+				}
+
+				sl := &SlaLifecycle{
+					EnvironmentMeta: v1.EnvironmentMeta{EnvironmentId: env.Id},
+					CreateTime:      types.UnixMilli(now),
+					DeleteTime:      types.UnixMilli(time.Unix(0, 0)),
```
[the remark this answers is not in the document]
yhabteab: I don't think so.

pkg/icingadb/sync.go
yhabteab: Overseen the 🐘

pkg/icingadb/runtime_updates.go
```diff
@@ -346,8 +488,19 @@ func structifyStream(
...
+					if service, ok := entity.(*v1.Service); ok {
+						if service.HostId == nil {
+							err := scanHostIdForService(ctx, service, db)
+							if err != nil {
+								return err
+							}
+
+							entity = service
+						}
+					}
```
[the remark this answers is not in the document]
yhabteab: Yes, it could be, but I just want to avoid an extra goroutine that receives the service, selects its host ID and forwards it to another one. Here on the other hand you don't need an extra step.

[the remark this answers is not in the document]
yhabteab: And I would need to add a differentiation for service and host types. But this way I don't need special treatment for services.

pkg/icingadb/sla.go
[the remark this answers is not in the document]
yhabteab: The whole file.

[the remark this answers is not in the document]
julianbrost: Ok, so this should probably be split as well. Most structs that represent database tables are somewhere inside `pkg/icingadb/v1/` whereas the code working with them and writing them to the database is somewhere else.

pkg/icingadb/db.go
```diff
@@ -287,6 +294,21 @@ func OnSuccessSendTo[T any](ch chan<- T) OnSuccess[T] {
 	}
 }
 
+// OnSuccessApplyAndSendTo applies the provided callback to all the rows and streams them to the passed chan
+func OnSuccessApplyAndSendTo[T any](ch chan<- T, fun func(T) T) OnSuccess[T] {
```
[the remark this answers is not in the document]
yhabteab: I wasn't aware of this feature! It's definitely 💯 better to have a second type there. Ty

pkg/icingadb/db.go
```diff
@@ -166,7 +166,7 @@ func (db *DB) BuildInsertIgnoreStmt(into interface{}) (string, int) {
 	switch db.DriverName() {
 	case driver.MySQL:
 		// MySQL treats UPDATE id = id as a no-op.
-		clause = fmt.Sprintf(`ON DUPLICATE KEY UPDATE "%s" = "%s"`, columns[0], columns[0])
+		clause = `ON DUPLICATE KEY UPDATE "id" = "id"`
```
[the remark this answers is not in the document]
yhabteab: What should I do then instead?

schema/mysql/schema.sql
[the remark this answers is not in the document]
yhabteab: Unless you set this mode `SET SESSION sql_mode='NO_AUTO_VALUE_ON_ZERO';`

pkg/icingadb/db.go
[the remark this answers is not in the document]
julianbrost: From your commit message:

> The order of the columns varies each time when calling `db.BuildColumns()`, and it isn't guaranteed to be `id` the first one.

That is probably already the case without any of the changes in this PR, isn't it? Does anything break in this PR if you keep this code as is? If not, I'd keep this code and open an issue for it.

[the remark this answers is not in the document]
yhabteab: Well, as Alex already commented there, mysql considers`ON DUPLICATE KEY UPDATE id = id` as no-op and when the used column is not an `id` then this could lead to unwanted updates.

schema/mysql/schema.sql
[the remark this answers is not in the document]
julianbrost: [`NO_AUTO_VALUE_ON_ZERO`](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_no_auto_value_on_zero) should only make a difference if there are any auto-increment columns, but there are none in this table.

I also did a quick test and found no apparent issues with a `primary key` on two `not null` columns where one stores 0 sometimes (on MariaDB 10.11, so this isn't to say that it works everywhere).

julianbrost (changes requested):

pkg/icingadb/runtime_updates.go
```diff
@@ -95,57 +95,176 @@ func (r *RuntimeUpdates) Sync(
...
+					com.NeverSplit[contracts.Entity], OnSuccessSendTo(updatedEntities), OnSuccessIncrement[contracts.Entity](&counter),
+				)
+			})
+
+			deletedIds := make(chan interface{}, r.redis.Options.XReadCount)
+			g.Go(func() error {
+				// extractEntityId is used as a callback for the on success mechanism to extract the checkable ids.
+				extractEntityId := func(e contracts.Entity) interface{} {
+					sl, _ := e.(*v1.SlaLifecycle)
+					return sl.SourceEntity.ID()
```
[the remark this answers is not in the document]
yhabteab: > This would panic if `e` isn't a `*v1.SlaLifecycle` with a nil dereference.

How should this ever happen? The `_` clearly indicates not to expect any errors, which implies that `e` being always `*v1.SlaLifecycle`. Though from a stylistic point of view, your suggestion looks way better.

pkg/icingadb/sla_lifecycle.go
```diff
@@ -0,0 +1,95 @@
...
+					if !sl.HostId.Valid() {
+						err := db.QueryRowxContext(
+							ctx, db.Rebind(fmt.Sprintf("%s %s", db.BuildSelectStmt(checkable, v1.HostMeta{}), ` WHERE "id" = ?`)),
+							sl.ServiceId,
+						).StructScan(sl)
+
+						if err != nil {
+							return err
+						}
+					}
```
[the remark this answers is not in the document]
yhabteab: For delete event of runtime updates.

pkg/icingadb/runtime_updates.go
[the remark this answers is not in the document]
julianbrost: Yes, this should never happen (if I'd expect it to happen, I'd have mentioned that instead). But in this situation, there is really no point in using the non-panicking type assertion when the following would crash nonetheless if the assertion failed.

And if there was a bug, the panicking version would result in a more useful error, compare these two for exampe:
- https://go.dev/play/p/QQmKf-t5iPG
- https://go.dev/play/p/yB7My67Nlmz

One clearly says that there was something totally wrong in the interface variable while in the other case, it's unclear from the error alone whether this was due to the type assertion failing or if an actual nil pointer value somehow made it into the interface variable.

pkg/icingadb/sla_lifecycle.go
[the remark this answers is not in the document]
julianbrost: I think it might make sense doing this in the runtime update specific code if it's only for one very specific situation.

This should (hopefully, I haven't checked if anything else might prevent this) also allow to use the existing chunks/bulks that are generated there and query the information for all entities in there in a single query.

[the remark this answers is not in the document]
yhabteab: > I think it might make sense doing this in the runtime update specific code if it's only for one very specific situation.

https://github.com/Icinga/icingadb/pull/566#discussion_r1140133758

> This should (hopefully, I haven't checked if anything else might prevent this) also allow to use the existing chunks/bulks that are generated there and query the information for all entities in there in a single query.

How is this supposed to work? The chunks are generated within the respective `DB#*BulkExec()` methods, which you don't have access to. Btw, this is just relevant for delete events of the runtime updates, which are IMAO not that expensive as might you think.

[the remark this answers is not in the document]
julianbrost: Ok, probably ignore that comment for now. Looks like I should look into this part in more detail first.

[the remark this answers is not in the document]
julianbrost: > > I think it might make sense doing this in the runtime update specific code if it's only for one very specific situation.
> 
> [#566 (comment)](https://github.com/Icinga/icingadb/pull/566#discussion_r1140133758)

That select is just untypical for Icinga DB, so it will probably be a bit weird everywhere. Probably indeed makes more sense here, but there should be a clear comment what this is for.

Also, is there an option for Icinga 2 to send this information directly? I don't mean as an replacement for this bit of code but as an optimization that would make this just a fallback?
 
> > This should (hopefully, I haven't checked if anything else might prevent this) also allow to use the existing chunks/bulks that are generated there and query the information for all entities in there in a single query.
> 
> How is this supposed to work? The chunks are generated within the respective `DB#*BulkExec()` methods, which you don't have access to.

My idea was that the `OnSuccess` functions all receive a slice and the `host_id` could have been queried for all in there as well.

> Btw, this is just relevant for delete events of the runtime updates, which are IMAO not that expensive as might you think.

This only affects runtime deletion of services, not something that's super common to do very often (things would be different here for other types like downtimes).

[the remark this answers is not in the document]
yhabteab: > My idea was that the `OnSuccess` functions all receive a slice and the `host_id` could have been queried for all in there as well.

The `OnSuccess` functions do all get a slice, but since the `SlaLifecycle` entries are processed first, this won't be useful in any way, as the `host_id` has to be set beforehand.

> Also, is there an option for Icinga 2 to send this information directly?

There is of course an option to set this information directly by Icinga 2, but none of you guys wanted a change on Icinga 2 side in this regard.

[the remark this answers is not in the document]
julianbrost: > but none of you guys wanted a change on Icinga 2 side in this regard.

Not as an requirement. But allowing for an optimization would be good. So my question is: with the current state of the PR, if there was a `host_id` in the Redis runtime update stream, would this be used and skip the query? Just so that we have the option to do this as an optimization later, without requiring it immediately.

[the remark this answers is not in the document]
yhabteab: > if there was a `host_id` in the Redis runtime update stream, would this be used and skip the query?

Yes, this would work if the Redis runtime **delete** stream contained `host_id`.

julianbrost (changes requested):

pkg/icingadb/runtime_updates.go
```diff
@@ -95,57 +95,184 @@ func (r *RuntimeUpdates) Sync(
...
+						r.logger.Infof("Upserted %d %s sla lifecycles", count, s.Name())
+					}
+				}).Stop()
+
+				sl := v1.NewSlaLifecycle()
+				stmt, _ := r.db.BuildInsertIgnoreStmt(sl)
+
+				// Not to mess up the already existing FIFO mechanism, we have to perform only a single query
+				// (semaphore 1) at a time, even the sla queries could be bulk executed.
+				// After successfully upserting a lifecycle entity, the original checkable entity is streamed to "entities".
```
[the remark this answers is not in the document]
yhabteab: > For delete, this is running `DELETE ... WHERE id IN (?)` with more than one ID at the same time.

Not when `deleteCount` is set to `1`, which is the case when it's not allowed to do this in parallel (`allowParallel`).

pkg/icingadb/sla.go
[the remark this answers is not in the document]
julianbrost: It should probably be in `pkg/icingadb/v1/history/` instead of `pkg/icingadb/v1/`. At least the other types like `SlaHistoryState` and `SlaHistoryDowntime` are defined in `history/` as well. Or would this again result in some import cycles?

julianbrost:
Now that I'm taking a fresh look at this after some time, I'm wondering: what was the reason for updating rows instead of inserting separate rows for create and delete events?

julianbrost:
I've updated the PR description with a summary of what queries are performed when (section "Implementation").

> Now that I'm taking a fresh look at this after some time, I'm wondering: what was the reason for updating rows instead of inserting separate rows for create and delete events?

This also gave sort of an answer to this question: having that database structure with the delete time being part of the primary key prevents duplicate rows to be inserted for runtime updates.

Al2Klimov:
Admittedly not the most beautiful concept, but it solves problems we'd have with 1 event = 1 row 👍

julianbrost:
While looking at this after over a year (🙈), I had an idea: what if during the schema upgrade, we just insert rows for all currently existing hosts and services into the new table, shouldn't that allow us to get rid of two special cases (emphasis by me)?

> 2\. Additionally, it performs an `INSERT` with ignore for duplicate keys with the same timestamp. So in case there was no row to be updated, it will now be inserted (otherwise, this query is a no-op). **This is especially important for the case where objects were created before this feature becomes available.**
> 
> ### Runtime updates
> #### Upsert
> Performs an `INSERT` with ignore for duplicate keys for both create and update events (these look identical in the runtime update stream). If the object is already marked as alive in `sla_history_lifecycle`, this will do nothing, otherwise it will mark it as created now (**including when an object that was created before this feature was enabled is updated**).

yhabteab:
> I had an idea: what if during the schema upgrade, we just insert rows for all currently existing hosts and services into the new table, shouldn't that allow us to get rid of two special cases (emphasis by me)?

How do you plan to perform this hashing with pure SQL?
https://github.com/Icinga/icingadb/blob/5a2396e49cfb05b35b4c9cde8cd6d86d373e94d0/pkg/icingadb/sla_lifecycle.go#L84

julianbrost:
> How do you plan to perform this hashing with pure SQL?
> 
> https://github.com/Icinga/icingadb/blob/5a2396e49cfb05b35b4c9cde8cd6d86d373e94d0/pkg/icingadb/sla_lifecycle.go#L84

That's introduced by this PR, so it could also be done differently if this makes things easier. Which brings me right to my next ideas, which I have plenty of. :sweat_smile: Please read them, ask any question that may come up and let me know if you think that the approach I'm suggesting in the following may indeed significantly simplify this PR.

Doesn't much of the new code for selecting the host ID for a service primarily exist just because of the choice to compute the `id` column based on it? The service ID itself already includes the host ID, so including it is not actually necessary for uniqueness of the resulting IDs, i.e. the same service on different hosts already has different IDs. So if that was just omitted from the hash, shouldn't good parts of the new code actually become unnecessary?

Going one more step further: host and service IDs already include the environment ID as well, so even simply using `sla_lifecycle.id = host.id` or `sla_lifecycle.id = service.id` should be an option[^1]. That would then allow a pretty simple query to be used in the schema migration (for hosts, would work for services analogously):

```sql
insert into sla_lifecycle (id, environment_id, host_id, create_time, delete_time)
select id, environment_id, id, UNIX_TIMESTAMP() * 1000, 0
from host;
```

That wasn't the last step/idea of mine, this could be taken even further: one could add a where clause so that instead of all hosts, only those that don't yet have a row in `sla_lifecycle` are inserted:

```sql
insert into sla_lifecycle (id, environment_id, host_id, create_time, delete_time)
select id, environment_id, id, UNIX_TIMESTAMP() * 1000, 0
from host
where id not in (select id from sla_lifecycle where service_id is null and delete_time = 0);
```

So this gives a simple query that could just be executed each time after the initial delta sync. That would take care of created objects, deleted objects could be handled similarly by updating all `sla_lifecycle` rows with `delete_time = 0` that reference dead IDs:

```sql
update sla_lifecycle
set delete_time = UNIX_TIMESTAMP() * 1000
where service_id is null
and delete_time = 0
and id not in (select id from host);
```

So with 4 queries in total[^2], this should take care of all the updates needed during the initial delta sync, leaving only the runtime updates, those should be possible to be reduced to a single insert/update each, as potentially missing rows were already created by the previous delta sync.

[^1]: If I remember correctly, there can't be conflicts between host IDs and service IDs, which would be a requirement for doing that way. That's something that has to be verified, but even if it's not the case, this could trivially be fixed by including an `object_type` column into the primary key (such a column already exists in the other SLA tables).
[^2]: Note that the queries I gave are still missing checks on the `environment_id` in the `where` clauses.

yhabteab:
I have reworked this as suggested by Julian in https://github.com/Icinga/icingadb/pull/566#issuecomment-2273088195 and also updated the PR description accordingly.

yhabteab:
> I have reworked this as suggested by Julian in [#566 (comment)](https://github.com/Icinga/icingadb/pull/566#issuecomment-2273088195) and also updated the PR description accordingly.

The proposals from https://github.com/Icinga/icingadb/pull/566#issuecomment-2273088195 should have been fully implemented by now and of course the PR description is also updated *again*.

pkg/icingadb/sla_lifecycle.go
```diff
@@ -0,0 +1,129 @@
...
+// the `host` and `service` tables and updates the `delete_time` of each of the sla lifecycle entries whose
+// host/service IDs cannot be found in the `host/service` tables.
+//
+// It's unlikely, but when a given Checkable doesn't already have a `create_time` entry in the database, the update
+// query won't update anything. Likewise, the insert statements may also become a no-op if the Checkables already
+// have a `create_time` entry with ´delete_time = 0`.
+//
+// This function retries any database errors for at least `5m` before giving up and failing with an error.
+func SyncCheckablesSlaLifecycle(ctx context.Context, db *database.DB) error {
+	hostUpdateStmt := `UPDATE %[1]s SET delete_time = %d WHERE service_id IS NULL AND delete_time = 0 AND NOT EXISTS(SELECT 1 FROM host WHERE host.id = %[1]s.id)`
```
oxzi, line 88: All these query format strings have numbered and explicit arguments. Maybe consider unifying them and just use numbered arguments?

pkg/icingadb/sla_lifecycle.go
```diff
@@ -0,0 +1,129 @@
...
+	hostInsertStmt := `
+INSERT INTO %[1]s (id, environment_id, host_id, create_time)
+  SELECT id, environment_id, id, %d AS create_time
+  FROM host WHERE NOT EXISTS(SELECT 1 FROM %[1]s WHERE service_id IS NULL AND delete_time = 0 AND host_id = host.id);`
+
+	serviceUpdateStmt := `UPDATE %[1]s SET delete_time = %d WHERE delete_time = 0 AND service_id IS NOT NULL AND NOT EXISTS(SELECT 1 FROM service WHERE service.id = %[1]s.id)`
+	serviceInsertStmt := `
+INSERT INTO %[1]s (id, environment_id, host_id, service_id, create_time)
+  SELECT id, environment_id, host_id, id, %d AS create_time
+  FROM service WHERE NOT EXISTS(SELECT 1 FROM %[1]s WHERE delete_time = 0 AND service_id = service.id);`
```
oxzi, line 98: Below, you first use `{host,service}InsertStmt` followed by the `{host,service}UpdateStmt`. However, the variables are defined in the other order.

pkg/icingadb/sla_lifecycle.go
```diff
@@ -0,0 +1,129 @@
...
+			if _, err := db.ExecContext(ctx, query); err != nil {
+				return database.CantPerformQuery(err, query)
+			}
+
+			query = fmt.Sprintf(serviceInsertStmt, slaLifecycleTable, time.Now().UnixMilli())
+			if _, err := db.ExecContext(ctx, query); err != nil {
+				return database.CantPerformQuery(err, query)
+			}
+
+			query = fmt.Sprintf(serviceUpdateStmt, slaLifecycleTable, time.Now().UnixMilli())
```
oxzi, line 118: Within this block, you have four subsequent `time.Now().UnixMilli()` calls. Due to the query execution time, the value may change in between. Is this desired? Otherwise just store it in a variable.

pkg/icingadb/sla_lifecycle.go
```diff
@@ -0,0 +1,129 @@
...
+			if _, err := db.ExecContext(ctx, query); err != nil {
+				return database.CantPerformQuery(err, query)
+			}
```
oxzi, line 121:
```suggestion
			for _, queryFmt := range []string{hostInsertStmt, hostUpdateStmt, serviceInsertStmt, serviceUpdateStmt} {
				query := fmt.Sprintf(queryFmt, slaLifecycleTable, time.Now().UnixMilli())
				if _, err := db.ExecContext(ctx, query); err != nil {
					return database.CantPerformQuery(err, query)
				}
			}
```

Or am I missing something? You might still consider the `time.Now().UnixMilli()` change I commented above.

oxzi (commented):
First of all, thanks for writing the documentation. It makes it easier for both users and me to understand the motivation for this change.

I just made a first walk through and added some comments.

doc/10-Sla-Reporting.md
```diff
@@ -0,0 +1,171 @@
...
+for event in (sla_history_state, sla_history_downtime) do
+    if (event.previous_hard_state is PENDING) then
+        // A PENDING state event indicates that the host or service in question has not yet had a check result that
+        // clearly identifies its state. Consequently, such events become irrelevant for the purposes of calculating
+        // the SLA and we must exclude the duration of that PENDING state from the total time.
+        total_time = total_time - (event.event_time - last_event_time)
+    else if (previous_hard_state is greater than OK/UP AND previous_hard_state is not PENDING AND checkable is not in DOWNTIME) then
+        // If the previous_hard_state is set to a non-OK state and the host or service in question was not in downtime,
+        // we consider that time slot to be problematic and add the duration to the problem time.
+        problem_time = problem_time + event.event_time - last_event_time
```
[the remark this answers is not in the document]
yhabteab: Does that make any difference? `80+20-10 = 90` / `80+(20-10) = 90`.

doc/images/sla_history_downtime.png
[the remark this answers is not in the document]
yhabteab: They are just a placeholder for the moment!

doc/10-Sla-Reporting.md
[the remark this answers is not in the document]
oxzi: Shouldn't, but you used parentheses in the other case. This just makes it uniform.

https://github.com/Icinga/icingadb/blob/12f577d16792e9ec111effa9f22414bbaf4e6592/doc/10-Sla-Reporting.md?plain=1#L131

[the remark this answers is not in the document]
yhabteab: Because in that case the parentheses matter, as performing `80-20-10` would result in `50`, while performing `80-(20-10)` results in `70`.

schema/mysql/upgrades/1.3.0.sql
```diff
@@ -7,5 +7,87 @@ ALTER TABLE notification MODIFY COLUMN properties_checksum binary(20) NOT NULL C
...
+
+-- Insert an sla lifecycle create_time entry for all existing hosts with the LEAST timestamp found in either
+-- the sla_history_state or sla_history_downtime table, otherwise fallback to the current Unix timestamp.
+INSERT INTO sla_lifecycle (id, environment_id, host_id, create_time)
+  SELECT host.id,
+    host.environment_id,
+    host.id,
+    COALESCE(LEAST(COALESCE(MIN(event_time), MIN(downtime_start)), COALESCE(MIN(downtime_start), MIN(event_time))), UNIX_TIMESTAMP() * 1000) AS create_time
+  FROM host
+    LEFT JOIN sla_history_state shs on host.id = shs.host_id AND service_id IS NULL
```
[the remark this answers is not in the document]
yhabteab: There is only a single `service_id` column, and that is `ssh.service_id`, so I don't necessarily need to fully qualify it's usage.

doc/10-Sla-Reporting.md
```diff
@@ -0,0 +1,171 @@
...
+## Technical Description
+
+!!! info
+
+    This documentation provides a detailed technical explanation of how Icinga DB fulfils all the
+    necessary requirements for the generation of an accurate service level agreement (SLA).
+
+Icinga DB provides a built-in support for automatically storing the relevant events of your hosts and services without
+a manual action. Generally, these events are every **hard** state change a particular checkable encounters and all the
+downtimes scheduled for that checkable throughout its entire lifetime. It is important to note that the aforementioned
```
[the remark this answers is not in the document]
oxzi: I know, but this is not reflected in the docs. Imho, this is something to note, highlight the difference in the final SLA value.

doc/10-Sla-Reporting.md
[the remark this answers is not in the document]
oxzi: As I wanted to write in my previous comment, this was just the suggestion to make it uniform. But it does not really matter :)

schema/pgsql/schema.sql
```diff
@@ -2171,6 +2171,27 @@ COMMENT ON COLUMN sla_history_downtime.downtime_id IS 'downtime.id (may referenc
...
+
+  -- These columns are nullable, but as we're using the delete_time to build the composed primary key, we have to set
+  -- this to `0` instead, since it's not allowed to use a nullable column as part of the primary key.
+  create_time biguint NOT NULL DEFAULT 0,
+  delete_time biguint NOT NULL DEFAULT 0,
+
+  CONSTRAINT pk_sla_lifecycle PRIMARY KEY (id, delete_time)
+);
+
+COMMENT ON COLUMN sla_lifecycle.id IS 'host.id or service.id depending on the checkable type';
```
oxzi, line 2188: > depending on the checkable type

This is a bit implicit. Regardless of having the same `id` field[^1] for both hosts and services, I would make this explicit by stating that a `NULL` value for `serivce_id` implies a host checkable, et vice versa.


[^1]: `2^(8*20) = 2^160` should be unlikely enough for collisions, unless a crafted SHA-1 attack appears.

schema/pgsql/schema.sql
```diff
@@ -2171,6 +2171,27 @@ COMMENT ON COLUMN sla_history_downtime.downtime_id IS 'downtime.id (may referenc
...
+  host_id bytea20 NOT NULL,
+  service_id bytea20 DEFAULT NULL,
+
+  -- These columns are nullable, but as we're using the delete_time to build the composed primary key, we have to set
+  -- this to `0` instead, since it's not allowed to use a nullable column as part of the primary key.
+  create_time biguint NOT NULL DEFAULT 0,
+  delete_time biguint NOT NULL DEFAULT 0,
+
+  CONSTRAINT pk_sla_lifecycle PRIMARY KEY (id, delete_time)
+);
```
oxzi, line 2186: I was thinking about creating some indices for the queries within the `SyncCheckablesSlaLifecycle` function.

In a huge environment, those queries may take some time. However, I cannot say how huge they must become to make those queries slow or if Icinga 2 will be the performance bottleneck before.

I have not tested this or anything, just wanted to put the idea out there.

oxzi (commented):
I have played with this PR in a testing environment and have not encountered any errors so far.

However, I have a more general question: How will the `sla_lifecycle` table be used later on? The `get_sla_ok_percent` function does not honor it and, unless I am missing something, cannot even do this for deleted checkable objects.

schema/pgsql/schema.sql
julianbrost: This is only a suggestion for the column comment as there are already `host_id` and `service_id` columns as well in the table, isn't it?

Regarding collisions: well that's nothing special for this table, it would be the same for collisions between two different hosts for example.

schema/pgsql/schema.sql
julianbrost: > or if Icinga 2 will be the performance bottleneck before

If there's an issue, you could probably trigger it without Icinga 2 becoming an issue. You can create and delete many hosts and services over time, keeping the active config in Icinga 2 at a constant size but growing this table.

schema/pgsql/schema.sql
oxzi: > This is only a suggestion for the column comment as there are already `host_id` and `service_id` columns as well in the table, isn't it?

Exactly.

> Regarding collisions: well that's nothing special for this table, it would be the same for collisions between two different hosts for example.

Compared to the other tables, a potential collision between hosts and services would be possible here. But, as already stated, this is a non-issue and I should not have brought it up.

oxzi: Thus, I would rephrase my initial comment here to:

Please make the comment for `id` more explicit regarding the "checkable type" part, which can be derived from a `NULL` value for the `serivce_id`.

yhabteab:
> The `get_sla_ok_percent` function does not honor it and, unless I am missing something, cannot even do this for deleted checkable objects.

That's right and it will never make use of it as well! As we discussed this last time, we will only collect the data for the time being until we get the time to finalise https://github.com/Icinga/icingadb-web/pull/710 which will make use of this newly introduced table and delegate the SLA generation from SQL procedure to PHP code and mark the `get_sla_ok_percent` procedure as a feature freeze.

doc/images/sla_history_downtime.png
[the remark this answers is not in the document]
yhabteab: You can now have another look at the images!

yhabteab:
I've pushed the new transparent Screenshots and also changed the main function a bit, so I won't be pushing anything from now on unless someone requests a change.

cmd/icingadb/main.go
```diff
@@ -242,7 +247,17 @@ func run() int {
 						})
 
 						g.Go(func() error {
+							defer close(initConfigSyncDone)
+							defer func() { initConfigSyncDone <- struct{}{} }()
```
oxzi, line 251: I would suggest merging those in one single `defer func`, as they belong together. Furthermore, the order matters and reading a stack backwards may lead to confusions.

cmd/icingadb/main.go
```diff
@@ -166,10 +166,15 @@ func run() int {
...
 						// WaitGroups for initial synchronization.
-						// Runtime updates must wait for initial synchronization to complete.
+						// Runtime updates and history pipelines must wait for the initial synchronization to
+						// complete by draining the `initConfigSyncDone` channel.
 						configInitSync := sync.WaitGroup{}
 						stateInitSync := &sync.WaitGroup{}
 
+						// A buffered channel used to notify both the runtime updates and history pipelines workers
+						// about the successful initial config sync completion including the SLA lifecycles.
+						initConfigSyncDone := make(chan struct{}, 1)
```
oxzi, line 176: Why aren't you using a `WaitGroup` here as well? As far as I can see, this channel is just used as a semaphore.

yhabteab: > Why aren't you using a `WaitGroup` here as well?

Because there are no worker groups I need to wait for! There is only a single worker that writes into that channel, so I don't need a `WaitGroup` for it.

> As far as I can see, this channel is just used as a semaphore.

Semaphore? I don't see how this relates to semaphores in any way, I only have a single worker writing just a **single** time in it to unlock the waiters.

oxzi: Pardon my French, but semaphore as in a lock for a resource.

Back to your diff: Why are you creating a buffered channel and why are you writing something into it before closing? Unless I am missing something, you can just use an unbuffered channel, wait for it in the consumers and close it in the producer. For example <https://go.dev/play/p/78zLpPxpMsW>.

doc/10-Sla-Reporting.md
```diff
@@ -0,0 +1,232 @@
...
+previous_hard_state := OK
+
+// Loop through all the state and downtime events within the provided timeframe ordered by their timestamp.
+for event in (sla_history_state, sla_history_downtime) do
+    if (event.previous_hard_state is PENDING) then
+        // A PENDING state event indicates that the host or service in question has not yet had a check result that
+        // clearly identifies its state. Consequently, such events become irrelevant for the purposes of calculating
+        // the SLA and we must exclude the duration of that PENDING state from the total time.
+        total_time = total_time - (event.event_time - last_event_time)
+    else if (previous_hard_state is greater than OK/UP
```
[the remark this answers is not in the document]
yhabteab: I'm not sure if I understand you correctly, but I don't quite agree that this being cryptic, rather it's the exact opposite. These are just normal Icinga 2 host - service states that have their own documentation except the `PENDING` state. Apart from that, I have actually listed the possible values that the [`previous_hard_state`](https://github.com/Icinga/icingadb/blob/add-create-delete-history-events/doc/10-Sla-Reporting.md#hard-state-vs-previous-hard-state) column represents and you want me to add another listing here?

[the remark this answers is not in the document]
oxzi: Please try to put yourself in a reader's position, who knows a bit about Icinga, but definitely no implementation details. When reading the comparison in this line, it may be unclear what "greater than" means in this context, as no numerical representation or hierarchy of states was introduced.

> Apart from that, I have actually listed the possible values that the [`previous_hard_state`](https://github.com/Icinga/icingadb/blob/add-create-delete-history-events/doc/10-Sla-Reporting.md#hard-state-vs-previous-hard-state) column represents and you want me to add another listing here?

Based on this table, one would assume the following order: PENDING < OK < WARNING < CRITICAL.
And what is about UNKNOWN and the host states UP and DOWN?

Coming back to the if cases, there is
1. `event.previous_hard_state is PENDING` and
2. `previous_hard_state is greater than OK/UP AND previous_hard_state is not PENDING AND checkable is not in DOWNTIME` (this one misses the `event.` prefix, but I guess it was just forgotten?).

Shouldn't it be possible to rewrite the second clause to `event.previous_hard_state is not in {OK, UP} AND checkable is not in DOWNTIME` as `event.previous_hard_state` cannot be PENDING due to the first check? Doing so eliminates any order which has to be explained otherwise.

PR #565, opened 2023-02-21
GitHub Actions: run Go lint job with at least 1.19

julianbrost:
The jobs were failing and the output contained: "note: module requires Go 1.19" (e.g. https://github.com/Icinga/icingadb/actions/runs/4233417770/jobs/7354361633)

FYI: `^1.19` means `>= 1.19 && <2.0` (https://www.npmjs.com/package/semver#caret-ranges-123-025-004)

PR #585, opened 2023-05-23
CI: upgrade dominikh/staticcheck-action to v1.3.0 to fix GHA

Al2Klimov:


julianbrost:
> to fix GHA

Well, that's why I've created #565 a while ago.

That pinning to a commit doesn't seem to do what we expect here and changes from the upstream tools still make it in here requiring a newer Go version.

PR #593, opened 2023-05-26
Merge network and database error retryability detection functions

### LINKED ISSUE #561: Wait for database to start rather than crashing
## Describe the bug

When starting, IcingaDB immediately tries to connect to the database. If it receives a rejection that states that the database is still starting, it will never retry and will instead quit immediately.

## To Reproduce

1. Create a docker-compose project with IcingaDB and PostgreSQL configured correctly, but without any depends_on field.
2. Bring the project up
3. Observe IcingaDB failing to start

## Expected behavior

IcingaDB should retry a few times before giving up, since that message is more of a "try again later" than a rejection.

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Icinga DB version: 1.1.0-1+debian11

Al2Klimov:
so that connection attempts will also be re-tried on RDBMS-specific errors, e.g. Postgres' 57P03 (the database system is starting up), not to crash. On the other hand, SQL operations which are safe to retry on SQL errors are also safe to retry on network errors.

fixes #561

## Before

https://github.com/Icinga/icingadb/issues/561#issuecomment-1564120245

## After

```
2023-05-26T12:19:58.736+0200	INFO	icingadb	Starting Icinga DB
2023-05-26T12:19:58.736+0200	INFO	icingadb	Connecting to database at 'localhost:2345'
2023-05-26T12:20:01.610+0200	WARN	database	Can't connect to database. Retrying	{"error": "pq: the database system is starting up"}
2023-05-26T12:20:08.612+0200	INFO	database	Reconnected to database{"after": "9.87615987s", "attempts": 4}
2023-05-26T12:20:08.620+0200	FATAL	icingadb	pq: relation "icingadb_schema" does not exist
can't check database schema version
...
```

### DIFF
--- a/pkg/icingadb/db.go
+++ b/pkg/icingadb/db.go
@@ -2,9 +2,7 @@ package icingadb
 
 import (
 	"context"
-	sqlDriver "database/sql/driver"
 	"fmt"
-	"github.com/go-sql-driver/mysql"
 	"github.com/icinga/icingadb/internal"
 	"github.com/icinga/icingadb/pkg/backoff"
 	"github.com/icinga/icingadb/pkg/com"
@@ -15,7 +13,6 @@ import (
 	"github.com/icinga/icingadb/pkg/retry"
 	"github.com/icinga/icingadb/pkg/utils"
 	"github.com/jmoiron/sqlx"
-	"github.com/lib/pq"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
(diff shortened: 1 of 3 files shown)


julianbrost (approved):
> are safe to retry

Do we have any cases where retrying something would actually be unsafe? Isn't the primary use of this function to say if there is a point in retrying, i.e. if we consider it fatal, there's no point in retrying if we can tell that the next 10 attempts will also fail?

But anyways, looks fine for me.

PR #601, opened 2023-06-13
Write a hint for empty arrays/dicts into `customvar_flat`

### LINKED ISSUE #597: Custom variable not visible nor filterable in Web
## Describe the bug

Some custom variables are not visible in an object's detail and are not available as filter.

## To Reproduce

* Use the config below
* Try to filter for any of the variables
  * Only `example.com` should be available
* Look at the service's detail
  * Only `example.com` should be visible

```
apply Service "Test" {
    import "generic-service"

    check_command = "random fortune"

    vars = {
      "example.com" = null
      "example[1]" = {}
      examples["foo"] = []
      examples["foo"][0] = []
    }

    assign where host.name == "docker-master"
}
```

If you query for the service's variables, you'll notice:

`SELECT * FROM customvar INNER JOIN service_customvar sc on customvar.id = sc.customvar_id INNER JOIN service s on s.id = sc.service_id WHERE s.name='Test'` returns the expected result.

`SELECT * FROM customvar_flat INNER JOIN service_customvar sc on customvar_flat.customvar_id = sc.customvar_id INNER JOIN service s on s.id = sc.service_id WHERE s.name='Test'` does not.

## Expected behavior

All four are displayed and available in filters.

yhabteab:
Just inserts empty custom vars of type `array` & `map` with `NULL` value in to the `customvar_flat` table.

fixes #597

### DIFF
--- a/pkg/icingadb/v1/customvar.go
+++ b/pkg/icingadb/v1/customvar.go
@@ -2,7 +2,6 @@ package v1
 
 import (
 	"context"
-	"fmt"
 	"github.com/icinga/icingadb/internal"
 	"github.com/icinga/icingadb/pkg/com"
 	"github.com/icinga/icingadb/pkg/contracts"
@@ -25,7 +24,7 @@ type CustomvarFlat struct {
 	CustomvarMeta    `json:",inline"`
 	Flatname         string       `json:"flatname"`
 	FlatnameChecksum types.Binary `json:"flatname_checksum"`
-	Flatvalue        string       `json:"flatvalue"`
+	Flatvalue        types.String `json:"flatvalue"`
 }
 
 func NewCustomvar() contracts.Entity {
@@ -117,11 +116,9 @@ func flattenCustomvars(ctx context.Context, g *errgroup.Group, cvs <-chan contra
 					flattened := flatten.Flatten(value, customvar.Name)
 
 					for flatname, flatvalue := range flattened {
-						var fv string
-						if flatvalue == nil {
-							fv = "null"
-						} else {
-							fv = fmt.Sprintf("%v", flatvalue)
+						var fv interface{}
+						if flatvalue.Valid {
+							fv = flatvalue.String
 						}
 
 						select {
@@ -131,7 +128,7 @@ func flattenCustomvars(ctx context.Context, g *errgroup.Group, cvs <-chan contra
 									IdMeta: IdMeta{
 										// TODO(el): Schema comment is wrong.
 										// Without customvar.Id we would produce duplicate keys here.
-										Id: utils.Checksum(objectpacker.MustPackSlice(customvar.EnvironmentId, customvar.Id, flatname, flatvalue)),
+										Id: utils.Checksum(objectpacker.MustPackSlice(customvar.EnvironmentId, customvar.Id, flatname, fv)),
 									},
 								},
 								EnvironmentMeta: EnvironmentMeta{
@@ -141,7 +138,7 @@ func flattenCustomvars(ctx context.Context, g *errgroup.Group, cvs <-chan contra
 							},
 							Flatname:         flatname,
 							FlatnameChecksum: utils.Checksum(flatname),
-							Flatvalue:        fv,
+							Flatvalue:        flatvalue,
 						}:
 						case <-ctx.Done():
 							return ctx.Err()
--- a/tests/object_sync_test.go
+++ b/tests/object_sync_test.go
@@ -3,6 +3,7 @@ package icingadb_test
 import (
 	"bytes"
 	"context"
+	"database/sql"
 	_ "embed"
 	"fmt"
 	"github.com/go-redis/redis/v8"
@@ -1194,9 +1195,9 @@ func newString(s string) *string {
 }
 
 type CustomVarTestData struct {
-	Value    interface{}       // Value to put into config or API
-	Vars     map[string]string // Expected values in customvar table
-	VarsFlat map[string]string // Expected values in customvar_flat table
+	Value    interface{}               // Value to put into config or API
+	Vars     map[string]string         // Expected values in customvar table
+	VarsFlat map[string]sql.NullString // Expected values in customvar_flat table
 }
 
 func (c *CustomVarTestData) Icinga2ConfigValue() string {
@@ -1246,20 +1247,22 @@ func (c *CustomVarTestData) verify(t require.TestingT, logger *zap.Logger, db *s
 	require.NoError(t, err, "querying customvars")
 	defer func() { _ = rows.Close() }()
 
-	expectedSrc := c.Vars
-	if flat {
-		expectedSrc = c.VarsFlat
-	}
-
 	// copy map to remove items while reading from the database
-	expected := make(map[string]string)
-	for k, v := range expectedSrc {
-		expected[k] = v
+	expected := make(map[string]sql.NullString)
+	if flat {
+		for k, v := range c.VarsFlat {
+			expected[k] = v
+		}
+	} else {
+		for k, v := range c.Vars {
+			expected[k] = toDBString(v)
+		}
 	}
 
 	for rows.Next() {
-		var cvName, cvValue string
-		err := rows.Scan(&cvName, &cvValue)
+		var cvName string
+		var cvValue sql.NullString
+		err = rows.Scan(&cvName, &cvValue)
 		require.NoError(t, err, "scanning query row")
 
 		logger.Debug("custom var from database",
@@ -1267,13 +1270,13 @@ func (c *CustomVarTestData) verify(t require.TestingT, logger *zap.Logger, db *s
 			zap.String("object-type", typeName),
 			zap.Any("object-name", name),
 			zap.String("custom-var-name", cvName),
-			zap.String("custom-var-value", cvValue))
+			zap.String("custom-var-value", cvValue.String))
 
 		if cvExpected, ok := expected[cvName]; ok {
 			assert.Equalf(t, cvExpected, cvValue, "custom var %q", cvName)
 			delete(expected, cvName)
 		} else if !ok {
-			assert.Failf(t, "unexpected custom var", "%q: %q", cvName, cvValue)
+			assert.Failf(t, "unexpected custom var", "%q: %q", cvName, cvValue.String)
 		}
 	}
 
@@ -1299,9 +1302,33 @@ func makeCustomVarTestData(t *testing.T) []*CustomVarTestData {
 			id + "-hello": `"` + id + ` world"`,
 			id + "-foo":   `"` + id + ` bar"`,
 		},
-		VarsFlat: map[string]string{
-			id + "-hello": id + " world",
-			id + "-foo":   id + " bar",
+		VarsFlat: map[string]sql.NullString{
+			id + "-hello": toDBString(id + " world"),
+			id + "-foo":   toDBString(id + " bar"),
+		},
+	})
+
+	// empty custom vars of type array and dictionaries
+	data = append(data, &CustomVarTestData{
+		Value: map[string]interface{}{
+			id + "-empty-list":        []interface{}{},
+			id + "-empty-nested-list": []interface{}{[]interface{}{}},
+			id + "-empty-dict":        map[string]interface{}{},
+			id + "-empty-nested-dict": map[string]interface{}{
+				"some-key": map[string]interface{}{},
+			},
+		},
+		Vars: map[string]string{
+			id + "-empty-list":        `[]`,
+			id + "-empty-nested-list": `[[]]`,
+			id + "-empty-dict":        `{}`,
+			id + "-empty-nested-dict": `{"some-key":{}}`,
+		},
+		VarsFlat: map[string]sql.NullString{
+			id + "-empty-list":                 {},
+			id + "-empty-nested-list[0]":       {},
+			id + "-empty-dict":                 {},
+			id + "-empty-nested-dict.some-key": {},
 		},
 	})
 
@@ -1342,29 +1369,29 @@ func makeCustomVarTestData(t *testing.T) []*CustomVarTestData {
 			id + "-nested-dict":  `{"array":["answer?",42],"dict":{"another-key":"another-value","yet-another-key":4711},"top-level-entry":"good morning"}`,
 			id + "-nested-array": `[[1,2,3],{"contains-a-map":"yes","really?":true},-42]`,
 		},
-		VarsFlat: map[string]string{
-			id + "-array[0]":                         `foo`,
-			id + "-array[1]":                         `23`,
-			id + "-array[2]":                         `bar`,
-			id + "-dict.some-key":                    `some-value`,
-			id + "-dict.other-key":                   `other-value`,
-			id + "-string":                           `hello icinga`,
-			id + "-int":                              `-1`,
-			id + "-float":                            `13.37`,
-			id + "-true":                             `true`,
-			id + "-false":                            `false`,
-			id + "-null":                             `null`,
-			id + "-nested-dict.dict.another-key":     `another-value`,
-			id + "-nested-dict.dict.yet-another-key": `4711`,
-			id + "-nested-dict.array[0]":             `answer?`,
-			id + "-nested-dict.array[1]":             `42`,
-			id + "-nested-dict.top-level-entry":      `good morning`,
-			id + "-nested-array[0][0]":               `1`,
-			id + "-nested-array[0][1]":               `2`,
-			id + "-nested-array[0][2]":               `3`,
-			id + "-nested-array[1].contains-a-map":   `yes`,
-			id + "-nested-array[1].really?":          `true`,
-			id + "-nested-array[2]":                  `-42`,
+		VarsFlat: map[string]sql.NullString{
+			id + "-array[0]":                         toDBString(`foo`),
+			id + "-array[1]":                         toDBString(`23`),
+			id + "-array[2]":                         toDBString(`bar`),
+			id + "-dict.some-key":                    toDBString(`some-value`),
+			id + "-dict.other-key":                   toDBString(`other-value`),
+			id + "-string":                           toDBString(`hello icinga`),
+			id + "-int":                              toDBString(`-1`),
+			id + "-float":                            toDBString(`13.37`),
+			id + "-true":                             toDBString(`true`),
+			id + "-false":                            toDBString(`false`),
+			id + "-null":                             toDBString(`null`),
+			id + "-nested-dict.dict.another-key":     toDBString(`another-value`),
+			id + "-nested-dict.dict.yet-another-key": toDBString(`4711`),
+			id + "-nested-dict.array[0]":             toDBString(`answer?`),
+			id + "-nested-dict.array[1]":             toDBString(`42`),
+			id + "-nested-dict.top-level-entry":      toDBString(`good morning`),
+			id + "-nested-array[0][0]":               toDBString(`1`),
+			id + "-nested-array[0][1]":               toDBString(`2`),
+			id + "-nested-array[0][2]":               toDBString(`3`),
+			id + "-nested-array[1].contains-a-map":   toDBString(`yes`),
+			id + "-nested-array[1].really?":          toDBString(`true`),
+			id + "-nested-array[2]":                  toDBString(`-42`),
 		},
 	})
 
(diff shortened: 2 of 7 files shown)


julianbrost:
Why does Icinga DB Web use the `customvar_flat` table for displaying there instead of `customvar`?

From my understanding, `customvar_flat` is an extra table to allow Icinga DB Web to implement some filtering capabilities on arrays/dictionaries in there. But after all, it's deliberately lossy, see also for example how in your example, you can't distinguish between `vars["example.com"]` and `vars["example"]["com"]` in `customvar_flat`.

lippserd:
I second @julianbrost's opinion.

nilmerg:
Because of deny lists and protection rules. They are evaluated on the database, based on the same rules the user is comfortable with using them in the restrictions and search bar.

julianbrost:
I'm not sure if I fully understand and how restrictions play into this. Is this change also useful for the filter/restriction evaluation you're doing? Or is this more like that you want to display information that's as close to to underlying data that will be used for restrictions?

nilmerg:
It's an implementation detail, nothing more. We don't need to go into detail for this, just think about custom variables that shouldn't be visible, that aren't fetched from the db at all instead of filtering them out later on.

julianbrost:
Wait, are you talking about some mechanism that controls the visibility of individual custom vars for users? If so, I wasn't even aware that this existed and I thought you were referring to host/service permission filters based on custom vars.

nilmerg:
https://icinga.com/docs/icinga-db-web/latest/doc/04-Security/#denylists
https://icinga.com/docs/icinga-db-web/latest/doc/04-Security/#variable-paths

yhabteab:
> But after all, it's deliberately lossy, see also for example how in your example, you can't distinguish between `vars["example.com"]` and `vars["example"]["com"]` in `customvar_flat`.

This also appears to be another bug that [Alex](https://github.com/Icinga/icingadb-web/issues/757#issuecomment-1561358713) didn't classify as such on Icingadb side, though it doesn't relate to this fix.

julianbrost:
I mean the primary user of all of this is Icinga DB Web, so if you say you want/need that, we can change it.

> > But after all, it's deliberately lossy, see also for example how in your example, you can't distinguish between `vars["example.com"]` and `vars["example"]["com"]` in `customvar_flat`.
> 
> This also appears to be another bug that [Alex](https://github.com/Icinga/icingadb-web/issues/757#issuecomment-1561358713) didn't classify as such on Icingadb side, though it doesn't relate to this fix.

Yeah, I knew I've recently seen something like that. However, that couldn't be changed in the `customvar_flat` table without changing the format in a way that would break existing (stored) filters (search, object restrictions, and as I've learned today, customvar denylists as well).

yhabteab:
> Why does Icinga DB Web use the `customvar_flat` table for displaying there instead of `customvar`?

> I mean the primary user of all of this is Icinga DB Web, so if you say you want/need that, we can change it.

This PR isn't supposed to fix only the representation problem in the object details, but also for filtering. You already mentioned above that the `customvar_flat` table is intended to be used for filtering, but if these variables aren't properly flattened, you can't filter by them either. Wether it's **wrong/correct** to use the `customvar_flat` table to display the custom vars in object details is quite out of this scope.

@nilmerg also has clearly described in the referenced issue. `Some custom variables are not visible in an object's detail and are not available as filter.`

So in the end, this PR only fixes some minor bugs, though doesn't break nor introduce any undesirable behaviors.

julianbrost:
There are two ways to review this. One would be to answer the question whether this writes `flatname=$x` and `flatvalue=[]` to `customvar_flat` if there is an empty list at `$x`. The other would be to understand the actual problem and see if this is the most viable solution.

Regarding filtering, can you provide an example how this change would actually be useful in this regard?

I tried to play around with list filters quickly, and added `vars.mixed_list = "[]"` (this should result in exactly the same contents in `customvar_flat` as this PR if it was an empty list) to one host and `vars.mixed_list = ["foo", "bar"]` to another. The behavior I got from Icinga DB (current icinga/icingaweb2:master image on Docker Hub) with that was quite strange: when I type "mixed" into the search bar, two visually indistinguishable suggestion "Host mixed_list" show up, when I actually use them, one expands to column `host.vars.mixed_list` and suggests `{}` as values, the other expands to column `host.vars.mixed_list[*]` and suggests `foo`/`bar` as values. Also, filtering for `host.vars.mixed_list=*` only gives the host where `vars.mixed_list` is an empty list, not the one where it actually contains values. With the `customvar_flat` table without this PR, what Icinga DB Web does works perfectly fine for "contains an element that matches the filter" (which does not have to know about empty lists, that's why these are not written to `customvar_flat` so far) and with this change, things get weird.

So does this PR do what it advertises? Probably yes. Am I convinced that what it does is actually useful? Not really.

yhabteab:
> when I type "mixed" into the search bar, two visually indistinguishable suggestion "Host mixed_list" show up,

This might yet another cosmetic bug that needs to be fixed but behaves exactly like how I would expect.

> Also, filtering for `host.vars.mixed_list=*` only gives the host where `vars.mixed_list` is an empty list, not the one where it actually contains values.

This also behaves exactly like how I would expect it to behave, since your custom variable `mixed_list` that actually contains some values in it does not have `mixed_list` as a flat name, but `mixed_list[0]` or whatever. If you want to match checkables that have the custom variable `mixed_list` set, you need to adjust your filter to something like this.

Bildschirmfoto 2023-06-14 um 15 24 11

julianbrost:
I understand what's going on and that's basically why I tried this, to check if Icinga DB Web would do any fancy magic that expands filter conditions in ways I didn't expect it to. For you, it works as expected because you know the implementation of the filters and what happens is exactly what was implemented.

But does it make sense for a user? Would users typically understand that to filter hosts that have `vars.test` set to anything (something I'd expect to be able to do if `hosts.vars.test=*` matched an empty list), they have to build an OR condition over the following three filters?
- `hosts.vars.test=*` - all hosts where it's a scalar value or an empty list or an empty dict (or a complex object like a function that got mangled to a string during serialization)
- `hosts.vars.test[*]*=*` - all hosts where it's a non-empty list
- `hosts.vars.test.*=*` - all hosts where it's a non-empty dictionary

Adapting your suggestion from the screenshot, this would be `host.vars.test*=*` here, which might be good enough in some cases but is not really correct as it would also match `vars.testy` as well.

julianbrost:
It could make sense to always emit `[]`/`{}` even for non-empty lists/dicts as a kind of "there is a list/dict here" marker without making any statement about its contents. That would make the `host.vars.test=*` filter just work for "give me everything where this is set to anything". On the other hand, it might be strange if `host.vars.test=[]` would also return hosts with non-empty lists there.

Maybe it could make sense to add some `type` column, that can then be used to mark "there's a list here" without providing a `flatvalue` at that particular `flatname`.

But in any case, it would make sense to understand what the underlying issues are. For displaying all vars that's clear, but for filtering, examples of what should be possible but can't be expressed at the moment would be helpful (buzzword: user stories).

yhabteab:
> This also behaves exactly like how I would expect it to behave, since your custom variable `mixed_list` that actually contains some values in it does not have `mixed_list` as a flat name, but `mixed_list[0]` or whatever.

Actually, Icinga DB Web should solve this internally automatically (filtering by `mixed_list=*` should also match custom variables that have `mixed_list[*]` as flatname), just like in the monitoring module, though @nilmerg can say more about this.

> But in any case, it would make sense to understand what the underlying issues are. For displaying all vars that's clear, but for filtering, examples of what should be possible but can't be expressed at the moment would be helpful (buzzword: user stories).

Isn't the goal of Icinga DB to be able to replace the monitoring module? If the user was able to filter for something like `_service_mixed_list={}` or `_service_mixed_list=*` in monitoring module, shouldn't Icinga DB Web provide the same functionality?
Bildschirmfoto 2023-06-14 um 16 44 59
Bildschirmfoto 2023-06-14 um 16 45 24

julianbrost:
Suggestion from a whiteboard discussion:
- Make `flatvalue` NULL-able.
- For an empty list `vars.empty_list = []`, emit `flatname = 'empty_list[]'`, `flatvalue=NULL`.
- For an empty dict `vars.empty_dict = {}`, emit `flatname = 'empty_dict.'`, `flatvalue=NULL`.

`flatvalue=NULL` should allows filter in Icinga DB Web work as. NULL is not a value, so it does not appear like there would be magic values appearing. Web may need some adaptions in the search bar to provide proper suggestions for `flatvalue` ending with `[]` and `.`.

julianbrost:
Does this work properly without any extra changes in Icinga DB Web? Or is there already a PR?

tests/go.mod
```diff
@@ -3,15 +3,16 @@ module github.com/icinga/icingadb/tests
...
 	github.com/go-sql-driver/mysql v1.6.0
-	github.com/goccy/go-yaml v1.9.5
+	github.com/goccy/go-yaml v1.9.6
 	github.com/google/uuid v1.3.0
 	github.com/icinga/icinga-testing v0.0.0-20220516144008-9600081b7a69
-	github.com/jmoiron/sqlx v1.3.4
-	github.com/lib/pq v1.10.5
-	github.com/stretchr/testify v1.7.0
-	go.uber.org/zap v1.21.0
+	github.com/icinga/icingadb v1.1.0
```
julianbrost, line 11: It was intentional that the tests module did not import anything from the actual icingadb module but rather only tests that code by running the binary (and as a nice benefit, this avoid the question which version of icingadb should be used here). From what I can tell, it looks like you're only using `types.String` from it, so I'd suggest to just go with `sql.NullString` from the standard library instead and drop the import.

tests/object_sync_test.go
```diff
@@ -1246,34 +1247,42 @@ func (c *CustomVarTestData) verify(t require.TestingT, logger *zap.Logger, db *s
...
-		expected[k] = v
+	expected := make(map[string]sql.NullString)
+	if flat {
+		for k, v := range c.VarsFlat {
+			expected[k] = v
+		}
+	} else {
+		for k, v := range c.Vars {
+			expected[k] = toDBString(v)
+		}
```
julianbrost, line 1259: If `v == ""` would `NULL` really be expected instead of `''`?

tests/object_sync_test.go
```diff
@@ -1299,9 +1308,33 @@ func makeCustomVarTestData(t *testing.T) []*CustomVarTestData {
...
+			id + "-empty-list":        `[]`,
+			id + "-empty-nested-list": `[[]]`,
+			id + "-empty-dict":        `{}`,
+			id + "-empty-nested-dict": `{"some-key":{}}`,
+		},
+		VarsFlat: map[string]sql.NullString{
+			id + "-empty-list[]":                toDBString(""),
+			id + "-empty-nested-list[0][]":      toDBString(""),
+			id + "-empty-dict.":                 toDBString(""),
+			id + "-empty-nested-dict.some-key.": toDBString(""),
```
julianbrost, line 1337: This could just use `types.String{}`.

tests/object_sync_test.go
```diff
@@ -1400,16 +1433,25 @@ func makeCustomVarTestData(t *testing.T) []*CustomVarTestData {
...
 }
+
+func toDBString(str string) sql.NullString {
+	result := sql.NullString{String: str, Valid: true}
+	if str == "" {
+		result.Valid = false
+	}
+
+	return result
+}
```
julianbrost, line 1457: If my intuition in my comment on `expected[k] = toDBString(v)` is correct, it should be possible to remove the `str == ""` case here as I don't think there's a situation in the tests where we would want this. This would then allow using `toDBString(str)` as a shortcut for `sql.NullString{String: str, Valid: true}` (whereas `NULL` can be achieved with `types.String{}`).

tests/object_sync_test.go
```diff
@@ -1246,34 +1247,42 @@ func (c *CustomVarTestData) verify(t require.TestingT, logger *zap.Logger, db *s
...
-		err := rows.Scan(&cvName, &cvValue)
+		var cvName string
+		var cvValue sql.NullString
+		var err error
+		if flat {
+			err = rows.Scan(&cvName, &cvValue)
+		} else {
+			err = rows.Scan(&cvName, &cvValue.String)
+			cvValue.Valid = true
+		}
```
julianbrost, line 1271: You shouldn't need this distinction, a value from a `NOT NULL` column will scan just fine into a null-able type in Go.

tests/object_sync_test.go
```diff
@@ -1342,29 +1375,29 @@ func makeCustomVarTestData(t *testing.T) []*CustomVarTestData {
...
+			id + "-array[2]":                         toDBString(`bar`),
+			id + "-dict.some-key":                    toDBString(`some-value`),
+			id + "-dict.other-key":                   toDBString(`other-value`),
+			id + "-string":                           toDBString(`hello icinga`),
+			id + "-int":                              toDBString(`-1`),
+			id + "-float":                            toDBString(`13.37`),
+			id + "-true":                             toDBString(`true`),
+			id + "-false":                            toDBString(`false`),
+			id + "-null":                             toDBString("null"),
+			id + "-nested-dict.dict.another-key":     toDBString(`another-value`),
```
julianbrost, line 1390: Not sure why you're changing the quotes around `null` here (and two more times below). Both do the same in this case.

julianbrost (changes requested):

tests/object_sync_test.go
yhabteab: `customvar.value` isn't nullable.

julianbrost: So if there was a test with an empty string variable, with the current state of `toDBString()`, this would expect `NULL` in a column that can't be null. With the change I suggested in https://github.com/Icinga/icingadb/pull/601#discussion_r1234088452, it would be fine.

pkg/icingadb/v1/customvar.go
```diff
@@ -117,11 +118,16 @@ func flattenCustomvars(ctx context.Context, g *errgroup.Group, cvs <-chan contra
...
-							fv = "null"
+							fv = types.String{NullString: sql.NullString{String: "null", Valid: true}}
 						} else {
-							fv = fmt.Sprintf("%v", flatvalue)
+							_, isSlice := flatvalue.([]interface{})
+							if _, ok := flatvalue.(map[string]interface{}); ok || isSlice {
+								fv = types.String{NullString: sql.NullString{String: "", Valid: false}}
+							} else {
+								fv = types.String{NullString: sql.NullString{String: fmt.Sprintf("%v", flatvalue), Valid: true}}
+							}
```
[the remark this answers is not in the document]
julianbrost: > > It looks weird mapping `nil` to `'null'`
> 
> That's not something introduced with this PR.

Yes and no. "weird" was referring to the combination of both `'null'` and `NULL` handling (just saw that my sentence didn't survive some rephrasing I did while writing it so I edited it so that it should hopefully be grammatically correct now). This PR adds the need to insert `NULL`, so it's questionable if it still makes sense mapping `nil` to `'null'` as this now necessitates some other magic value for actual `NULL`.

> > and any (including non-empty) []interface{} to map[string]interface{} as NULL here.
> 
> That's not quite correct. There can be neither `non-empty` `[]interface{}` nor `maps` as flatvalue. But if it did, that would be a bug in the `Flatten()` function.

Non-empty isn't really the point here. It's that it uses surprising values for representing `NULL` which wouldn't be the case if `Flatten()` just returned `map[string]types.String`. Is there a downside in doing that?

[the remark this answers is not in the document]
yhabteab: >  Is there a downside in doing that?

No. But we still have to map `nil` values to `null` string, otherwise, you won't be able to filter by something `host.some_custom_vars=null`.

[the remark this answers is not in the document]
julianbrost: `Flatten()` would then either add `types.String{}` or `types.String{Value: "null", Valid: true}` to the map it returns. Then it would be clear inside that function what's going on there instead of it returning some non-flat (= non-scalar, I guess) values that are magically mapped to `NULL` somewhere else.

pkg/icingadb/v1/customvar.go
```diff
@@ -117,21 +116,14 @@ func flattenCustomvars(ctx context.Context, g *errgroup.Group, cvs <-chan contra
...
-
 						select {
 						case flatCustomvars <- &CustomvarFlat{
 							CustomvarMeta: CustomvarMeta{
 								EntityWithoutChecksum: EntityWithoutChecksum{
 									IdMeta: IdMeta{
 										// TODO(el): Schema comment is wrong.
 										// Without customvar.Id we would produce duplicate keys here.
-										Id: utils.Checksum(objectpacker.MustPackSlice(customvar.EnvironmentId, customvar.Id, flatname, flatvalue)),
+										Id: utils.Checksum(objectpacker.MustPackSlice(customvar.EnvironmentId, customvar.Id, flatname, flatvalue.String)),
```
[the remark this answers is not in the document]
yhabteab: > I think it will at least rewrite all non-string values (for those, the ID should stay identical).

I think the only way to achieve this is to revert my changes.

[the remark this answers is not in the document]
julianbrost: That's not what I was trying to say with that: If `flatvalue` in the old code was of type `string`, the ID generated by the new code should be the same. Only those for values of other types would change.

Now it's up to us to assess the effects of that and depending on the result, those change may or may not be revert.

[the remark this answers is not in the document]
julianbrost: `customvar.id` should not be affected by the change, just `customvar_flat.id`. And this should really be referenced nowhere, so it might be fine to change this if we say it's okay to potentially rewrite a good part of the `customvar_flat` table on the first sync after the upgrade.

nilmerg (dismissed):
The current attempt to solve the display and search issues with empty variables doesn't make much sense to me anymore.
Take a look at the search bar changes for example. We need the suffixes (`.` and `[]`) to show the user that a given variable is an empty dict or list (respectively).
Though, if one object uses the name "foo" to define an integer and another object uses it to define a string, we just show the name "foo" once and then the user can differentiate by the search value.

That is, we ignore the fact that the value of a variable might be of interest (name conflict) and we don't (empty lists/dicts).

So the suffixes are an entirely new thing, just to indicate empty lists or dicts. There's no other use or benefit. The trailing dot also causes issues with other parts in the Web code base, which have no notion of "custom variables". (Which they shouldn't have, that's fine)

So let's drop the suffixes, please. Yes, the searchbar cannot indicate an empty dict or list then anymore.

Though, what I'd keep is that empty lists and dicts are written to the customvar_flat table. Otherwise they're not shown anymore in the searchbar and are not visible in an object detail.
I also don't see this as a deviation of what the customvar_flat table is meant to be. Technically, it contains all concrete custom variable values. And their paths. But not any of the intermediate values or types. That's also the reason why empty variables weren't there in the first place, as they have no concrete value, as they're empty. Though, isn't the empty value the concrete value then? So, inserting a flatvalue of `NULL` into the database for such variables, is fine to me with regard to these assumptions.

With a flatvalue of `NULL` for only empty dicts and lists, the search bar can show them, the user can filter for them and they can be displayed in an object detail.

julianbrost:
> Though, if one object uses the name "foo" to define an integer and another object uses it to define a string, we just show the name "foo" once and then the user can differentiate by the search value.

Isn't that just a side effect of everything being represented as strings in the database? When it's the same name mixed with values of type string and array on different objects, it's shown twice at the moment.

> Though, isn't the empty value the concrete value then?

It's whatever we define it to be. For the current implementation (as in current master), there is a value if there is a scalar value at that path/flatname.

> With a flatvalue of `NULL` for only empty dicts and lists, the search bar can show them, the user can filter for them and they can be displayed in an object detail.

So empty arrays and dicts will become indistinguishable in the database just like they [already are in the current PHP implementation](https://github.com/Icinga/icingadb-web/pull/779#issuecomment-1607227829)?

> the search bar can show them, the user can filter for them

Can you briefly explain what filters could be performed and how these interact with empty arrays (or also empty dicts for that matter if you can't distinguish them).

julianbrost:
So in regards to expressiveness of search filters, this would allow `x!=*` (or `x!~*` as that recently changed?) to be used as a filter for "there exists an empty array or dict at `x`"?

I'm not entirely sure what that did previously. Would it never match as `flatvalue IS NULL` could never match in the past? Or could `x!=*` be used to filter for "`x` is not set"?

nilmerg:
> Sounds fine for me, even though treating empty arrays and dicts as the same feels somewhat artificially limited.

It is. But that's a topic for #610.

> One remaining question regarding search filters: how are host.vars.empty~*x* and host.vars.empty!~*x* handled? There must be a row with flatvalue (NOT) LIKE '%x%' (which implies flatvalue IS NOT NULL) and therefore the new flatvalue = NULL rows are ignored and the result remains unaffected by this change?

Correct.

julianbrost:
> > Sounds fine for me, even though treating empty arrays and dicts as the same feels somewhat artificially limited.
> 
> It is. But that's a topic for #610.

Is it? If that was resolved, it would allow you to tell whether the empty thing is at `vars["foo"]["bar"]` or `vars["foo.bar"]` but not whether that empty thing is actually an array or dict.

julianbrost (changes requested):
The PR description itself shows outdated behavior, so please update that as well.

Overall, the PR on the Go side now does what I expect it to do given the recent comments. I'd wait though and only merge this after https://github.com/Icinga/icingadb/pull/601 is fully reviewed (I don't find the current behavior of search filters very intuitive, some of that could probably be improved by better labels but not sure what exactly is supposed to be the scope of these two PRs in the end).

julianbrost (approved):
The Go part seems to work fine, but even with a freshly pulled `icinga/icingaweb2:master` image (which should include https://github.com/Icinga/icingadb-web/pull/779), I still receive two visually identical suggestion in the search bar for lists that are empty on some hosts and non-empty on others. One renders in the URL as `array`, the other as `array[*]` (± URL enconding). Is this still expected/intended @nilmerg?

julianbrost:
> To solve this, the `list[*]` filter could be labelled as _Service list (Array)_ or _Service list (contains)_.

I'd say should instead of could. At least to me, the current behavior looks more like a bug than intentional.

Issue #574, opened 2023-03-15
Icinga DB Migration Script: Inform user about wrong/missing YAML entries


LordHepipud:
I just ran into a problem on which the migration script for IDO to Icinga DB threw the following error:

```
2023-03-15T11:28:19.695+0100    FATAL   icingadb-migrate/main.go:449    Error 1048: Column 'environment_id' cannot be null
can't perform "INSERT INTO \"comment_history\" (\"entry_time\", \"author\", \"is_persistent\", \"expire_time\", \"removed_by\", \"entry_type\", \"comment_id\", \"service_id\", \"has_been_removed\", \"is_sticky\", \"environment_id\", \"endpoint_id\", \"object_type\", \"remove_time\", \"comment\", \"host_id\") VALUES (:entry_time, :author, :is_persistent, :expire_time, :removed_by, :entry_type, :comment_id, :service_id, :has_been_removed, :is_sticky, :environment_id, :endpoint_id, :object_type, :remove_time, :comment, :host_id) ON DUPLICATE KEY UPDATE \"entry_time\" = \"entry_time\""
```

The reason for that was the missing environment Id, which was set inside the YAML. However, two issues occurred on my side (full user error on my side):

* the section element was named `Icinga2` instead of `icinga2`
* The `env` entry was added without indentation of two spaces

It would be great if the script would check the YAML before the execution and validate that the section `icinga2` (or any other section mandatory) is present as well as the `env` entry is set on the correct indentation.

This would help a lot with troubleshooting.

Al2Klimov:
# TODO

* [ ] #605
* [ ] Also scan maps for unknown keys as per https://github.com/Icinga/icingadb/pull/605#issuecomment-1611358051
* [ ] Validate **known** keys

Al2Klimov:
No, I thought of hooking our validators into https://pkg.go.dev/github.com/go-playground/validator/v10#Validate.Struct

julianbrost:
What would this do better than the existing `Validate()` functions?

Al2Klimov:
E.g.

```patch
diff --git a/pkg/config/config.go b/pkg/config/config.go
index 1388ba93..ca1e5cd5 100644
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -4,6 +4,7 @@ import (
        "crypto/tls"
        "crypto/x509"
        "github.com/creasty/defaults"
+       "github.com/go-playground/validator/v10"
[...]

PR #605, opened 2023-06-19
Config parsing: fail on unknown fields and print them

Al2Klimov:
Useful against config validation or runtime failures caused by wrong field spelling or YAML indentation.

```
➜  icingadb git:(yaml.DisallowUnknownField) ✗ go run ./cmd/icingadb-migrate -c i2i.yml -t `mktemp -d`
can't parse config file: [1:1] unknown field "Icinga2"
>  1 | Icinga2:
       ^
   2 |   env: da39a3ee5e6b4b0d3255bfef95601890afd80709
   3 | ido:
   4 |   type: mysql
   5 |
exit status 2
```

refs #574

## TODO

* [x] review (non-author)
* [x] unit test that particular case in FromYAMLFile() (author)

### DIFF
--- a/cmd/icingadb-migrate/main.go
+++ b/cmd/icingadb-migrate/main.go
@@ -125,7 +125,7 @@ func parseConfig(f *Flags) (_ *Config, exit int) {
 		return nil, 2
 	}
 
-	if err := yaml.NewDecoder(cf).Decode(c); err != nil {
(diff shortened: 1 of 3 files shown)


Al2Klimov:
Regarding the issue, this is only half of the rent. But this is already an improvement on its own.

julianbrost (commented):
> review (non-author)

The code changes look fine to me.

julianbrost (approved):

PR #612, opened 2023-07-04
{Bool,Binary,UnixMilli}#MarshalJSON(): return valid JSON, not empty string

Al2Klimov:
in case an instance is null.

### DIFF
--- a/pkg/types/binary_test.go
+++ b/pkg/types/binary_test.go
@@ -1,0 +1,29 @@
+package types
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+	"unicode/utf8"
+)
+
+func TestBinary_MarshalJSON(t *testing.T) {
+	subtests := []struct {
+		name   string
+		input  Binary
+		output string
+	}{
+		{"nil", nil, `null`},
+		{"empty", make(Binary, 0, 1), `null`},
+		{"space", Binary(" "), `"20"`},
+	}
+
+	for _, st := range subtests {
+		t.Run(st.name, func(t *testing.T) {
+			actual, err := st.input.MarshalJSON()
+
+			require.NoError(t, err)
+			require.True(t, utf8.Valid(actual))
+			require.Equal(t, st.output, string(actual))
+		})
+	}
+}
--- a/pkg/types/binary.go
+++ b/pkg/types/binary.go
@@ -65,7 +65,7 @@ func (binary *Binary) UnmarshalText(text []byte) error {
 // Supports JSON null.
 func (binary Binary) MarshalJSON() ([]byte, error) {
 	if !binary.Valid() {
-		return nil, nil
+		return []byte("null"), nil
 	}
 
 	return internal.MarshalJSON(binary.String())
--- a/pkg/types/bool.go
+++ b/pkg/types/bool.go
@@ -26,7 +26,7 @@ type Bool struct {
 // MarshalJSON implements the json.Marshaler interface.
 func (b Bool) MarshalJSON() ([]byte, error) {
 	if !b.Valid {
-		return nil, nil
+		return []byte("null"), nil
 	}
 
 	return internal.MarshalJSON(b.Bool)
--- a/pkg/types/bool_test.go
+++ b/pkg/types/bool_test.go
@@ -1,0 +1,30 @@
+package types
+
+import (
+	"fmt"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"unicode/utf8"
+)
+
+func TestBool_MarshalJSON(t *testing.T) {
+	subtests := []struct {
+		input  Bool
+		output string
+	}{
+		{Bool{Bool: false, Valid: false}, `null`},
+		{Bool{Bool: false, Valid: true}, `false`},
+		{Bool{Bool: true, Valid: false}, `null`},
+		{Bool{Bool: true, Valid: true}, `true`},
+	}
+
+	for _, st := range subtests {
+		t.Run(fmt.Sprintf("Bool-%#v_Valid-%#v", st.input.Bool, st.input.Valid), func(t *testing.T) {
+			actual, err := st.input.MarshalJSON()
+
+			require.NoError(t, err)
+			require.True(t, utf8.Valid(actual))
+			require.Equal(t, st.output, string(actual))
+		})
+	}
+}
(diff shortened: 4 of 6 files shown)


yhabteab (commented):
The changes are fine, however if there were units for them from the beginning, such bugs would have been easy to spot. Thus, it would be nice if you could add the missing unit tests here as well.

pkg/types/binary_test.go
```diff
@@ -0,0 +1,40 @@
+package types
+
+import (
+	"encoding/json"
+	"reflect"
+	"testing"
+)
+
+func TestBinary_MarshalJSON(t *testing.T) {
```
yhabteab, line 9: Please use `PascalCase` and `camelCase` as test names.

yhabteab (changes requested):
I don't quite get why you're asserting the tests manually and with some type reflecting to top it all off. Why can't you just use `require.*` and `assert.*`?

pkg/types/binary_test.go
Al2Klimov: Goland suggested these. So they must be a convention or something.

Al2Klimov:
I copy&paste-d existing pkg/icingadb/objectpacker/objectpacker_test.go

yhabteab:
> I copy&paste-d existing pkg/icingadb/objectpacker/objectpacker_test.go

And why does it have to be exactly the same here? If you just want to extend the error messages and add some contexts in case of errors, you can do that with `require.NoError(t, err, "test failed")` as well. I don't see any advantage in checking the results manually with `reflect.*` and `t.Errorf()`.

pkg/types/binary_test.go
yhabteab: Wait! I have a hint from the IDE, at least for the helpers to use camel case instead of snake case. @julianbrost

julianbrost: Typically, you don't use `_` in function names in Go. But test function names seem to be a quite common exception:
```console
% grep -r 'func Test[^(]*_' "$GOPATH" | wc -l 
10014
```

yhabteab: I was referring to the auxiliary functions:
https://github.com/Icinga/icingadb/blob/b0f2aa3c42bac1651e020e404b08e65655dd1ae0/pkg/types/binary_test.go#L15-L16

pkg/types/binary_test.go
```diff
@@ -0,0 +1,25 @@
...
+func assertBinaryMarshalJSON(t *testing.T, input Binary, expect any) {
+	t.Helper()
+
+	jsn, err := input.MarshalJSON()
+	require.NoError(t, err)
+
+	var actual any
+	require.NoError(t, json.Unmarshal(jsn, &actual))
+
+	require.Equal(t, expect, actual)
```
julianbrost, line 24: The use of `require` here will result in the entire test case being terminated on failure. Currently, a failing test for `nil` would stop the test for `" "` from running entirely, so a sub-test per input would be nicer. Personally, I like a slice of structs with test case name, input, expected output that's looped over starting individual sub-tests.

julianbrost, line 24: I also find it strange that you unmarshal here instead of just providing the expected JSON as a string.

pkg/types/binary_test.go
julianbrost: Personally, I think `_` looks strange in function names in Go, but that's very subjective. Anyways, I think this discussion is now irrelevant as that was changed already.

pkg/types/binary_test.go
Al2Klimov: The JSON's whitespace can vary, but that's irrelevant. Go values are more expressive and syntax-highlighted.

julianbrost: There are no JSON arrays or objects involved here, so where should there be spaces?

Al2Klimov: Also, I want the JSON to be valid/decodable at all in the first place, that code verifies it for me.

julianbrost: I mean part of what makes this look strange to me is that it uses `Unmarshal()`, but into type `any` instead of `Binary`. And so far, there is no test for `Unmarshal()`, so my suggestion would actually be a test for both: `Binary` as input, `Marshal()`, check that it matches an expected string, `Unmarshal()` into a new `Binary`, check that it's equal to the first one. Tests everything you want and as a bonus, also adds tests for `Binary.UnmarshalJSON()`.

Al2Klimov: > There are no JSON arrays or objects involved here, so where should there be spaces?

You know what, you're right. My mistake.

pkg/types/binary_test.go
```diff
@@ -0,0 +1,27 @@
...
+		{make(Binary, 0, 1), `null`},
+		{Binary(" "), `"20"`},
+	}
+
+	for _, st := range subtests {
+		t.Run(fmt.Sprintf("Binary(%#v)", []byte(st.input)), func(t *testing.T) {
+			actual, err := st.input.MarshalJSON()
+
+			require.NoError(t, err)
+			require.Equal(t, []byte(st.output), actual)
```
julianbrost, line 24: Casting towards `string` instead of `[]byte` results in more readable output when a test fails, compare:

#### []byte
```
    --- FAIL: TestBinary_MarshalJSON/Binary([]byte(nil)) (0.00s)
        binary_test.go:24: 
            	Error Trace:	/home/jbrost/src/icingadb/pkg/types/binary_test.go:24
            	Error:      	Not equal: 
            	            	expected: []byte{0x6e, 0x75, 0x6c, 0x6c}
            	            	actual  : []byte{0x6e, 0x69, 0x6c}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1,3 +1,3 @@
            	            	-([]uint8) (len=4) {
            	            	- 00000000  6e 75 6c 6c                                       |null|
            	            	+([]uint8) (len=3) {
            	            	+ 00000000  6e 69 6c                                          |nil|
            	            	 }
            	Test:       	TestBinary_MarshalJSON/Binary([]byte(nil))
```

#### string
```
    --- FAIL: TestBinary_MarshalJSON/Binary([]byte(nil)) (0.00s)
        binary_test.go:24: 
            	Error Trace:	/home/jbrost/src/icingadb/pkg/types/binary_test.go:24
            	Error:      	Not equal: 
            	            	expected: "null"
            	            	actual  : "nil"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1 @@
            	            	-null
            	            	+nil
            	Test:       	TestBinary_MarshalJSON/Binary([]byte(nil))
```

julianbrost (approved):

PR #619, opened 2023-07-24
Make integration tests work with Go 1.20.6

julianbrost:
As a security fix, Go implemented stricter checks for the Host header in the http package. This [breaks the Go Docker client library](https://github.com/moby/moby/issues/45935) resulting in "http: invalid Host header" errors in our integration tests.

A [fix](https://github.com/moby/moby/pull/45942) was merged to the 24.0 branch, but is not yet included in a release. Therefore, this commit updates the dependency to a specific commit (`go get -u github.com/docker/docker@24.0`). Once it's included in a release, this can be switched back to using a release version.

Example of a currently failing pipeline: https://github.com/Icinga/icingadb/actions/runs/5640729645/job/15277750971

### DIFF
--- a/tests/go.mod
+++ b/tests/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f // indirect
 	github.com/docker/distribution v2.8.2+incompatible // indirect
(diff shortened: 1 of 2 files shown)


julianbrost:
Also affects the SQL tests, see https://github.com/Icinga/icingadb/actions/runs/5517316837/job/15288121537 for example. According to the GitHub Actions of this PR, this is fixed as well.

lippserd (approved):

PR #631, opened 2023-08-03
Only warn about unknown config options instead of rejecting them

julianbrost:
This is a quick proof of concept for an idea I had how to make #605 more suitable for a bug fix release.

Currently, it would just fail with an error. This PR basically parses the config like v1.1.0 did, but additionally parses the config a second time with `yaml.DisallowUnknownField()` (like the current master and therefore v1.2 will do) and issue a warning if it returns an error.

## TODO

* [ ] Do we want this?
* [ ] Should it be a bit nicer? Like maybe store the error and only log it at a later time when proper logging was initialized. But in case of an error, the warning would still be nice, so it shouldn't get lost in that case.
* [ ] Do we want a similar change in `icingadb-migrate`? This is more of a manual one-shot tool, so making it an error there right away shouldn't break anything.

### Config with unknown options and other errors

First prints the warning and then the error:

```console
$ go run ./cmd/icingadb -c <(echo unknown: 42)
WARNING: ignoring unknown config option, this will become a fatal error in Icinga DB 1.2:

[1:1] unknown field "unknown"
    >  1 | unknown: 42
           ^

invalid configuration: database host missing
```

### Unknown option in an otherwise valid config

Prints the warning and then continues to run:

```console
$ go run ./cmd/icingadb -c config.yml
WARNING: ignoring unknown config option, this will become a fatal error in Icinga DB 1.2:

[41:1] unknown field "unknown"
      38 |   options:
      39 |     notification: 14
      40 | 
    > 41 | unknown: 42
           ^

2023-08-03T17:28:54.397+0200	INFO	icingadb	Starting Icinga DB
[...running normally...]
```


julianbrost:
I've pushed a new version that delays the warning until logging is initialized. In case there is an error (and logging won't be initialized), the warning is embedded in the error message.

pkg/config/config.go
```diff
@@ -64,8 +67,16 @@ func FromYAMLFile(name string) (*Config, error) {
...
+	c.DecodeWarning = yaml.NewDecoder(bytes.NewReader(f), yaml.DisallowUnknownField()).Decode(&Config{})
+
 	if err := c.Validate(); err != nil {
-		return nil, errors.Wrap(err, "invalid configuration")
+		msg := "invalid configuration"
+		if warn := c.DecodeWarning; warn != nil {
+			return nil, fmt.Errorf("%s: %w\n\nwarning: ignored unknown config option:\n\n%v", msg, err, warn)
+		} else {
+			return nil, errors.Wrap(err, msg)
+		}
```
Al2Klimov, line 79: I doubt we need this. After all it failed because of _err_, not _warn_.

pkg/config/config_test.go
```diff
@@ -20,33 +20,34 @@ redis:
...
-			}(),
+			name:   "mini",
+			input:  miniConf,
+			output: miniOutput,
+			warn:   false,
 		},
 		{
 			name:   "mini-with-unknown",
 			input:  miniConf + "\nunknown: 42",
-			output: nil,
```
Al2Klimov, line 49: This was the only one of its kind. You've made the existing if below obsolete.

Al2Klimov (commented):
Even better (for the ones who read logs 🪵 at all).

pkg/config/config.go
```diff
@@ -64,8 +67,16 @@ func FromYAMLFile(name string) (*Config, error) {
 		return nil, errors.Wrap(err, "can't parse YAML file "+name)
 	}
 
+	// Decode again with yaml.DisallowUnknownField() (like v1.2 will do) and issue a warning if it returns an error.
+	c.DecodeWarning = yaml.NewDecoder(bytes.NewReader(f), yaml.DisallowUnknownField()).Decode(&Config{})
+
 	if err := c.Validate(); err != nil {
-		return nil, errors.Wrap(err, "invalid configuration")
+		msg := "invalid configuration"
+		if warn := c.DecodeWarning; warn != nil {
```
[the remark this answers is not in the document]
julianbrost: Shorthand for the longer name.

pkg/config/config.go
julianbrost: I think it can be quite useful, for example if things are missing because you've misnamed them:

```
invalid configuration: database host missing

warning: ignored unknown config option:

[3:3] unknown field "hostname"
       1 | database:
       2 |   type: mysql
    >  3 |   hostname: localhost
             ^
       4 |   port: 3306
       5 |   database: icingadb
       6 |   user: icingadb
       7 |   
```

Al2Klimov: Ack.

Al2Klimov (approved):
> Do we want a similar change in icingadb-migrate? This is more of a manual one-shot tool, so making it an error there right away shouldn't break anything.

You are right. To be even more specific, there are three kinds of users:

* have done migration – won't use the tool again – not affected
* doing migration – affected on interrupt – solutions:
  * wait with the update
  * just adjust the config
* didn't start migration – didn't use the tool yet – not affected by my change as such (or yours)

pkg/config/config.go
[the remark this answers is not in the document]
julianbrost: I was just trying hard to avoid breaking the next line, especially somewhere in the middle of the format string.

Al2Klimov (approved):

PR #633, opened 2023-08-04
GitHub Actions: use actions/setup-go@v4 and go-version: 1.x

julianbrost:
There was a mix of diferent action and Go versions specified throughout the GitHub Actions workflow configs. In general, we want to be compatible with the latest Go 1.x version, so just specify that instead of bumping versions everywhere when necessary.

At the same time, bump the actions/setup-go to v4, the latest currently available version. This prefers versions already cached on the runner and enables automatic caching for dependencies.

Also requires some reformatting of comments, looks like `go fmt` likes indentation within comments to look differently now.

The main motivation for this change is to allow upgrading go.uber.org/zap from 1.24.0 to 1.25.0 which no longer works with Go 1.18 (see #632).

### DIFF
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -12,9 +12,9 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - uses: actions/setup-go@v2
+      - uses: actions/setup-go@v4
         with:
-          go-version: 1.18.x
+          go-version: 1.x
       - name: Check version
         # TODO(elippmann): Needs adjustments as soon as we release from (support) branches too.
         run: |
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -11,9 +11,9 @@ jobs:
     steps:
       - run: sudo apt install -y moreutils
 
-      - uses: actions/setup-go@v2
+      - uses: actions/setup-go@v4
         with:
-          go-version: 1.18.x
+          go-version: 1.x
 
       - uses: actions/checkout@v2
 
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -16,9 +16,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions/setup-go@v2
+      - uses: actions/setup-go@v4
         with:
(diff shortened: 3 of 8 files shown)


.github/workflows/version.yml
```diff
@@ -12,9 +12,9 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - uses: actions/setup-go@v2
+      - uses: actions/setup-go@v4
         with:
-          go-version: 1.18.x
+          go-version: 1.x
```
Al2Klimov, line 17: This change and everything above it complies to our GitLab pipelines. 👍

### But the meta question is:

Should we keep doing this?

OK, nothing **should** break with a Go update. The comment fix you've pushed is a fmt-only thing. But I personally would feel more comfortable with manual Go version bumps. Maybe like our customers who don't install a .0 version. Maybe even like my Nextcloud update policy: vX isn't EOL yet = don't upgrade to v{X+1} yet

Al2Klimov (approved):
It is acceptable. (But read on.)

lippserd (approved):

.github/workflows/version.yml
lippserd: I don't see any downside in doing it like the way we do it (now).

Al2Klimov: > > Should we keep doing this?
> > OK, nothing **should** break with a Go update. The comment fix you've pushed is a fmt-only thing. But I personally would feel more comfortable with manual Go version bumps.
> 
> I mean if there was a reason to stop doing this, that would not only apply to Go versions but everything. But I really don't see a reason for this unless this would regularly cause severe problems.

The latter is a valid argument.

> > Maybe like our customers who don't install a .0 version. Maybe even like my Nextcloud update policy: vX isn't EOL yet = don't upgrade to v{X+1} yet
> 
> Weren't you pushing the use of generics early on? If we did it that way, you'd have waited for them quite a long time. 😅

This is a special case. 😅 Generics are a killer feature.

Al2Klimov: Related:

> There will not be a Go 2 that breaks Go 1 programs.

– https://go.dev/blog/compat

PR #643, opened 2023-08-10
Add dependabot config for github actions

RincewindsHat:


RincewindsHat:
In other projects we got deprecation warnings for the pipelines after some time.
The specific versions seem to be nailed to a specific `node` version which does not get updated within the version of the action. Therefore those actions will have to be updated eventually (if Github continues this practice) and IMHO it is better to get receive a PR for that beforehand instead of getting a deprecation notice unexpectedly.

Al2Klimov (approved):
I, in contrast, don't even want a reason for applying updates – GHA or not.

lippserd:
To be honest, I overlooked this PR. In the meantime, this has been fixed with PR #813.

PR #656, opened 2023-10-10
Unify check attempt data type to uint32 already used somewhere

Al2Klimov:
A float isn't necessary as in Icinga 2 Checkable#max_check_attempts and check_attempt are ints. But uint8 isn't enough for e.g. 1 check/s to get HARD after 5m (300s > 255).

### DIFF
--- a/cmd/icingadb-migrate/convert.go
+++ b/cmd/icingadb-migrate/convert.go
@@ -724,8 +724,8 @@ type stateRow = struct {
 	StateTimeUsec       uint32
 	State               uint8
 	StateType           uint8
-	CurrentCheckAttempt uint16
-	MaxCheckAttempts    uint16
+	CurrentCheckAttempt uint32
+	MaxCheckAttempts    uint32
 	LastState           uint8
 	LastHardState       uint8
 	Output              sql.NullString
@@ -798,10 +798,10 @@ func convertStateRows(
 			HardState:         row.LastHardState,
 			PreviousSoftState: row.LastState,
 			PreviousHardState: previousHardState,
-			CheckAttempt:      uint8(row.CurrentCheckAttempt),
+			CheckAttempt:      row.CurrentCheckAttempt,
 			Output:            icingadbTypes.String{NullString: row.Output},
 			LongOutput:        icingadbTypes.String{NullString: row.LongOutput},
-			MaxCheckAttempts:  uint32(row.MaxCheckAttempts),
+			MaxCheckAttempts:  row.MaxCheckAttempts,
 			CheckSource:       icingadbTypes.String{NullString: row.CheckSource},
 		})
 
--- a/doc/04-Upgrading.md
+++ b/doc/04-Upgrading.md
@@ -9,11 +9,39 @@ Please apply the `1.1.2.sql` upgrade script to your database. For package instal
 `/usr/share/icingadb/schema/mysql/upgrades/` or `/usr/share/icingadb/schema/pgsql/upgrades/`, depending on your
 database vendor.
 
+As the daemon checks the schema version, the recommended way to perform the upgrade is to stop the daemon, apply the
+schema upgrade and then start the new daemon version. If you want to minimize downtime as much as possible, it is safe
+to apply this schema upgrade while the Icinga DB v1.1.1 daemon is still running and then restart the daemon with the
+new version. Please keep in mind that depending on the distribution, your package manager may automatically attempt to
+restart the daemon when upgrading the package.
+
 !!! warning
 
     With MySQL and MariaDB, a locking issue can occur if the schema upgrade is applied while the history view is
     accessed in Icinga DB Web. This can result in the upgrade being delayed unnecessarily and blocking other queries.
     Please see [unblock history tables](#unblock-history-tables) for how to detect and resolve this situation.
+
+### Upgrading the state_history Table
+
+This release includes fixes for hosts and services reaching check attempt 256. However, on existing installations,
+the schema upgrade required to fix the history tables isn't automatically applied by `1.1.2.sql` as a rewrite of the
+whole `state_history` table is required. This can take a lot of time depending on the history size and the performance
+of the database. During this time that table will be locked exclusively and can't be accessed otherwise. This means that
+the existing history can't be viewed in Icinga Web and new history entries will be buffered in Redis.
+
+There is a separate upgrade script `optional/1.1.2-history.sql` to perform the rewrite of the `state_history` table.
+This allows you to postpone part of the upgrade to a longer maintenance window in the future, or skip it entirely
+if you deem this safe for your installation.
+
+!!! warning
+
+    Until `optional/1.1.2-history.sql` is applied, you'll have to lower `max_check_attempts` to 255 or less, otherwise
+    Icinga DB will crash with a database error if hosts/services reach check attempt 256. If you need to lower
+    `max_check_attempts` but want to keep the same timespan from an outage to a hard state, you can raise
+    `retry_interval` instead so that `max_check_attempts * retry_interval` stays the same.
+
+If you apply it, be sure that `1.1.2.sql` was already applied before. Do not interrupt it! At best use tmux/screen not
+to lose your SSH session.
 
 ### Unblock History Tables
 
--- a/schema/mysql/schema.sql
+++ b/schema/mysql/schema.sql
@@ -292,7 +292,7 @@ CREATE TABLE host_state (
   hard_state tinyint unsigned NOT NULL,
   previous_soft_state tinyint unsigned NOT NULL,
   previous_hard_state tinyint unsigned NOT NULL,
-  check_attempt tinyint unsigned NOT NULL,
+  check_attempt int unsigned NOT NULL,
   severity smallint unsigned NOT NULL,
 
   output longtext DEFAULT NULL,
@@ -460,7 +460,7 @@ CREATE TABLE service_state (
   hard_state tinyint unsigned NOT NULL,
   previous_soft_state tinyint unsigned NOT NULL,
   previous_hard_state tinyint unsigned NOT NULL,
-  check_attempt tinyint unsigned NOT NULL,
+  check_attempt int unsigned NOT NULL,
   severity smallint unsigned NOT NULL,
 
   output longtext DEFAULT NULL,
@@ -1147,7 +1147,7 @@ CREATE TABLE state_history (
   hard_state tinyint unsigned NOT NULL,
   previous_soft_state tinyint unsigned NOT NULL,
   previous_hard_state tinyint unsigned NOT NULL,
-  check_attempt tinyint unsigned NOT NULL,
+  check_attempt int unsigned NOT NULL, -- may be a tinyint unsigned, see https://icinga.com/docs/icinga-db/latest/doc/04-Upgrading/#upgrading-to-icinga-db-v112
   output longtext DEFAULT NULL,
   long_output longtext DEFAULT NULL,
   max_check_attempts int unsigned NOT NULL,
@@ -1343,4 +1343,4 @@ CREATE TABLE icingadb_schema (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
 
 INSERT INTO icingadb_schema (version, timestamp)
-  VALUES (4, UNIX_TIMESTAMP() * 1000);
+  VALUES (5, UNIX_TIMESTAMP() * 1000);
--- a/schema/mysql/upgrades/1.1.2.sql
+++ b/schema/mysql/upgrades/1.1.2.sql
@@ -2,3 +2,12 @@ UPDATE icingadb_schema SET timestamp = UNIX_TIMESTAMP(timestamp / 1000) * 1000 W
 
 ALTER TABLE history ADD INDEX idx_history_event_time_event_type (event_time, event_type) COMMENT 'History filtered/ordered by event_time/event_type';
 ALTER TABLE history DROP INDEX idx_history_event_time;
+
+ALTER TABLE host_state MODIFY COLUMN check_attempt int unsigned NOT NULL;
+
+ALTER TABLE service_state MODIFY COLUMN check_attempt int unsigned NOT NULL;
+
+ALTER TABLE state_history MODIFY COLUMN check_attempt tinyint unsigned NOT NULL COMMENT 'optional schema upgrade not applied yet, see https://icinga.com/docs/icinga-db/latest/doc/04-Upgrading/#upgrading-to-icinga-db-v112';
+
+INSERT INTO icingadb_schema (version, timestamp)
+  VALUES (5, UNIX_TIMESTAMP() * 1000);
--- a/pkg/icingadb/db.go
+++ b/pkg/icingadb/db.go
@@ -93,8 +93,8 @@ func NewDb(db *sqlx.DB, logger *logging.Logger, options *Options) *DB {
 }
 
 const (
-	expectedMysqlSchemaVersion    = 4
-	expectedPostgresSchemaVersion = 2
+	expectedMysqlSchemaVersion    = 5
+	expectedPostgresSchemaVersion = 3
 )
 
 // CheckSchema asserts the database schema of the expected version being present.
--- a/pkg/icingadb/v1/checkable.go
+++ b/pkg/icingadb/v1/checkable.go
@@ -30,7 +30,7 @@ type Checkable struct {
 	IconImageAlt          string       `json:"icon_image_alt"`
 	IconImageId           types.Binary `json:"icon_image_id"`
 	IsVolatile            types.Bool   `json:"is_volatile"`
-	MaxCheckAttempts      float64      `json:"max_check_attempts"`
+	MaxCheckAttempts      uint32       `json:"max_check_attempts"`
 	Notes                 string       `json:"notes"`
 	NotesUrlId            types.Binary `json:"notes_url_id"`
 	NotificationsEnabled  types.Bool   `json:"notifications_enabled"`
--- a/pkg/icingadb/v1/history/state.go
+++ b/pkg/icingadb/v1/history/state.go
@@ -14,7 +14,7 @@ type StateHistory struct {
 	HardState          uint8           `json:"hard_state"`
 	PreviousSoftState  uint8           `json:"previous_soft_state"`
 	PreviousHardState  uint8           `json:"previous_hard_state"`
-	CheckAttempt       uint8           `json:"check_attempt"`
+	CheckAttempt       uint32          `json:"check_attempt"`
 	Output             types.String    `json:"output"`
 	LongOutput         types.String    `json:"long_output"`
 	MaxCheckAttempts   uint32          `json:"max_check_attempts"`
--- a/pkg/icingadb/v1/state.go
+++ b/pkg/icingadb/v1/state.go
@@ -9,7 +9,7 @@ type State struct {
 	EnvironmentMeta           `json:",inline"`
 	AcknowledgementCommentId  types.Binary               `json:"acknowledgement_comment_id"`
 	LastCommentId             types.Binary               `json:"last_comment_id"`
-	CheckAttempt              uint8                      `json:"check_attempt"`
+	CheckAttempt              uint32                     `json:"check_attempt"`
 	CheckCommandline          types.String               `json:"check_commandline"`
 	CheckSource               types.String               `json:"check_source"`
 	SchedulingSource          types.String               `json:"scheduling_source"`
--- a/schema/mysql/upgrades/optional/1.1.2-history.sql
+++ b/schema/mysql/upgrades/optional/1.1.2-history.sql
@@ -1,0 +1,1 @@
+ALTER TABLE state_history MODIFY COLUMN check_attempt int unsigned NOT NULL;
--- a/schema/pgsql/schema.sql
+++ b/schema/pgsql/schema.sql
@@ -405,7 +405,7 @@ CREATE TABLE host_state (
   hard_state tinyuint NOT NULL,
   previous_soft_state tinyuint NOT NULL,
   previous_hard_state tinyuint NOT NULL,
-  check_attempt tinyuint NOT NULL,
+  check_attempt uint NOT NULL,
   severity smalluint NOT NULL,
 
   output text DEFAULT NULL,
@@ -675,7 +675,7 @@ CREATE TABLE service_state (
   hard_state tinyuint NOT NULL,
   previous_soft_state tinyuint NOT NULL,
   previous_hard_state tinyuint NOT NULL,
-  check_attempt tinyuint NOT NULL,
+  check_attempt uint NOT NULL,
   severity smalluint NOT NULL,
 
   output text DEFAULT NULL,
@@ -1846,7 +1846,7 @@ CREATE TABLE state_history (
   hard_state tinyuint NOT NULL,
   previous_soft_state tinyuint NOT NULL,
   previous_hard_state tinyuint NOT NULL,
-  check_attempt tinyuint NOT NULL,
+  check_attempt uint NOT NULL, -- may be a tinyuint, see https://icinga.com/docs/icinga-db/latest/doc/04-Upgrading/#upgrading-to-icinga-db-v112
   output text DEFAULT NULL,
   long_output text DEFAULT NULL,
   max_check_attempts uint NOT NULL,
@@ -2181,4 +2181,4 @@ CREATE TABLE icingadb_schema (
 ALTER SEQUENCE icingadb_schema_id_seq OWNED BY icingadb_schema.id;
 
 INSERT INTO icingadb_schema (version, timestamp)
-  VALUES (2, extract(epoch from now()) * 1000);
+  VALUES (3, extract(epoch from now()) * 1000);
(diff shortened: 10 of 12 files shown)


Al2Klimov:
## Before

```
2023-10-10 11:36:23 2023-10-10T09:36:23.002Z    FATAL   icingadb        strconv.ParseUint: parsing "256": value out of range
2023-10-10 11:36:23 can't parse check_attempt into the uint8 HostState#State.CheckAttempt: 256
2023-10-10 11:36:23 github.com/icinga/icingadb/pkg/structify.structifyMapByTree
2023-10-10 11:36:23     /icingadb-src/pkg/structify/structify.go:97
2023-10-10 11:36:23 github.com/icinga/icingadb/pkg/structify.structifyMapByTree
2023-10-10 11:36:23     /icingadb-src/pkg/structify/structify.go:102
2023-10-10 11:36:23 github.com/icinga/icingadb/pkg/structify.MakeMapStructifier.func1
2023-10-10 11:36:23     /icingadb-src/pkg/structify/structify.go:42
2023-10-10 11:36:23 github.com/icinga/icingadb/pkg/icingadb.(*RuntimeUpdates).Sync.structifyStream.func6
2023-10-10 11:36:23     /icingadb-src/pkg/icingadb/runtime_updates.go:324
2023-10-10 11:36:23 golang.org/x/sync/errgroup.(*Group).Go.func1
2023-10-10 11:36:23     /go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75
2023-10-10 11:36:23 runtime.goexit
2023-10-10 11:36:23     /usr/local/go/src/runtime/asm_amd64.s:1650
2023-10-10 11:36:23 can't structify map map[string]interface {}{"check_attempt":"256", "check_commandline":"'dummy'", "check_source":"master2", "check_timeout":"60000", "checksum":"b2ce7a04e37e4dd2bfad51f78d7a055244fef255", "environment_id":"36470b09ec644b7dc09863cdf0fbd4e68bc7f91b", "execution_time":"0", "hard_state":"1", "host_id":"acf2ca3316e8f6ca6ee42645d13a539445b62aaf", "id":"acf2ca3316e8f6ca6ee42645d13a539445b62aaf", "in_downtime":"0", "is_acknowledged":"0", "is_active":"1", "is_flapping":"0", "is_handled":"0", "is_problem":"1", "is_reachable":"1", "last_state_change":"0", "last_update":"1696930582403", "latency":"0", "next_check":"1696930582911", "next_update":"1696930582913", "output":"Check was successful.", "previous_hard_state":"99", "previous_soft_state":"99", "redis_key":"icinga:host:state", "runtime_type":"upsert", "scheduling_source":"master2", "severity":"2112", "soft_state":"1", "state_type":"0"} by tree []structify.structBranch{structify.structBranch{field:0, leaf:"", subTree:[]structify.structBranch{structify.structBranch{field:0, leaf:"", subTree:[]structify.structBranch{structify.structBranch{field:0, leaf:"", subTree:[]structify.structBranch{structify.structBranch{field:0, leaf:"", subTree:[]structify.structBranch{structify.structBranch{field:0, leaf:"id", subTree:[]structify.structBranch(nil)}}}}}, structify.structBranch{field:1, leaf:"", subTree:[]structify.structBranch{structify.structBranch{field:0, leaf:"checksum", subTree:[]structify.structBranch(nil)}}}}}, structify.structBranch{field:1, leaf:"", subTree:[]structify.structBranch{structify.structBranch{field:0, leaf:"environment_id", subTree:[]structify.structBranch(nil)}}}, structify.structBranch{field:2, leaf:"acknowledgement_comment_id", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:3, leaf:"last_comment_id", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:4, leaf:"check_attempt", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:5, leaf:"check_commandline", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:6, leaf:"check_source", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:7, leaf:"scheduling_source", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:8, leaf:"execution_time", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:9, leaf:"hard_state", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:10, leaf:"in_downtime", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:11, leaf:"is_acknowledged", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:12, leaf:"is_flapping", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:13, leaf:"is_handled", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:14, leaf:"is_problem", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:15, leaf:"is_reachable", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:16, leaf:"last_state_change", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:17, leaf:"last_update", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:18, leaf:"latency", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:19, leaf:"long_output", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:20, leaf:"next_check", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:21, leaf:"next_update", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:22, leaf:"output", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:23, leaf:"performance_data", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:24, leaf:"normalized_performance_data", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:25, leaf:"previous_soft_state", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:26, leaf:"previous_hard_state", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:27, leaf:"severity", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:28, leaf:"soft_state", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:29, leaf:"state_type", subTree:[]structify.structBranch(nil)}, structify.structBranch{field:30, leaf:"check_timeout", subTree:[]structify.structBranch(nil)}}}, structify.structBranch{field:1, leaf:"host_id", subTree:[]structify.structBranch(nil)}}
2023-10-10 11:36:23 github.com/icinga/icingadb/pkg/structify.MakeMapStructifier.func1
2023-10-10 11:36:23     /icingadb-src/pkg/structify/structify.go:42
2023-10-10 11:36:23 github.com/icinga/icingadb/pkg/icingadb.(*RuntimeUpdates).Sync.structifyStream.func6
2023-10-10 11:36:23     /icingadb-src/pkg/icingadb/runtime_updates.go:324
2023-10-10 11:36:23 golang.org/x/sync/errgroup.(*Group).Go.func1
2023-10-10 11:36:23     /go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75
2023-10-10 11:36:23 runtime.goexit
2023-10-10 11:36:23     /usr/local/go/src/runtime/asm_amd64.s:1650
2023-10-10 11:36:23 can't structify values map[string]interface {}{"check_attempt":"256", "check_commandline":"'dummy'", "check_source":"master2", "check_timeout":"60000", "checksum":"b2ce7a04e37e4dd2bfad51f78d7a055244fef255", "environment_id":"36470b09ec644b7dc09863cdf0fbd4e68bc7f91b", "execution_time":"0", "hard_state":"1", "host_id":"acf2ca3316e8f6ca6ee42645d13a539445b62aaf", "id":"acf2ca3316e8f6ca6ee42645d13a539445b62aaf", "in_downtime":"0", "is_acknowledged":"0", "is_active":"1", "is_flapping":"0", "is_handled":"0", "is_problem":"1", "is_reachable":"1", "last_state_change":"0", "last_update":"1696930582403", "latency":"0", "next_check":"1696930582911", "next_update":"1696930582913", "output":"Check was successful.", "previous_hard_state":"99", "previous_soft_state":"99", "redis_key":"icinga:host:state", "runtime_type":"upsert", "scheduling_source":"master2", "severity":"2112", "soft_state":"1", "state_type":"0"}
2023-10-10 11:36:23 github.com/icinga/icingadb/pkg/icingadb.(*RuntimeUpdates).Sync.structifyStream.func6
2023-10-10 11:36:23     /icingadb-src/pkg/icingadb/runtime_updates.go:326
2023-10-10 11:36:23 golang.org/x/sync/errgroup.(*Group).Go.func1
2023-10-10 11:36:23     /go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75
2023-10-10 11:36:23 runtime.goexit
2023-10-10 11:36:23     /usr/local/go/src/runtime/asm_amd64.s:1650
```

## After

No crash so far.

Al2Klimov:
ref/IP/48137

A41susan (approved):
looks fine

cmd/icingadb-migrate/convert.go
julianbrost, line 1: There's another place in this file where this is limited to 16 bit:

https://github.com/Icinga/icingadb/blob/f26ecbf6daf199a452e988f9117e14528008b5fe/cmd/icingadb-migrate/convert.go#L742-L743

This does match the IDO MySQL schema (which uses `smallint`) but the PostgreSQL schema uses `integer`:
https://github.com/Icinga/icinga2/blob/2c9117b4f71e00b2072e7dbe6c4ea4e48c882a87/lib/db_ido_pgsql/schema/pgsql.sql#L1330-L1331

julianbrost (changes requested):
This should have a corresponding schema change as well. Currently, the `check_attempt` columns have type `tinyint unsigned` (MySQL) or `tinyuint` (PostgreSQL).

julianbrost (commented):
Looks good overall. However, given that the schema upgrade changes column types in history tables (thus potentially resulting in time-consuming rewrites thereof), I'm not yet sure if we'd want to backport this and what this means for the schema version number change (which I want to consider before a final approval).

lippserd (commented):
Our upgrading docs should mention this bug and the schema fix and why it is not included in the upgrade scripts. It should also mention a workaround for this bug, which is to increase `retry_interval` so that `time_to_hard_state = 255 * x * retry_interval`. We should also document the statements to be executed and make clear that they take a very long time and should not be aborted under any circumstances.

schema/mysql/upgrades/1.1.2.sql
```diff
@@ -0,0 +1,10 @@
+-- See also ./1.1.2-history.sql!
+
+ALTER TABLE host_state MODIFY COLUMN check_attempt int unsigned NOT NULL;
+
+ALTER TABLE service_state MODIFY COLUMN check_attempt int unsigned NOT NULL;
+
+ALTER TABLE state_history MODIFY COLUMN check_attempt tinyint unsigned NOT NULL COMMENT 'optional schema/mysql/upgrades/1.1.2-history.sql not applied, yet';
```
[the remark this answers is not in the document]
julianbrost: The purpose is to have a hint when looking at the schema of a database where the optional part of the upgrade wasn't applied.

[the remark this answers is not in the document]
julianbrost: > you can look into the actual type to determine whether it's applied

You can do this if you know what you're looking for.

However, fast-forward a few years, you don't know what you're looking for, something weird happens with that table, couldn't that comment be a useful hint for this kind of situation?

schema/mysql/schema.sql
```diff
@@ -1343,4 +1343,4 @@ CREATE TABLE icingadb_schema (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
 
 INSERT INTO icingadb_schema (version, timestamp)
-  VALUES (4, CURRENT_TIMESTAMP() * 1000);
+  VALUES (5, CURRENT_TIMESTAMP() * 1000);
```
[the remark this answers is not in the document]
Al2Klimov: Sure, there are already upgrade files w/o even such changes. But the specific things our 1.1.2 upgrade files already do () don't break the daemon and are even idempotent:

* #700
* #710

So they don't need a new schema version, but this PR does.

julianbrost:
> Edit: I don't like the additional upgrade scripts. Since there's something to be written in the upgrading docs anyway, the statements should be there.

There and only there? I mean if these exist as a file like all the other ones, one can apply them the same way and in one go like this:

```
mysql [...] icingadb < /usr/share/icingadb/mysql/upgrades/1.1.2.sql
mysql [...] icingadb < /usr/share/icingadb/mysql/upgrades/1.1.2-history.sql
```

If it's only in the upgrading docs, you have to copy & paste, just another step where mistakes can be made.

schema/mysql/schema.sql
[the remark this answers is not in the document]
yhabteab: > So they don't need a new schema version, but this PR does.

Of course they do! What makes them different in any way from this PR?

As I said before, these specific schema version changes should be committed either in a separate commit or even better in a separate PR. This schema change has nothing to do with the actual fix and as such, should better be part of the #707 PR.

lippserd:
> > Edit: I don't like the additional upgrade scripts. Since there's something to be written in the upgrading docs anyway, the statements should be there.
> 
> There and only there? I mean if these exist as a file like all the other ones, one can apply them the same way and in one go like this:
> 
> ```
> mysql [...] icingadb < /usr/share/icingadb/mysql/upgrades/1.1.2.sql
> mysql [...] icingadb < /usr/share/icingadb/mysql/upgrades/1.1.2-history.sql
> ```
> 
> If it's only in the upgrading docs, you have to copy & paste, just another step where mistakes can be made.

We'd have to have full documentation in both places to make absolutely sure people read and understand it, although I'm pretty sure there'll still be people who just import the upgrade script without reading it, because why not. Of course, we could rename the file to `caution-read-before-use`, but that doesn't feel right either. Even if it's not a big problem, but if we ever introduce automatic schema migrations, we need to handle this one special case as well.

doc/04-Upgrading.md
```diff
@@ -3,6 +3,44 @@
...
+Applying `1.1.2.sql` itself will allow you to set hosts/services' `max_check_attempts` above 255.
+But actual soft state changes with check attempt above 255 in your history will still crash Icinga DB v1.1.2.
+To avoid that, you can lower `max_check_attempts`. Optionally increase the respective `retry_interval`
+not to change the timespan from an outage to a hard state (`max_check_attempts x retry_interval`).
+
+If lower `max_check_attempts` are not an option, you'll have to execute the SQL statements below.
+Note that this can take a lot of time depending on the history size and the performance of the database.
+While the upgrade is running, that table will be locked and can't be accessed.
+This means that the existing history can't be viewed in Icinga Web and new history entries will be buffered in Redis.
+Do not interrupt them not to corrupt the state_history table!
```
julianbrost, line 29: This should give a bit of context for the reader. Simply starting with a sentence that earlier versions weren't able to handle `max_check_attempts > 255` should be enough.

> If lower `max_check_attempts` are not an option, you'll have to execute the SQL statements below.

This makes it sound like not applying the optional part of the upgrade was the preferred way. I don't think it is. Instead, it's a trade-off where both options have pros and cons. Thus, it should be presented as such so that users are encouraged to make an informed decision here.

I think it's also worth mentioning that the optional part can be applied at a later time, for example during the next longer maintenance window.

> Do not interrupt them not to corrupt the state_history table!

This won't corrupt the table (unless there's a bug in the database server), it might just result in an expensive rollback that could make everything take much longer than necessary.

julianbrost (changes requested):

julianbrost:
> > > Edit: I don't like the additional upgrade scripts. Since there's something to be written in the upgrading docs anyway, the statements should be there.
> > 
> > 
> > There and only there? I mean if these exist as a file like all the other ones, one can apply them the same way and in one go like this:
> > ```
> > mysql [...] icingadb < /usr/share/icingadb/mysql/upgrades/1.1.2.sql
> > mysql [...] icingadb < /usr/share/icingadb/mysql/upgrades/1.1.2-history.sql
> > ```
> > If it's only in the upgrading docs, you have to copy & paste, just another step where mistakes can be made.
> 
> We'd have to have full documentation in both places to make absolutely sure people read and understand it, although I'm pretty sure there'll still be people who just import the upgrade script without reading it, because why not.

I don't think it's possible to make everyone ignoring the upgrading instructions happy here. If it was, there shouldn't be an optional part. There will probably be users that ignore the upgrading docs, upgrade their setup, and then run into the bug a year later.

> Of course, we could rename the file to `caution-read-before-use`, but that doesn't feel right either.

Sure, the name `1.1.2-history.sql` is pretty non-descriptive in that regard. But shouldn't simply be including "optional" in the name be enough to make people think whether they need it? Maybe even placing it into `schema/mysql/upgrades/optional/1.1.2-history.sql` to make it even more distinct from the other upgrades?

> Even if it's not a big problem, but if we ever introduce automatic schema migrations, we need to handle this one special case as well.

But that special handling would be necessary regardless where we put it now. But if the optional part is only mentioned in the upgrading docs and not where all the other schema upgrades are, I think there's a higher chance that we would simply forget that there was this optional upgrade some time in the past.

lippserd:
> Maybe even placing it into `schema/mysql/upgrades/optional/1.1.2-history.sql` to make it even more distinct from the other upgrades?

That is a good idea.

doc/04-Upgrading.md
julianbrost: I don't agree that this is clearly the less dangerous operation. The trade-off is between performing a somewhat expensive maintenance operation right now, or do I want to put the burden on myself to never ever configure `max_check_attempts > 255` because if I do, it will break. You might not be able to know the exact duration the schema upgrade will take but you know when it will happen, so it's annoying but predictable. Skipping the upgrade and being an innocent-looking config option away from crashing Icinga DB in the future, which I'd call dangerous.

Al2Klimov: Then I opt for presenting the history table upgrade as preferred with the duration disclaimer.

julianbrost: Yes, I think the following key points would appropriately describe the what optional means here:

- History table used too small data type, fixing it is expensive, so we don't do it automatically on upgrade.
- If applying the fix is fine for you, you should do it.
- If not, you have to ensure no `max_check_attempts > 255` now or in the future (if you can't, reconsider the previous step).
- Applying at a later time also works, but until then there must be no `max_check_attempts > 255`.

doc/04-Upgrading.md
```diff
@@ -3,6 +3,49 @@
...
+```mysql
+ALTER TABLE state_history MODIFY COLUMN check_attempt int unsigned NOT NULL;
+```
+
+#### PostgreSQL
+
+```postgresql
+ALTER TABLE state_history ALTER COLUMN check_attempt TYPE uint;
+COMMENT ON COLUMN state_history.check_attempt IS NULL;
+```
```
julianbrost, line 47: If the schema upgrade now exists in a file, why not reference it there?

doc/04-Upgrading.md
```diff
@@ -3,6 +3,49 @@
...
+Databases created with the v1.1.2 schema are not affected, but upgraded ones are.
+The schema upgrade required to fix this isn't included in `1.1.2.sql` as it re-writes the whole `state_history` table.
+This can take a lot of time depending on the history size and the performance of the database.
+During this time that table will be locked exclusively and can't be accessed otherwise.
+This means that the existing history can't be viewed in Icinga Web and new history entries will be buffered in Redis.
+
+That optional table re-write is given below and should be applied during the next longer maintenance window.
+Do not interrupt it! At best use tmux/screen not to loose your SSH session.
+Until this upgrade you'll have to lower `max_check_attempts`. Optionally increase the respective `retry_interval`
+not to change the timespan from an outage to a hard state (`max_check_attempts x retry_interval`).
```
julianbrost, line 34: `x` as a multiplication sign inside a code block looks strange to me. IMHO all of these would fit better:

- `max_check_attempts * retry_interval`
- `max_check_attempts` × `retry_interval` (unicode multiplication sign)
- `max_check_attempts` times `retry_interval`

schema/mysql/upgrades/1.1.2.sql
```diff
@@ -1 +1,10 @@
+ALTER TABLE host_state MODIFY COLUMN check_attempt int unsigned NOT NULL;
+
+ALTER TABLE service_state MODIFY COLUMN check_attempt int unsigned NOT NULL;
+
+ALTER TABLE state_history MODIFY COLUMN check_attempt tinyint unsigned NOT NULL COMMENT 'optional schema upgrade not applied yet, see https://icinga.com/docs/icinga-db/latest/doc/04-Upgrading/#upgrading-to-icinga-db-v112';
```
julianbrost, line 5: Indeed, seems impossible to change the comment without repeating the column type: https://bugs.mysql.com/bug.php?id=64439

Thus, if it was applied in the wrong order, it would actually undo the type change. So the upgrading docs should explicitly state to apply it after `1.1.2.sql`.

doc/04-Upgrading.md
```diff
@@ -3,6 +3,49 @@
...
+to apply this schema upgrade while the Icinga DB v1.1.1 daemon is still running and then restart the daemon with the
+new version. Please keep in mind that depending on the distribution, your package manager may automatically attempt to
+restart the daemon when upgrading the package.
+
+### Upgrading the state_history Table (optional)
+
+Icinga DB 1.1.1 already handles hosts/services' `max_check_attempts` above 255.
+But actual soft state changes with check attempt above 255 in your history will crash Icinga DB,
+even after upgrading to v1.1.2. The `check_attempt` column of the `state_history` table is too small
+to fit larger values. ([#655](https://github.com/Icinga/icingadb/issues/655))
```
julianbrost, line 23: > Icinga DB 1.1.1 already handles hosts/services' `max_check_attempts` above 255.

I'm not even sure what this is trying to tell me here. Why 1.1.1 in particular? Also, I don't think it's relevant here whether some other version would have worked with `max_check_attempts > 255` as long as nothing ever reached that number of check attempts.

> But actual soft state changes with check attempt above 255 in your history will crash Icinga DB, even after upgrading to v1.1.2.

That sounds rather worrisome to me without any extra context given at that point.

Maybe simply start by explaining that previous versions used an insufficient datatype to store the check attempt in the history and that this is fixed for new installations that import the full schema, but it's not automatically fixed by the regular schema upgrade because [insert existing explanation here].

julianbrost (changes requested):

doc/04-Upgrading.md
Al2Klimov: 1. v1.1.1 is obviously the previous version. Also, I can't just say that large max_check_attempts didn't work, because they did. This paragraph precisely describes the issue for informed decision purposes.
2. If we prefer the user to apply the upgrade, we should reason it. A crash is a pretty good reason.

julianbrost: > Also, I can't just say that large max_check_attempts didn't work, because they did.

Well, technically it's not just the `max_check_attempts` value that causes the problem, you can configure a high value there without immediately/certainly running in a problem. But I wouldn't say high `max_check_attempts` are working if Icinga DB crashes if the maximum is actually reached. Also, you could use wording like "checkable reaches a check attempt higher than 255". That would accurately describe it. I don't think "but technically you could configure `max_check_attempts` over 9000 on an always OK check and it would work fine" is necessary information nor does it help. Why would you increase `max_check_attempts` if you don't expect it to reach that value unless you're testing for bugs?

> If we prefer the user to apply the upgrade, we should reason it. A crash is a pretty good reason.

I'm not saying that you shouldn't mention it at all. It's just that if you read that section top to bottom, with the information you were given up to that sentence, you may ask yourself "well, if it still crashes, what does this update even do?".

doc/04-Upgrading.md
```diff
@@ -3,6 +3,35 @@
...
+As the daemon checks the schema version, the recommended way to perform the upgrade is to stop the daemon, apply the
+schema upgrade and then start the new daemon version. If you want to minimize downtime as much as possible, it is safe
+to apply this schema upgrade while the Icinga DB v1.1.1 daemon is still running and then restart the daemon with the
+new version. Please keep in mind that depending on the distribution, your package manager may automatically attempt to
+restart the daemon when upgrading the package.
+
+### Upgrading the state_history Table (optional)
+
+Icinga DB crashes if hosts/services reach check attempt 256. The `check_attempt` column of the `state_history` table
+is too small to fit values greater 255. ([#655](https://github.com/Icinga/icingadb/issues/655))
```
julianbrost, line 21: > It's just that if you read that section top to bottom, with the information you were given up to that sentence, you may ask yourself "well, if it still crashes, what does this update even do?".

-- https://github.com/Icinga/icingadb/pull/656#discussion_r1549488023

Al2Klimov: Your claim sounds better with _if you read **only the 1st 1/3 of** that section_. Yes. There's a crash. And the update doesn't fix it.\* But we have good reasons and wrote them down in the upgrading docs.

\* Actually it does, you "just" have to apply two files.

julianbrost: > Your claim sounds better with _if you read **only the 1st 1/3 of** that section_.

Well yes, people tend to read text from top to bottom. And usually it's a feature if understanding a sentence doesn't depend on information only given later in the text. Otherwise, you end up with a text you have to read multiple times until you understand it.

julianbrost: I've prepared a suggestion for how this can be written in 1298c304cc2adff7c8b13b2f7e732bcda8bbcc67. It's based on the text provided in this PR (hence the `Co-authored-by`) but restructured so that it first explains why the schema upgrade is split and then mentions the options including the "may crash" warning if the upgrade is skipped. Apart from that, I've tweaked the wording in a few places.

Al2Klimov: Good in general, but for max_check_attempts=256 (which already works btw.) the highest possible check_attempt is 255: https://github.com/Icinga/icinga2/blob/9e31b8b5590c6d67b7dd538b2c884bd377a4e486/lib/icinga/checkable-check.cpp#L233-L237

julianbrost (approved):

PR #657, opened 2023-10-10
Flatten(): render even large numbers as-is, not using scientific notation

Al2Klimov:
E.g. 2000000000000000000 (explicitly), not 2e+18 (as with fmt.Sprintf("%v")).

fixes Icinga/icingadb#659

## TODO

* [ ] review (not @Al2Klimov)
* [x] unit test the whole function as already on it (@Al2Klimov)
* [x] transfer Icinga/icingadb#659
* [x] adjust "fixes Icinga/icingadb#659"

### DIFF
--- a/pkg/flatten/flatten.go
+++ b/pkg/flatten/flatten.go
@@ -31,12 +31,10 @@ func Flatten(value interface{}, prefix string) map[string]types.String {
 			for i, v := range value {
 				flatten(key+"["+strconv.Itoa(i)+"]", v)
 			}
+		case nil:
+			flattened[key] = types.MakeString("null")
 		default:
-			val := "null"
-			if value != nil {
-				val = fmt.Sprintf("%v", value)
-			}
-			flattened[key] = types.MakeString(val)
+			flattened[key] = types.MakeString(fmt.Sprintf("%v", value))
 		}
 	}
 
--- a/pkg/flatten/flatten.go
+++ b/pkg/flatten/flatten.go
@@ -33,6 +33,8 @@ func Flatten(value interface{}, prefix string) map[string]types.String {
 			}
 		case nil:
 			flattened[key] = types.MakeString("null")
+		case float64:
+			flattened[key] = types.MakeString(strconv.FormatFloat(value, 'f', -1, 64))
 		default:
 			flattened[key] = types.MakeString(fmt.Sprintf("%v", value))
 		}
--- a/pkg/flatten/flatten.go
+++ b/pkg/flatten/flatten.go
@@ -1,7 +1,6 @@
 package flatten
 
 import (
-	"database/sql"
 	"fmt"
 	"github.com/icinga/icingadb/pkg/types"
 	"strconv"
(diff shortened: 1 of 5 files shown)


pkg/flatten/flatten.go
```diff
@@ -34,9 +34,14 @@ func Flatten(value interface{}, prefix string) map[string]types.String {
...
 		default:
 			val := "null"
-			if value != nil {
+			switch v := value.(type) {
+			case nil:
+			case float64:
+				val = strconv.FormatFloat(v, 'f', -1, 64)
+			default:
 				val = fmt.Sprintf("%v", value)
 			}
```
julianbrost, line 43: This is already inside `switch value := value.(type)`, so shouldn't adding a `case float64:` there suffice?

Al2Klimov: That would duplicate `flattened[key] = ...`.

julianbrost: I see. The annoying part is the right hand side though, I'd consider just adding a helper function instead so that one could just write something like `flattened[key] = types.MakeString("null")`. That might benefit other places in the long run too as those nested struct initialization is just annoying.

julianbrost:
Anyway, the tests don't have to wait for the implementation details. Part of the tests should be to verify that the implementation isn't changed in other ways, so please add them as a first commit so that they pass on the old implementation and then add another case for large floats with the updated implementation.

pkg/flatten/flatten_test.go
```diff
@@ -0,0 +1,33 @@
...
+		{"int", "c", 42, map[string]types.String{"c": types.NewString("42")}},
+		{"float", "d", 77.7, map[string]types.String{"d": types.NewString("77.7")}},
+		{"large_float", "e", 1e23, map[string]types.String{"e": types.NewString("100000000000000000000000")}},
+		{"string", "f", "\x00", map[string]types.String{"f": types.NewString("\x00")}},
+		{"nil_slice", "g", []any(nil), map[string]types.String{"g": {}}},
+		{"empty_slice", "h", []any{}, map[string]types.String{"h": {}}},
+		{"slice", "i", []any{nil}, map[string]types.String{"i[0]": types.NewString("null")}},
+		{"nil_map", "j", map[string]any(nil), map[string]types.String{"j": {}}},
+		{"empty_map", "k", map[string]any{}, map[string]types.String{"k": {}}},
+		{"map", "l", map[string]any{" ": nil}, map[string]types.String{"l. ": types.NewString("null")}},
```
julianbrost, line 27: Please add an complex test case as well With things like values in slice in map and the other way around, i.e. values in map in slice.

pkg/types/string.go
```diff
@@ -15,6 +15,14 @@ type String struct {
 	sql.NullString
 }
 
+// NewString constructs a new non-NULL String from s.
+func NewString(s string) String {
```
julianbrost, line 19: I suggested the name `MakeString` because for `NewString`, I'd kind of expect it to return `*String`, similar to `new(String)` or pretty much any other `New*` function I can think of.

julianbrost (changes requested):
Please reorder the commits so that that the tests are added first without the `1e23` case, so that when the tests are run on this commit, it shows that the old implementation conforms to the tests, and then fixing the implementation including an additional test case, showing that the new implementation passes the same tests as the old one plus a new one.

julianbrost (approved):

PR #665, opened 2023-11-15
MySQL driver: on connect try setting wsrep_sync_wait=7, swallow error 1193

### LINKED ISSUE #577: icingadb process crashes with Error 1452: Cannot add or update a child row: a foreign key constraint fails
The icingadb process crashes at irregular intervals with the following output:

Error 1452: Cannot add or update a child row: a foreign key constraint fails ("icingadb"."history", CONSTRAINT "fk_history_state_history" FOREIGN KEY ("state_history_id") REFERENCES "state_history" ("id") ON DELETE CASCADE)
                                             can't perform "INSERT INTO \"history\" (\"environment_id\", \"endpoint_id\", \"host_id\", \"service_id\", \"event_time\", \"object_type\", \"event_type\", \"id\", \"state_history_id\") VALUES (:environment_id,:endpoint_id,:host_id,:service_id,:event_time,:object_type,:event_type,:id,:state_history_id) ON DUPLICATE KEY UPDATE \"id\" = VALUES(\"id\")"
                                                     github.com/icinga/icingadb/pkg/icingadb/db.go:394
                                                     github.com/icinga/icingadb/pkg/retry/retry.go:45
                                                     github.com/icinga/icingadb/pkg/icingadb/db.go:389
                                                     [golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57](mailto:golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57)
                                             runtime.goexit
                                                     runtime/asm_amd64.s:1594
                                             can't retry
                                                     github.com/icinga/icingadb/pkg/retry/retry.go:64
                                                     github.com/icinga/icingadb/pkg/icingadb/db.go:389
                                                     [golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57](mailto:golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57)
                                             runtime.goexit
                                                     runtime/asm_amd64.s:1594

__

I can't find more detailed log or any posibillity to figure out which entry causes the crash. 

## Your Environment

Icinga DB version: v1.1.0

Build information:
  Go version: go1.19.3 (linux, amd64)
  Git commit: a0093d1df6a227bfc923717072d8cb0e3331799f

System information:
  Platform: Debian GNU/Linux
  Platform version: 11 (bullseye)

Al2Klimov:
In Galera clusters wsrep_sync_wait=7 lets statements catch up all    pending sync between nodes first. This way new child rows await fresh parent    ones from other nodes not to run into foreign key errors. MySQL single nodes    will reject this with error 1193 "Unknown system variable" which is OK.

fixes #577

# To do

* [x] Code structure OK? (@julianbrost)
* [x] GHA pass?
* [x] Before/after tests with actual Galera cluster and single node (@Al2Klimov)
* [x] Test in prod (customer)
* [x] Consider unit testing setGaleraOpts() (@Al2Klimov)

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -85,7 +85,7 @@ func (d Driver) OpenConnector(name string) (driver.Connector, error) {
 
 // Register makes our database Driver available under the name "icingadb-*sql".
 func Register(logger *logging.Logger) {
-	sql.Register(MySQL, &Driver{ctxDriver: &mysql.MySQLDriver{}, Logger: logger})
+	sql.Register(MySQL, &Driver{ctxDriver: MySQLDriver{}, Logger: logger})
 	sql.Register(PostgreSQL, &Driver{ctxDriver: PgSQLDriver{}, Logger: logger})
 	_ = mysql.SetLogger(mysqlLogger(func(v ...interface{}) { logger.Debug(v...) }))
 	sqlx.BindDriver(PostgreSQL, sqlx.DOLLAR)
--- a/pkg/driver/mysql.go
+++ b/pkg/driver/mysql.go
@@ -1,0 +1,100 @@
+package driver
+
+import (
+	"context"
+	"database/sql/driver"
+	"github.com/go-sql-driver/mysql"
+	"github.com/pkg/errors"
+)
+
+// MySQLDriver extends mysql.MySQLDriver with auto-SETting Galera cluster options.
+type MySQLDriver struct {
+	mysql.MySQLDriver
+}
+
+// Open implements the driver.Driver interface.
+func (md MySQLDriver) Open(name string) (driver.Conn, error) {
+	connector, err := md.OpenConnector(name)
+	if err != nil {
+		return nil, err
+	}
+
+	return connector.Connect(context.Background())
+}
+
+// OpenConnector implements the driver.DriverContext interface.
+func (md MySQLDriver) OpenConnector(name string) (driver.Connector, error) {
+	connector, err := md.MySQLDriver.OpenConnector(name)
+	if err != nil {
+		return nil, err
+	}
+
+	return &galeraAwareConnector{connector, md}, nil
+}
+
+// galeraAwareConnector extends mysql.connector with auto-SETting Galera cluster options.
+type galeraAwareConnector struct {
+	driver.Connector
+
+	driver driver.Driver
+}
+
+// Connect implements the driver.Connector interface.
+func (gac *galeraAwareConnector) Connect(ctx context.Context) (driver.Conn, error) {
+	conn, err := gac.Connector.Connect(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := setGaleraOpts(ctx, conn); err != nil {
+		_ = conn.Close()
+		return nil, err
+	}
+
+	return conn, nil
+}
+
+// Driver implements the driver.Connector interface.
+func (gac *galeraAwareConnector) Driver() driver.Driver {
+	return gac.driver
+}
+
+var errUnknownSysVar = &mysql.MySQLError{Number: 1193}
+
+// setGaleraOpts tries SET SESSION wsrep_sync_wait=7.
+//
+// This ensures causality checks will take place before executing anything,
+// ensuring that every statement is executed on a fully synced node.
+// https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait
+//
+// It prevents running into foreign key errors while inserting into linked tables on different MySQL nodes.
+// Error 1193 "Unknown system variable" is ignored to support MySQL single nodes.
+func setGaleraOpts(ctx context.Context, conn driver.Conn) error {
+	const galeraOpts = "SET SESSION wsrep_sync_wait=7"
+
+	stmt, err := conn.(driver.ConnPrepareContext).PrepareContext(ctx, galeraOpts)
+	if err != nil {
+		err = errors.Wrap(err, "can't prepare "+galeraOpts)
+	} else if _, err = stmt.(driver.StmtExecContext).ExecContext(ctx, nil); err != nil {
+		err = errors.Wrap(err, "can't execute "+galeraOpts)
+	}
+
+	if err != nil && errors.Is(err, errUnknownSysVar) {
+		err = nil
+	}
+
+	if stmt != nil {
+		if errClose := stmt.Close(); errClose != nil && err == nil {
+			err = errors.Wrap(errClose, "can't close statement "+galeraOpts)
+		}
+	}
+
+	return err
+}
+
+// Assert interface compliance.
+var (
+	_ driver.Driver        = MySQLDriver{}
+	_ driver.DriverContext = MySQLDriver{}
+	_ driver.Connector     = (*galeraAwareConnector)(nil)
+)
--- a/pkg/driver/mysql.go
+++ b/pkg/driver/mysql.go
@@ -5,6 +5,7 @@ import (
 	"database/sql/driver"
 	"github.com/go-sql-driver/mysql"
 	"github.com/pkg/errors"
+	"strconv"
 )
 
 // MySQLDriver extends mysql.MySQLDriver with auto-SETting Galera cluster options.
@@ -24,19 +25,34 @@ func (md MySQLDriver) Open(name string) (driver.Conn, error) {
 
 // OpenConnector implements the driver.DriverContext interface.
 func (md MySQLDriver) OpenConnector(name string) (driver.Connector, error) {
+	var wsrepSyncWait int64
+
+	if config, err := mysql.ParseDSN(name); err == nil {
+		if s, ok := config.Params["wsrep_sync_wait"]; ok {
+			if i, err := strconv.ParseInt(s, 10, 64); err == nil {
+				// MySQL single nodes don't know wsrep_sync_wait and fail with error 1193 "Unknown system variable".
+				// We have to SET it manually later and swallow error 1193 not to fail our connections.
+				wsrepSyncWait = i
+				delete(config.Params, "wsrep_sync_wait")
+				name = config.FormatDSN()
+			}
+		}
+	}
+
 	connector, err := md.MySQLDriver.OpenConnector(name)
 	if err != nil {
 		return nil, err
 	}
 
-	return &galeraAwareConnector{connector, md}, nil
+	return &galeraAwareConnector{connector, wsrepSyncWait, md}, nil
 }
 
 // galeraAwareConnector extends mysql.connector with auto-SETting Galera cluster options.
 type galeraAwareConnector struct {
 	driver.Connector
 
-	driver driver.Driver
+	wsrepSyncWait int64
+	driver        driver.Driver
 }
 
 // Connect implements the driver.Connector interface.
@@ -46,7 +62,7 @@ func (gac *galeraAwareConnector) Connect(ctx context.Context) (driver.Conn, erro
 		return nil, err
 	}
 
-	if err := setGaleraOpts(ctx, conn); err != nil {
+	if err := setGaleraOpts(ctx, conn, gac.wsrepSyncWait); err != nil {
 		_ = conn.Close()
 		return nil, err
 	}
@@ -61,22 +77,25 @@ func (gac *galeraAwareConnector) Driver() driver.Driver {
 
 var errUnknownSysVar = &mysql.MySQLError{Number: 1193}
 
-// setGaleraOpts tries SET SESSION wsrep_sync_wait=7.
+// setGaleraOpts tries SET SESSION wsrep_sync_wait.
 //
-// This ensures causality checks will take place before executing anything,
+// This ensures causality checks will take place before execution,
 // ensuring that every statement is executed on a fully synced node.
 // https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait
 //
 // It prevents running into foreign key errors while inserting into linked tables on different MySQL nodes.
 // Error 1193 "Unknown system variable" is ignored to support MySQL single nodes.
-func setGaleraOpts(ctx context.Context, conn driver.Conn) error {
-	const galeraOpts = "SET SESSION wsrep_sync_wait=7"
+func setGaleraOpts(ctx context.Context, conn driver.Conn, wsrepSyncWait int64) error {
+	const galeraOpts = "SET SESSION wsrep_sync_wait=?"
 
 	stmt, err := conn.(driver.ConnPrepareContext).PrepareContext(ctx, galeraOpts)
 	if err != nil {
 		err = errors.Wrap(err, "can't prepare "+galeraOpts)
-	} else if _, err = stmt.(driver.StmtExecContext).ExecContext(ctx, nil); err != nil {
-		err = errors.Wrap(err, "can't execute "+galeraOpts)
+	} else {
+		_, err = stmt.(driver.StmtExecContext).ExecContext(ctx, []driver.NamedValue{{Value: wsrepSyncWait}})
+		if err != nil {
+			err = errors.Wrap(err, "can't execute "+galeraOpts)
+		}
 	}
 
 	if err != nil && errors.Is(err, errUnknownSysVar) {
--- a/pkg/icingadb/db.go
+++ b/pkg/icingadb/db.go
@@ -54,6 +54,10 @@ type Options struct {
 	// MaxRowsPerTransaction defines the maximum number of rows per transaction.
 	// The default is 2^13, which in our tests showed the best performance in terms of execution time and parallelism.
 	MaxRowsPerTransaction int `yaml:"max_rows_per_transaction" default:"8192"`
+
+	// WsrepSyncWait defines which kinds of SQL statements catch up all pending sync between nodes first, see:
+	// https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait
+	WsrepSyncWait int `yaml:"wsrep_sync_wait" default:"7"`
 }
 
 // Validate checks constraints in the supplied database options and returns an error if they are violated.
--- a/pkg/config/database.go
+++ b/pkg/config/database.go
@@ -61,7 +61,11 @@ func (d *Database) Open(logger *logging.Logger) (*icingadb.DB, error) {
 
 		config.DBName = d.Database
 		config.Timeout = time.Minute
-		config.Params = map[string]string{"sql_mode": "ANSI_QUOTES"}
+
+		config.Params = map[string]string{
+			"sql_mode":        "ANSI_QUOTES",
+			"wsrep_sync_wait": strconv.FormatInt(int64(d.Options.WsrepSyncWait), 10),
+		}
 
 		tlsConfig, err := d.TlsOptions.MakeConfig(d.Host)
 		if err != nil {
--- a/pkg/driver/mysql_test.go
+++ b/pkg/driver/mysql_test.go
@@ -96,7 +96,7 @@ func TestSetGaleraOpts(t *testing.T) {
 
 	for _, st := range subtests {
 		t.Run(st.name, func(t *testing.T) {
-			assert.ErrorIs(t, setGaleraOpts(context.Background(), &st.input), st.output)
+			assert.ErrorIs(t, setGaleraOpts(context.Background(), &st.input, 7), st.output)
 			assert.GreaterOrEqual(t, st.input.prepareCalls, uint8(1))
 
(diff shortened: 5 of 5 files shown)


Al2Klimov:
# Single node test

https://github.com/Al2Klimov/twintowers with this patch:

```patch
--- docker-compose.yml
+++ docker-compose.yml
@@ -25,12 +25,12 @@ x-ido: &x-ido
     MARIADB_DATABASE: ido

 x-dbicinga: &x-dbicinga
-  image: mariadb:10
+  image: mysql:5.7
   environment:
-    MARIADB_RANDOM_ROOT_PASSWORD: '1'
-    MARIADB_USER: icingadb
-    MARIADB_PASSWORD: icingadb
-    MARIADB_DATABASE: icingadb
+    MYSQL_RANDOM_ROOT_PASSWORD: '1'
+    MYSQL_USER: icingadb
+    MYSQL_PASSWORD: icingadb
+    MYSQL_DATABASE: icingadb

 x-icingadb-env: &x-icingadb-env
   ICINGADB_REDIS_PORT: '6379'
@@ -186,7 +186,7 @@ services:
         condition: service_started
       dbicinga1:
         condition: service_started
-    image: icinga/icingadb
+    image: icinga/icingadb:test
     environment:
       <<: *x-icingadb-env
       ICINGADB_REDIS_HOST: redis1
@@ -279,7 +279,7 @@ services:
         condition: service_started
       dbicinga2:
         condition: service_started
-    image: icinga/icingadb
+    image: icinga/icingadb:test
     environment:
       <<: *x-icingadb-env
       ICINGADB_REDIS_HOST: redis2
```

The MySQL image throws 1193 on the command line with SET SESSION wsrep_sync_wait=4.
:+1: But Icinga DB still works.

julianbrost:
What effect does changing `mariadb:10` to `mysql:5.7`? I'd expect both to behave identically. Is that what you wanted to test here, that old MySQL behaves like somewhat recent MariaDB?

Al2Klimov:
> The MySQL image throws 1193 on the command line with SET SESSION wsrep_sync_wait=4.

In contrast to the MariaDB **Docker image** it doesn't even ship cluster functionality and rejects the var. The perfect single node for testing IMAO.

Al2Klimov:
# Galera test

Cluster as per https://github.com/Icinga/icingadb/issues/577#issuecomment-1682443163 verified with https://github.com/Al2Klimov/provoke-galera (that #577-ish errors can happen under laboratory conditions). In addition there's a Debian 12 load balancer:

```
root@aklimov-lb:~# iptables -t nat -A PREROUTING -d 10.27.2.59 -p tcp --dport 3306 -m statistic --mode random --probability 0.5 -j DNAT --to-destination 10.27.3.229:3306
root@aklimov-lb:~# iptables -t nat -A PREROUTING -d 10.27.2.59 -p tcp --dport 3306 -j DNAT --to-destination 10.27.1.199:3306
root@aklimov-lb:~# iptables -t nat -A POSTROUTING -o ens3 -j MASQUERADE
root@aklimov-lb:~# sysctl net.ipv4.ip_forward=1
root@aklimov-lb:~#
```

With this LB in https://github.com/Al2Klimov/twintowers

```patch
--- docker-compose.yml
+++ docker-compose.yml
@@ -190,7 +190,7 @@ services:
     environment:
       <<: *x-icingadb-env
       ICINGADB_REDIS_HOST: redis1
-      ICINGADB_DATABASE_HOST: dbicinga1
+      ICINGADB_DATABASE_HOST: 10.27.2.59

   web1:
     depends_on:
@@ -209,7 +209,7 @@ services:
       icingaweb.modules.icingadb.redis.redis1.host: redis1
       icingaweb.modules.monitoring.commandtransports.icinga2.host: master1
       icingaweb.resources.icingaweb_db.host: dbweb1
-      icingaweb.resources.icingadb.host: dbicinga1
+      icingaweb.resources.icingadb.host: 10.27.2.59
       icingaweb.resources.icinga_ido.host: ido1
     volumes:
       - ./volumes/icingaweb2/1:/data
```

and a bunch of history

```patch
--- icinga2.conf
+++ icinga2.conf
@@ -44,3 +44,17 @@ object IdoMysqlConnection "ido-mysql" {
     database = "ido"
     enable_ha = false
 }
+
+object Host "ok" {
+    check_command = "dummy"
+}
+
+for (i in range(10000)) {
+object Service i {
+    host_name = "ok"
+    check_command = "random"
+    max_check_attempts = 1
+    check_interval = 1ms
+    retry_interval = 1ms
+}
+}
```

bad things happen:

```
twintowers-icingadb1-1      | 2023-11-15T16:05:34.004Z	FATAL	icingadb	Error 1452 (23000): Cannot add or update a child row: a foreign key constraint fails ("icingadb"."history", CONSTRAINT "fk_history_state_history" FOREIGN KEY ("state_history_id") REFERENCES "state_history" ("id") ON DELETE CASCADE)
twintowers-icingadb1-1      | can't perform "INSERT INTO \"history\" (\"host_id\", \"event_type\", \"state_history_id\", \"event_time\", \"environment_id\", \"object_type\", \"endpoint_id\", \"service_id\", \"id\") VALUES (:host_id,:event_type,:state_history_id,:event_time,:environment_id,:object_type,:endpoint_id,:service_id,:id) ON DUPLICATE KEY UPDATE \"id\" = VALUES(\"id\")"
twintowers-icingadb1-1      | github.com/icinga/icingadb/internal.CantPerformQuery
twintowers-icingadb1-1      | 	/icingadb-src/internal/internal.go:30
twintowers-icingadb1-1      | github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.func2.1
twintowers-icingadb1-1      | 	/icingadb-src/pkg/icingadb/db.go:391
twintowers-icingadb1-1      | github.com/icinga/icingadb/pkg/retry.WithBackoff
twintowers-icingadb1-1      | 	/icingadb-src/pkg/retry/retry.go:49
twintowers-icingadb1-1      | github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.func2
twintowers-icingadb1-1      | 	/icingadb-src/pkg/icingadb/db.go:386
twintowers-icingadb1-1      | golang.org/x/sync/errgroup.(*Group).Go.func1
twintowers-icingadb1-1      | 	/go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75
twintowers-icingadb1-1      | runtime.goexit
twintowers-icingadb1-1      | 	/usr/local/go/src/runtime/asm_amd64.s:1650
twintowers-icingadb1-1      | can't retry
twintowers-icingadb1-1      | github.com/icinga/icingadb/pkg/retry.WithBackoff
twintowers-icingadb1-1      | 	/icingadb-src/pkg/retry/retry.go:68
twintowers-icingadb1-1      | github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.func2
twintowers-icingadb1-1      | 	/icingadb-src/pkg/icingadb/db.go:386
twintowers-icingadb1-1      | golang.org/x/sync/errgroup.(*Group).Go.func1
twintowers-icingadb1-1      | 	/go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75
twintowers-icingadb1-1      | runtime.goexit
twintowers-icingadb1-1      | 	/usr/local/go/src/runtime/asm_amd64.s:1650
twintowers-icingadb1-1 exited with code 1
```

But not with this PR! :+1:

```patch
--- docker-compose.yml
+++ docker-compose.yml
@@ -186,7 +186,7 @@ services:
         condition: service_started
       dbicinga1:
         condition: service_started
-    image: icinga/icingadb
+    image: icinga/icingadb:test
     environment:
       <<: *x-icingadb-env
       ICINGADB_REDIS_HOST: redis1
@@ -279,7 +279,7 @@ services:
         condition: service_started
       dbicinga2:
         condition: service_started
-    image: icinga/icingadb
+    image: icinga/icingadb:test
     environment:
       <<: *x-icingadb-env
       ICINGADB_REDIS_HOST: redis2
```

pkg/driver/driver.go
```diff
@@ -85,7 +85,7 @@ func (d Driver) OpenConnector(name string) (driver.Connector, error) {
 
 // Register makes our database Driver available under the name "icingadb-*sql".
 func Register(logger *logging.Logger) {
-	sql.Register(MySQL, &Driver{ctxDriver: &mysql.MySQLDriver{}, Logger: logger})
+	sql.Register(MySQL, &Driver{ctxDriver: &mySQLDriver{}, Logger: logger})
```
julianbrost, line 88: Wait, we are already wrapping the driver. Why do you wrap it once more instead of just initializing the connection there?

Al2Klimov: Sure, we have a common _Driver_, but my stuff is only for MySQL.

Al2Klimov: Done.

pkg/driver/mysql.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+	"database/sql/driver"
+	"github.com/go-sql-driver/mysql"
+	"github.com/pkg/errors"
+)
+
+var errUnknownSysVar = &mysql.MySQLError{Number: 1193}
+
+// setGaleraOpts tries SET SESSION wsrep_sync_wait=4.
+// Error 1193 "Unknown system variable" is ignored to support MySQL single nodes.
```
julianbrost, line 13: This comment would be more helpful if it also said why.

julianbrost (commented):
> In Galera clusters wsrep_sync_wait=4 ensures inserted rows to be synced over all nodes before reporting success to their inserter.

I don't think that's what this variable does, the [documentation](https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait) says:

> Setting this variable ensures causality checks will take place **before executing an operation of the type specified by the value**, ensuring that the statement is executed on a fully synced node.

Which I read as this waits for everything to sync onto the local node before executing the query, not waiting for the result of the query to sync to all other nodes before reporting success.

> > The MySQL image throws 1193 on the command line with SET SESSION wsrep_sync_wait=4.
> 
> In contrast to the MariaDB **Docker image** it doesn't even ship cluster functionality and rejects the var. The perfect single node for testing IMAO.

So a non-Galera-cluster MariaDB understands the variable but treats it as a no-op?

FWIW: no testing from my side so far, but looks like it should allow doing what we want to do: testing the effect of `wsrep_sync_wait=4`.

Al2Klimov:
> So a non-Galera-cluster MariaDB understands the variable but treats it as a no-op?

Yes, our tests confirm this.

oxzi (commented):
This might not be the perfect place for those questions, but I failed finding them being addressed somewhere else. If it was already discussed, please point me there and excuse this comment.

1. With no intention to sound rude, but isn't setting `wsrep_sync_wait` more a hack than a solution? The core problem, a foreign key violation due to non-linear or non-synchronized `INSERT`, is not really mitigated by doing so.
How about putting the `state_history` and `history` `UPSERT` commands into one transaction? Of course, this would require some refactoring of `history.Sync`..
2. Why should mode 4 (_INSERT and REPLACE_) be used? What about `UPDATE`, which is, e.g., covered by mode 6 (_UPDATE, DELETE, INSERT and REPLACE_)? Furthermore, are `READ` operations now not ensured to be executed on a fully synced node?
Can we be sure that there are no other races when using a "master-master" replication or a cluster instead of a single node?
3. What about increasing `wsrep_retry_autocommit`, even if this also seems a bit desperate? Or might this only increase the chance of a deadlock?

Please excuse my rambling above. I hope it's still understandable.

Furthermore, I have added two more trivial code comments below.

julianbrost:
> This might not be the perfect place for those questions, but I failed finding them being addressed somewhere else. If it was already discussed, please point me there and excuse this comment.

There's the history of the referenced issue starting at https://github.com/Icinga/icingadb/issues/577#issuecomment-1669636845 where `wsrep_sync_wait` was proposed for the first time.

> 1. With no intention to sound rude, but isn't setting `wsrep_sync_wait` more a hack than a solution? The core problem, a foreign key violation due to non-linear or non-synchronized `INSERT`, is not really mitigated by doing so.

Our code should synchronize them insofar that it will only issue the second insert after it has received the result for the insert the second row depends on. For a single server, this should be good enough, even if it happens on different connections. However, in Galera cluster, if both queries are issues on connections to different nodes, the result of the first query might not have propagated yet. `wsrep_sync_wait` is supposed to work around this by forcing the server to wait to be synchronized (probably comes at some cost, the documentation hints this and it also seems plausible, but at least I have no idea how big of an impact this is).

>    How about putting the `state_history` and `history` `UPSERT` commands into one transaction? Of course, this would require some refactoring of `history.Sync`..

Shouldn't require a transaction, simply using the same connection should suffice. But maybe using a transaction would provide the nicer API on the Go side, but then we should also see if this has a performance impact (I think we observed some non-obvious performance penalties from explicit transactions over auto-commit in the past, maybe @lippserd remembers some details?).

As you might deduce from #577, I'm not yet convinced that `wsrep_sync_wait` is the best way to address this problem, I'm still open to other approaches.

pkg/driver/mysql.go
```diff
@@ -0,0 +1,41 @@
...
+	"database/sql/driver"
+	"github.com/go-sql-driver/mysql"
+	"github.com/pkg/errors"
+)
+
+var errUnknownSysVar = &mysql.MySQLError{Number: 1193}
+
+// setGaleraOpts tries SET SESSION wsrep_sync_wait=7.
+//
+// This ensures causality checks will take place before executing anything,
```
oxzi, line 14: Please excuse me being pedantic, but `7` is anything _except_ `SHOW`, which we shouldn't be using here.

pkg/icingadb/db.go
```diff
@@ -54,6 +54,10 @@ type Options struct {
 	// MaxRowsPerTransaction defines the maximum number of rows per transaction.
 	// The default is 2^13, which in our tests showed the best performance in terms of execution time and parallelism.
 	MaxRowsPerTransaction int `yaml:"max_rows_per_transaction" default:"8192"`
+
+	// WsrepSyncWait defines which kinds of SQL statements catch up all pending sync between nodes first, see:
+	// https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait
+	WsrepSyncWait int `yaml:"wsrep_sync_wait" default:"7"`
```
oxzi, line 60: This should also be documented in a place that users can easily find. The current example config does not contain an `options` block. This could perhaps be added, as for many people a look at the default config is a starting point.

Textual documentation would also be helpful. Perhaps also referring to the HA area?

pkg/driver/mysql.go
```diff
@@ -0,0 +1,117 @@
...
+
+// OpenConnector implements the driver.DriverContext interface.
+func (md MySQLDriver) OpenConnector(name string) (driver.Connector, error) {
+	var wsrepSyncWait int64
+
+	if config, err := mysql.ParseDSN(name); err == nil {
+		if s, ok := config.Params["wsrep_sync_wait"]; ok {
+			if i, err := strconv.ParseInt(s, 10, 64); err == nil {
+				wsrepSyncWait = i
+				delete(config.Params, "wsrep_sync_wait")
```
oxzi, line 34: Storing `wsrep_sync_wait` inside the `config.Params` feels a bit like a cheat, especially as it will be deleted at this point. Maybe add at least a short comment explaining this?

pkg/driver/mysql.go
```diff
@@ -0,0 +1,41 @@
+package driver
+
+import (
+	"context"
+	"database/sql/driver"
+	"github.com/go-sql-driver/mysql"
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
oxzi: When doing so, the `errors.Wrap` calls below should be replaced by `fmt.Errorf("wrap wrap %w", ...)`.

oxzi (commented):
I have successfully tested the current state on top of my three MariaDB node Galera testing cluster. Therefore, I have also altered the `database.options.wsrep_sync_wait` value.

pkg/icingadb/db.go
Al2Klimov: > The current example config does not contain an `options` block.

That's exactly my reason for not documenting it (yet).

pkg/driver/mysql.go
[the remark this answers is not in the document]
Al2Klimov: I'm just complying to the existing codebase. I don't consider it a good idea to mix different ways of doing the same thing.

pkg/icingadb/db.go
julianbrost: The reasoning for the existing options was that those those should not need to be changed by users. The default values should work fine everywhere, they were merely exposed to the config so that we might try to tweak them even in some production environment if problems would arise (but so far, those defaults seem to have worked out fine).

pkg/driver/mysql.go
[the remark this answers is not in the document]
julianbrost: One part of that package we use is that it also records a stack trace when wrapping errors. So it's not really a drop-in replacement.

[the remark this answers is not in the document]
oxzi: A short `grep` only showed one occurrence of `errors.WithStack` within [`RuntimeUpdates`](https://github.com/Icinga/icingadb/blob/cd96f0de6f33b6750ba44d3934d4380a62ab14a6/pkg/icingadb/runtime_updates.go#L282). Thus, this might easily get replaced. But getting into a discussion about this feels like bikeshedding.

I just wanted to notice that the package in question is now over four years old and was marked dead in late 2021 as _almost_ all functionality is now part of the Go standard library.

And even as lots of other places in the code are still relying on it, imho, we do not have to introduce more usage in new code.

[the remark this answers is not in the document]
julianbrost: `github.com/pkg/errors.Wrap()` also records a stack trace: https://pkg.go.dev/github.com/pkg/errors#Wrap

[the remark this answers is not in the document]
oxzi: Never mind then, sorry.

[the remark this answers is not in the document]
lippserd: How does the stdlib provide stack traces like `errors.Wrap()`? It doesn't. `pkg/errors` is fine and replacing it with the sole argument that it is dead makes no sense as it just works and has benefits over the stdlib. Otherwise we would also have to replace `sqlx`.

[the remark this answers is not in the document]
lippserd: Sorry I'm late to the party, but I had written this comment anyway and wanted to add my two cents.

pkg/driver/mysql.go
```diff
@@ -0,0 +1,119 @@
...
+	if config, err := mysql.ParseDSN(name); err == nil {
+		if s, ok := config.Params["wsrep_sync_wait"]; ok {
+			if i, err := strconv.ParseInt(s, 10, 64); err == nil {
+				// MySQL single nodes don't know wsrep_sync_wait and fail with error 1193 "Unknown system variable".
+				// We have to SET it manually later and swallow error 1193 not to fail our connections.
+				wsrepSyncWait = i
+				delete(config.Params, "wsrep_sync_wait")
+				name = config.FormatDSN()
+			}
+		}
```
yhabteab, line 39: Since we recently bumped the [`mysql`](https://github.com/Icinga/icingadb/pull/687) driver to version `1.8.0`, how about registering a before connect callback like this instead?

```go
mysql.BeforeConnect(func(ctx context.Context, config *mysql.Config) error {
    // Do some stuff with "wsrep_sync_wait"
})
```

That way we wouldn't have to parse the DSN twice for each connection.

julianbrost, line 39: It should actually be possible to completely avoid the DSN for connection to a database by chaining these functions:

- https://pkg.go.dev/github.com/go-sql-driver/mysql#NewConnector
- https://pkg.go.dev/database/sql#OpenDB
- https://pkg.go.dev/github.com/jmoiron/sqlx#NewDb

This should then allow simply passing the config option as a struct member instead.

(And maybe this could also allow us to avoid having to register the custom `"icingadb-mysql"` driver at all.)

yhabteab: This doesn't make any difference to the current version.

Al2Klimov: Why just MySQL? Our Pg driver has a NewConnector() as well.

And what's the problem in general? Describing a connection to a database is the DSN's _purpose_.

PR #679, opened 2024-02-13
retry.Retryable(): re-try MySQL error 1047 for a "not yet prepared node"

Al2Klimov:
During start a Galera cluster complains with "WSREP has not yet prepared node for application use". But the error code, 1047, stands for "Unknown command". So re-try only a "not yet prepared node", but not an "Unknown command".

ref/IP/48748

@NavidSassan Do you want packages with status quo plus this one?

### DIFF
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -150,6 +150,8 @@ func Retryable(err error) bool {
 			// 1213: Deadlock found when trying to get lock
 			// 2006: MySQL server has gone away
 			return true
+		case 1047: // Unknown command
+			return strings.Contains(e.Message, "not yet prepared node for application use")
 		default:
 			return false
 		}


NavidSassan:
yes that would be great, thanks!

pkg/retry/retry.go
```diff
@@ -150,6 +150,8 @@ func Retryable(err error) bool {
 			// 1213: Deadlock found when trying to get lock
 			// 2006: MySQL server has gone away
 			return true
+		case 1047: // Unknown command
+			return strings.Contains(e.Message, "not yet prepared node")
```
oxzi, line 154: Why not match against the fully known message as stated in the PR's initial text?

Furthermore, maybe add a comment to explain this line as it is not self-explaining.

Al2Klimov:
@NavidSassan You can download the Job artifacts here: https://git.icinga.com/packages/icingadb/-/jobs/481689

That includes #665 as well.

PR #684, opened 2024-03-08
Remove redundant closure

Al2Klimov:
which just wraps a method with equal signature.


lippserd:
@yhabteab Please consider this change while working on #711. Feel free to just merge it.

PR #686, opened 2024-03-11
Add a correct composite `INDEX` for the history table

### LINKED ISSUE #683: Missing index on `history` table
## Describe the bug

https://github.com/Icinga/icingadb-web/pull/755 introduced an additional filtering of the histories by `event_type` in Icinga DB Web, but didn't add this column to the existing `event_time` index of the `history` table. As a consequence, loading the history view of Icinga DB Web runs into a timeout on MySQL/MariaDB and never finishes loading with ~1-3 million entries.

ref/IP/48748

yhabteab:
No complaints with PostgreSQL so far, but to maintain the consistency with MySQL/MariaDB, the index for pgsql has also been adjusted. The output of the explain command also shows the differences between before and after this change.

### PostgreSQL (EXPLAIN)

**Before:**
```sql
icingadb=# EXPLAIN SELECT * FROM history ORDER BY event_time, event_type LIMIT 25;
                                              QUERY PLAN                                               
-------------------------------------------------------------------------------------------------------
 Limit  (cost=0.39..7.36 rows=25 width=368)
   ->  Incremental Sort  (cost=0.39..56.14 rows=200 width=368)
         Sort Key: event_time, event_type
         Presorted Key: event_time
         ->  Index Scan using idx_history_event_time on history  (cost=0.14..47.15 rows=200 width=368)
(5 rows)
```

**After:**
```sql
icingadb=# EXPLAIN SELECT * FROM history ORDER BY event_time, event_type LIMIT 25;
                                                 QUERY PLAN                                                 
------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.14..6.02 rows=25 width=368)
   ->  Index Scan using idx_history_event_time_event_type on history  (cost=0.14..47.15 rows=200 width=368)
(2 rows)
```

fixes #683


julianbrost:
> ```sql
>    ->  Incremental Sort  (cost=0.39..56.14 rows=200 width=368)
> ```

That optimization was only added [PostgreSQL 13](https://www.postgresql.org/docs/release/13.0/) though, so the changed index might even be useful for older versions.

oxzi (approved):
I have successfully tested this PR on a local setup with round about 300k entries in the history table. LGTM.

### Times with the old schema
```
MariaDB [icingadb]> SELECT 1 FROM history ORDER BY event_time LIMIT 1\G
1: 1
1 row in set (0.001 sec)

MariaDB [icingadb]> SELECT 1 FROM history ORDER BY event_time, event_type LIMIT 1\G
1: 1
1 row in set (0.260 sec)
```

### Performing schema migration
```
$ time mariadb  […] icingadb < /tmp/schema-1.1.2.sql

real    0m3.122s
user    0m0.013s
sys     0m0.004s
```

### Times with the new schema
```
MariaDB [icingadb]> SELECT 1 FROM history ORDER BY event_time LIMIT 1\G
1: 1
1 row in set (0.001 sec)

MariaDB [icingadb]> SELECT 1 FROM history ORDER BY event_time, event_type LIMIT 1\G
1: 1
1 row in set (0.001 sec)

MariaDB [icingadb]> SELECT COUNT(*) FROM history\G
COUNT(*): 310850
1 row in set (0.169 sec)
```

julianbrost:
I noticed some strange, possibly noteworthy behavior, when testing the schema upgrade on MariaDB (11.0.2). Before I started to apply the schema change, I tried to access `/icingadb/history` in Icinga Web (as that's the view this index is supposed to improve). As my history table is somewhat big (21M rows), this didn't load quickly and I just left the tab open in the background. Then I sent the schema upgrade to the server, waited for a bit and wondered how long it would take.

I looked at `show processlist` and noticed something strange: there's a connections currently executing a `SELECT` statement in state "Creating sort index" whereas the `ALTER TABLE ... ADD INDEX` connection as well as many `INSERT` connections are in state "Waiting for table metadata lock":

```
MariaDB [icingadb]> show processlist;
+------+----------+-------------------+----------+---------+------+---------------------------------+------------------------------------------------------------------------------------------------------+----------+
| Id   | User     | Host              | db       | Command | Time | State                           | Info                                                                                                 | Progress |
+------+----------+-------------------+----------+---------+------+---------------------------------+------------------------------------------------------------------------------------------------------+----------+
|   12 | icingadb | 172.22.0.36:45694 | icingadb | Sleep   |    0 |                                 | NULL                                                                                                 |    0.000 |
| 1444 | icingadb | 172.22.0.36:41460 | icingadb | Prepare | 1215 | Waiting for table metadata lock | INSERT INTO "notification_history" ("object_type", "host_id", "send_time", "author", "text", "endpoi |    0.000 |
| 1458 | icingadb | 172.22.0.36:58514 | icingadb | Prepare | 1219 | Waiting for table metadata lock | INSERT INTO "state_history" ("endpoint_id", "id", "hard_state", "previous_hard_state", "max_check_at |    0.000 |
| 1461 | icingadb | 172.22.0.36:58542 | icingadb | Prepare | 1221 | Waiting for table metadata lock | INSERT INTO "state_history" ("endpoint_id", "id", "hard_state", "previous_hard_state", "max_check_at |    0.000 |
| 1465 | icingadb | 172.22.0.35:37240 | icingadb | Sleep   |    0 |                                 | NULL                                                                                                 |    0.000 |
| 1466 | icingadb | 172.22.0.35:37246 | icingadb | Prepare | 1217 | Waiting for table metadata lock | INSERT INTO "notification_history" ("previous_hard_state", "text", "object_type", "notification_id", |    0.000 |
| 1467 | icingadb | 172.22.0.35:37254 | icingadb | Prepare | 1222 | Waiting for table metadata lock | INSERT INTO "notification_history" ("previous_hard_state", "text", "object_type", "notification_id", |    0.000 |
| 1468 | icingadb | 172.22.0.36:59864 | icingadb | Sleep   |   14 |                                 | NULL                                                                                                 |    0.000 |
| 1469 | icingadb | 172.22.0.35:52672 | icingadb | Prepare | 1222 | Waiting for table metadata lock | INSERT INTO "state_history" ("soft_state", "previous_soft_state", "check_attempt", "service_id", "st |    0.000 |
| 1470 | icingadb | 172.22.0.36:46530 | icingadb | Prepare | 1217 | Waiting for table metadata lock | INSERT INTO "notification_history" ("object_type", "host_id", "send_time", "author", "text", "endpoi |    0.000 |
| 1474 | icingadb | 172.22.0.35:52686 | icingadb | Prepare | 1215 | Waiting for table metadata lock | INSERT INTO "state_history" ("soft_state", "previous_soft_state", "check_attempt", "service_id", "st |    0.000 |
| 1475 | icingadb | 172.22.0.36:36368 | icingadb | Prepare | 1222 | Waiting for table metadata lock | INSERT INTO "notification_history" ("object_type", "host_id", "send_time", "author", "text", "endpoi |    0.000 |
| 1477 | icingadb | 172.22.0.35:33222 | icingadb | Sleep   |    8 |                                 | NULL                                                                                                 |    0.000 |
| 1485 | icingadb | 172.22.0.37:44446 | icingadb | Execute | 1262 | Creating sort index             | SELECT history.id, history.environment_id, history.endpoint_id, history.object_type, history.host_id |    0.000 |
| 1494 | icingadb | localhost         | icingadb | Query   | 1224 | Waiting for table metadata lock | ALTER TABLE history ADD INDEX idx_history_event_time_event_type (event_time, event_type) COMMENT 'Hi |    0.000 |
| 1499 | icingadb | 172.22.0.35:60688 | icingadb | Prepare | 1215 | Waiting for table metadata lock | INSERT INTO "notification_history" ("previous_hard_state", "text", "object_type", "notification_id", |    0.000 |
| 1500 | icingadb | 172.22.0.35:60704 | icingadb | Prepare | 1215 | Waiting for table metadata lock | INSERT INTO "state_history" ("soft_state", "previous_soft_state", "check_attempt", "service_id", "st |    0.000 |
| 1501 | icingadb | 172.22.0.35:60706 | icingadb | Prepare | 1214 | Waiting for table metadata lock | INSERT INTO "state_history" ("soft_state", "previous_soft_state", "check_attempt", "service_id", "st |    0.000 |
| 1502 | icingadb | 172.22.0.35:60716 | icingadb | Prepare | 1214 | Waiting for table metadata lock | INSERT INTO "state_history" ("soft_state", "previous_soft_state", "check_attempt", "service_id", "st |    0.000 |
| 1503 | icingadb | 172.22.0.35:60728 | icingadb | Prepare | 1213 | Waiting for table metadata lock | INSERT INTO "state_history" ("soft_state", "previous_soft_state", "check_attempt", "service_id", "st |    0.000 |
| 1504 | icingadb | 172.22.0.35:60742 | icingadb | Prepare | 1213 | Waiting for table metadata lock | INSERT INTO "state_history" ("soft_state", "previous_soft_state", "check_attempt", "service_id", "st |    0.000 |
| 1505 | icingadb | 172.22.0.36:44314 | icingadb | Prepare | 1213 | Waiting for table metadata lock | INSERT INTO "notification_history" ("object_type", "host_id", "send_time", "author", "text", "endpoi |    0.000 |
| 1506 | icingadb | 172.22.0.35:60752 | icingadb | Prepare | 1213 | Waiting for table metadata lock | INSERT INTO "notification_history" ("previous_hard_state", "text", "object_type", "notification_id", |    0.000 |
| 1507 | icingadb | 172.22.0.36:44324 | icingadb | Prepare | 1212 | Waiting for table metadata lock | INSERT INTO "state_history" ("endpoint_id", "id", "hard_state", "previous_hard_state", "max_check_at |    0.000 |
| 1508 | icingadb | 172.22.0.35:60764 | icingadb | Prepare | 1212 | Waiting for table metadata lock | INSERT INTO "state_history" ("soft_state", "previous_soft_state", "check_attempt", "service_id", "st |    0.000 |
| 1510 | icingadb | 172.22.0.35:50576 | icingadb | Prepare | 1207 | Waiting for table metadata lock | INSERT INTO "notification_history" ("previous_hard_state", "text", "object_type", "notification_id", |    0.000 |
| 1512 | icingadb | 172.22.0.36:35386 | icingadb | Prepare | 1207 | Waiting for table metadata lock | INSERT INTO "notification_history" ("object_type", "host_id", "send_time", "author", "text", "endpoi |    0.000 |
| 1516 | icingadb | 172.22.0.35:50582 | icingadb | Prepare | 1207 | Waiting for table metadata lock | INSERT INTO "notification_history" ("previous_hard_state", "text", "object_type", "notification_id", |    0.000 |
| 1517 | icingadb | 172.22.0.36:35398 | icingadb | Prepare | 1207 | Waiting for table metadata lock | INSERT INTO "notification_history" ("object_type", "host_id", "send_time", "author", "text", "endpoi |    0.000 |
| 1518 | icingadb | 172.22.0.36:35410 | icingadb | Prepare | 1206 | Waiting for table metadata lock | INSERT INTO "notification_history" ("object_type", "host_id", "send_time", "author", "text", "endpoi |    0.000 |
| 1519 | icingadb | 172.22.0.36:35414 | icingadb | Prepare | 1206 | Waiting for table metadata lock | INSERT INTO "state_history" ("endpoint_id", "id", "hard_state", "previous_hard_state", "max_check_at |    0.000 |
| 1520 | icingadb | 172.22.0.36:41900 | icingadb | Prepare | 1197 | Waiting for table metadata lock | INSERT INTO "notification_history" ("object_type", "host_id", "send_time", "author", "text", "endpoi |    0.000 |
| 1521 | icingadb | 172.22.0.35:34160 | icingadb | Prepare | 1197 | Waiting for table metadata lock | INSERT INTO "notification_history" ("previous_hard_state", "text", "object_type", "notification_id", |    0.000 |
| 1522 | icingadb | 172.22.0.36:41906 | icingadb | Prepare | 1197 | Waiting for table metadata lock | INSERT INTO "state_history" ("endpoint_id", "id", "hard_state", "previous_hard_state", "max_check_at |    0.000 |
| 1523 | icingadb | 172.22.0.36:41918 | icingadb | Prepare | 1194 | Waiting for table metadata lock | INSERT INTO "state_history" ("endpoint_id", "id", "hard_state", "previous_hard_state", "max_check_at |    0.000 |
| 1524 | icingadb | 172.22.0.36:41928 | icingadb | Prepare | 1184 | Waiting for table metadata lock | INSERT INTO "state_history" ("endpoint_id", "id", "hard_state", "previous_hard_state", "max_check_at |    0.000 |
| 1525 | icingadb | 172.22.0.35:34166 | icingadb | Prepare | 1193 | Waiting for table metadata lock | INSERT INTO "notification_history" ("previous_hard_state", "text", "object_type", "notification_id", |    0.000 |
| 1530 | icingadb | 172.22.0.36:59066 | icingadb | Prepare | 1182 | Waiting for table metadata lock | INSERT INTO "state_history" ("endpoint_id", "id", "hard_state", "previous_hard_state", "max_check_at |    0.000 |
| 1649 | icingadb | localhost         | icingadb | Query   |    0 | starting                        | show processlist                                                                                     |    0.000 |
+------+----------+-------------------+----------+---------+------+---------------------------------+------------------------------------------------------------------------------------------------------+----------+
39 rows in set (0.001 sec)
```

Killing the `SELECT` connection unblocked everything and from then on, even the `INSERT` statements could run in parallel with the `ALTER TABLE ... ADD INDEX` (note how with each `show processlist`, more and more connections become idle again as they finished all the outstanding work):

```
MariaDB [icingadb]> kill 1485;
Query OK, 0 rows affected (0.001 sec)

MariaDB [icingadb]> show processlist;
+------+----------+-------------------+----------+---------+------+----------------+------------------------------------------------------------------------------------------------------+----------+
| Id   | User     | Host              | db       | Command | Time | State          | Info                                                                                                 | Progress |
+------+----------+-------------------+----------+---------+------+----------------+------------------------------------------------------------------------------------------------------+----------+
|   12 | icingadb | 172.22.0.36:45694 | icingadb | Sleep   |    1 |                | NULL                                                                                                 |    0.000 |
| 1458 | icingadb | 172.22.0.36:58514 | icingadb | Execute |    0 | Update         | INSERT INTO "history" ("environment_id", "host_id", "service_id", "event_type", "notification_histor |    0.000 |
| 1465 | icingadb | 172.22.0.35:37240 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
| 1466 | icingadb | 172.22.0.35:37246 | icingadb | Sleep   |    1 |                | NULL                                                                                                 |    0.000 |
| 1467 | icingadb | 172.22.0.35:37254 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
| 1468 | icingadb | 172.22.0.36:59864 | icingadb | Sleep   |   45 |                | NULL                                                                                                 |    0.000 |
| 1477 | icingadb | 172.22.0.35:33222 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
| 1494 | icingadb | localhost         | icingadb | Query   | 1255 | altering table | ALTER TABLE history ADD INDEX idx_history_event_time_event_type (event_time, event_type) COMMENT 'Hi |    0.000 |
| 1502 | icingadb | 172.22.0.35:60716 | icingadb | Execute |    0 | Update         | INSERT INTO "history" ("notification_history_id", "environment_id", "object_type", "host_id", "event |    0.000 |
| 1503 | icingadb | 172.22.0.35:60728 | icingadb | Execute |    0 | Update         | INSERT INTO "state_history" ("soft_state", "previous_soft_state", "check_attempt", "service_id", "st |    0.000 |
| 1508 | icingadb | 172.22.0.35:60764 | icingadb | Execute |    0 | Commit         | INSERT INTO "history" ("id", "state_history_id", "environment_id", "endpoint_id", "object_type", "ho |    0.000 |
| 1519 | icingadb | 172.22.0.36:35414 | icingadb | Execute |    0 | Update         | INSERT INTO "notification_history" ("object_type", "host_id", "send_time", "author", "text", "endpoi |    0.000 |
| 1521 | icingadb | 172.22.0.35:34160 | icingadb | Execute |    1 | Update         | INSERT INTO "notification_history" ("previous_hard_state", "text", "object_type", "notification_id", |    0.000 |
| 1523 | icingadb | 172.22.0.36:41918 | icingadb | Execute |    0 | Commit         | INSERT INTO "user_notification_history" ("notification_history_id", "user_id", "environment_id", "id |    0.000 |
| 1524 | icingadb | 172.22.0.36:41928 | icingadb | Execute |    0 | Update         | INSERT INTO "service_state" ("is_flapping", "next_check", "performance_data", "severity", "id", "hos |    0.000 |
| 1530 | icingadb | 172.22.0.36:59066 | icingadb | Execute |    0 | Update         | INSERT INTO "history" ("endpoint_id", "object_type", "service_id", "state_history_id", "environment_ |    0.000 |
| 1649 | icingadb | localhost         | icingadb | Query   |    0 | starting       | show processlist                                                                                     |    0.000 |
| 1651 | icingadb | 172.22.0.36:55346 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
| 1652 | icingadb | 172.22.0.36:55358 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
+------+----------+-------------------+----------+---------+------+----------------+------------------------------------------------------------------------------------------------------+----------+
19 rows in set (0.001 sec)

MariaDB [icingadb]> show processlist;
+------+----------+-------------------+----------+---------+------+----------------+------------------------------------------------------------------------------------------------------+----------+
| Id   | User     | Host              | db       | Command | Time | State          | Info                                                                                                 | Progress |
+------+----------+-------------------+----------+---------+------+----------------+------------------------------------------------------------------------------------------------------+----------+
|   12 | icingadb | 172.22.0.36:45694 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
| 1458 | icingadb | 172.22.0.36:58514 | icingadb | Sleep   |    4 |                | NULL                                                                                                 |    0.000 |
| 1465 | icingadb | 172.22.0.35:37240 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
| 1466 | icingadb | 172.22.0.35:37246 | icingadb | Sleep   |    6 |                | NULL                                                                                                 |    0.000 |
| 1467 | icingadb | 172.22.0.35:37254 | icingadb | Sleep   |    5 |                | NULL                                                                                                 |    0.000 |
| 1468 | icingadb | 172.22.0.36:59864 | icingadb | Sleep   |   51 |                | NULL                                                                                                 |    0.000 |
| 1477 | icingadb | 172.22.0.35:33222 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
| 1494 | icingadb | localhost         | icingadb | Query   | 1260 | altering table | ALTER TABLE history ADD INDEX idx_history_event_time_event_type (event_time, event_type) COMMENT 'Hi |    0.000 |
| 1502 | icingadb | 172.22.0.35:60716 | icingadb | Sleep   |    4 |                | NULL                                                                                                 |    0.000 |
| 1503 | icingadb | 172.22.0.35:60728 | icingadb | Execute |    2 | Update         | INSERT INTO "history" ("notification_history_id", "environment_id", "object_type", "host_id", "event |    0.000 |
| 1508 | icingadb | 172.22.0.35:60764 | icingadb | Sleep   |    3 |                | NULL                                                                                                 |    0.000 |
| 1521 | icingadb | 172.22.0.35:34160 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
| 1523 | icingadb | 172.22.0.36:41918 | icingadb | Sleep   |    4 |                | NULL                                                                                                 |    0.000 |
| 1524 | icingadb | 172.22.0.36:41928 | icingadb | Sleep   |    4 |                | NULL                                                                                                 |    0.000 |
| 1649 | icingadb | localhost         | icingadb | Query   |    0 | starting       | show processlist                                                                                     |    0.000 |
| 1651 | icingadb | 172.22.0.36:55346 | icingadb | Sleep   |    4 |                | NULL                                                                                                 |    0.000 |
| 1652 | icingadb | 172.22.0.36:55358 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
| 1653 | icingadb | 172.22.0.36:55362 | icingadb | Execute |    1 | Update         | INSERT INTO "history" ("environment_id", "host_id", "service_id", "event_type", "notification_histor |    0.000 |
+------+----------+-------------------+----------+---------+------+----------------+------------------------------------------------------------------------------------------------------+----------+
18 rows in set (0.002 sec)

MariaDB [icingadb]> show processlist;
+------+----------+-------------------+----------+---------+------+----------------+------------------------------------------------------------------------------------------------------+----------+
| Id   | User     | Host              | db       | Command | Time | State          | Info                                                                                                 | Progress |
+------+----------+-------------------+----------+---------+------+----------------+------------------------------------------------------------------------------------------------------+----------+
|   12 | icingadb | 172.22.0.36:45694 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
| 1458 | icingadb | 172.22.0.36:58514 | icingadb | Sleep   |    1 |                | NULL                                                                                                 |    0.000 |
| 1465 | icingadb | 172.22.0.35:37240 | icingadb | Sleep   |    0 |                | NULL                                                                                                 |    0.000 |
| 1466 | icingadb | 172.22.0.35:37246 | icingadb | Sleep   |   12 |                | NULL                                                                                                 |    0.000 |
| 1467 | icingadb | 172.22.0.35:37254 | icingadb | Sleep   |   10 |                | NULL                                                                                                 |    0.000 |
| 1468 | icingadb | 172.22.0.36:59864 | icingadb | Sleep   |   56 |                | NULL                                                                                                 |    0.000 |
| 1477 | icingadb | 172.22.0.35:33222 | icingadb | Sleep   |    5 |                | NULL                                                                                                 |    0.000 |
| 1494 | icingadb | localhost         | icingadb | Query   | 1266 | altering table | ALTER TABLE history ADD INDEX idx_history_event_time_event_type (event_time, event_type) COMMENT 'Hi |    0.000 |
| 1502 | icingadb | 172.22.0.35:60716 | icingadb | Sleep   |   10 |                | NULL                                                                                                 |    0.000 |
| 1508 | icingadb | 172.22.0.35:60764 | icingadb | Sleep   |    1 |                | NULL                                                                                                 |    0.000 |
| 1521 | icingadb | 172.22.0.35:34160 | icingadb | Sleep   |    1 |                | NULL                                                                                                 |    0.000 |
| 1523 | icingadb | 172.22.0.36:41918 | icingadb | Sleep   |   10 |                | NULL                                                                                                 |    0.000 |
| 1524 | icingadb | 172.22.0.36:41928 | icingadb | Sleep   |   10 |                | NULL                                                                                                 |    0.000 |
| 1649 | icingadb | localhost         | icingadb | Query   |    0 | starting       | show processlist                                                                                     |    0.000 |
| 1651 | icingadb | 172.22.0.36:55346 | icingadb | Sleep   |   10 |                | NULL                                                                                                 |    0.000 |
| 1652 | icingadb | 172.22.0.36:55358 | icingadb | Sleep   |    1 |                | NULL                                                                                                 |    0.000 |
+------+----------+-------------------+----------+---------+------+----------------+------------------------------------------------------------------------------------------------------+----------+
16 rows in set (0.001 sec)
```

So looks like there's a risk of blocking the whole database if there's some long-running query (in my case, the one that the new index is supposed to speed up). I'm not sure though if we can do anything about this apart from warning about it in the upgrade docs.

yhabteab:
I think it's because MySQL/MariaDB performs a `filesort` operation on a simple `SELECT`, I have no idea why that would block other queries though.

```sql
MariaDB [icingadb]> EXPLAIN select * from history order by event_time, event_type limit 25;
+------+-------------+---------+------+---------------+------+---------+------+-------+----------------+
| id   | select_type | table   | type | possible_keys | key  | key_len | ref  | rows  | Extra          |
+------+-------------+---------+------+---------------+------+---------+------+-------+----------------+
|    1 | SIMPLE      | history | ALL  | NULL          | NULL | NULL    | NULL | 18878 | Using filesort |
+------+-------------+---------+------+---------------+------+---------+------+-------+----------------+
1 row in set (0.005 sec)

MariaDB [icingadb]> EXPLAIN select * from history order by event_time limit 25;
+------+-------------+---------+-------+---------------+------------------------+---------+------+------+-------+
| id   | select_type | table   | type  | possible_keys | key                    | key_len | ref  | rows | Extra |
+------+-------------+---------+-------+---------------+------------------------+---------+------+------+-------+
|    1 | SIMPLE      | history | index | NULL          | idx_history_event_time | 8       | NULL | 25   |       |
+------+-------------+---------+-------+---------------+------------------------+---------+------+------+-------+
1 row in set (0.001 sec
```

julianbrost:
My guess for what's going on is some interaction of shared locks: Probably the `SELECT` holds a read lock on the metadata (if it didn't hold a lock, all the other connections wouldn't have to wait). On its own, this lock allows other queries. Maybe doing an `ALTER TABLE` now requests an exclusive lock which blocks all new readers as well (in an effort to guarantee that the exclusive lock is acquired at some point).

lippserd:
> So looks like there's a risk of blocking the whole database if there's some long-running query (in my case, the one that the new index is supposed to speed up). I'm not sure though if we can do anything about this apart from warning about it in the upgrade docs.

I also think it's a good idea to have this written somewhere. Having the processlists documented as you did would be also good. So a user can quickly identify such a scenario and fix it.

julianbrost (changes requested):
After #656 is done (otherwise this would just add merge conflicts anyways), please do the following:
- Rebase and resolve conflicts
- Add hints to the upgrading docs added in #656 that point out the potential locking problem during the upgrade (https://github.com/Icinga/icingadb/pull/686#issuecomment-1988441677, https://github.com/Icinga/icingadb/pull/686#issuecomment-2016937396)

yhabteab:
> After #656 is done (otherwise this would just add merge conflicts anyways)

Just like the aforementioned PR would do here! That is not an excuse for not having merged this before, which by the way has already been approved.

doc/04-Upgrading.md
```diff
@@ -3,6 +3,54 @@
...
 
+## Upgrading to Icinga DB v1.1.2
+
+Please apply the `1.1.2.sql` upgrade script to your database. For package installations, you can find this file at
+`/usr/share/icingadb/schema/mysql/upgrades/` or `/usr/share/icingadb/schema/pgsql/upgrades/`, depending on your
+database vendor.
+
+!!! warning
+
+    In order to fix a loading performance issue of the `icingadb/history` view, this upgrade script adds an appropriate
```
julianbrost, line 14: You might want to explicitly mention Icinga Web here. While I get that this is probably a relative URL in Icinga Web, I don't think it'll be clear to the average user. So I'd just refer to it as something like "history view in Icinga DB Web".

doc/04-Upgrading.md
```diff
@@ -3,6 +3,54 @@
...
+    index on the `history` table. However, this may hang for a relatively considerable time and block all Icinga DB
+    queries on all `*_history` tables and the `history` table inclusively if there is an ongoing, long-running query
+    on the `history` table. One way of causing this to happen is if an Icinga Web user accesses the `icingadb/history`
```
julianbrost, line 17: I'd add a brief sentence in between here stating the desired behavior, something like: "Creating this index normally happens without blocking other queries. However, [there is this specific situation where some locking can go wrong ...]".

julianbrost (changes requested):

doc/04-Upgrading.md
julianbrost: Oops, by putting it into square brackets, I wanted to signal it as a sloppy summary of what is about to follow, not a word by word suggestion :joy:

yhabteab: I will change that!

doc/04-Upgrading.md
```diff
@@ -3,6 +3,64 @@
...
+!!! warning
+
+    In order to fix a loading performance issue of the history view in Icinga DB Web, this upgrade script adds an
+    appropriate index on the `history` table. Creating this new index normally takes place without blocking any other
+    queries. However, this may hang for a relatively considerable time, blocking all Icinga DB queries on all
+    `*_history` tables and the `history` table inclusively if there is an ongoing, long-running query on the `history`
+    table. One way of causing this to happen is if an Icinga Web user accesses the `icingadb/history` view at the exact
+    moment you are running this script. Depending on how many entries you have in the history table, Icinga DB Web may
+    take quite a long time to load, until your web servers timeout kicks in. 
+    See [unblock history tables](#unblock-history-tables) for details.
```
julianbrost, line 21: > at the exact moment you are running this script

Not really. Given that it's a long-running query, it's actually a quite long time window. As you mention the web server timeout, I'm not sure about the details. At least when using the `icinga/icingaweb2` docker image, there seems to be no timeout or a very long timeout as you can see from [my `show processlist` output](https://github.com/Icinga/icingadb/pull/686#issuecomment-1988441677) where I had some of these queries running for over 1200s (=20min).

If there was a timeout that would abort the request and its query after say 30 seconds, I wouldn't really bother writing this in the documentation because the wait would be shorter than you'd need to even figure out that you're in this situation.

doc/04-Upgrading.md
```diff
@@ -3,6 +3,64 @@
...
+and all Icinga DB queries on all `*_history` tables or forcibly terminate them and let the remaining queries do their
+work. In this case, cancelling that one blocking `SELECT` query will let the upgrade script continue normally without
+blocking any other queries.
+``` 
+MariaDB [icingadb]> kill 1485;
+```
+
+In case you are insecure about which Icinga DB Web queries are blocking, you may simply cancel all `SELECT` statements
+listed with `show processlist`. Cancelling a `SELECT` query is usually not that disastrous, so feel free to abort every
+single one of them.
```
julianbrost, line 62: I'd be more cautious with such suggestions. Keep in mind that `show processlist` shows everything on the server including other databases/applications if present. I wouldn't want to recommend "just kill it" for everything, maybe just for long-running (see time column) selects on that specific database.

julianbrost (changes requested):

doc/04-Upgrading.md
yhabteab: I don't know how and why the configurations are getting lost with `icinga/icingaweb2` since we don't use this image for our dev environment, but there is for sure a default PHP [`max_execution_time`](https://www.php.net/manual/en/info.configuration.php) of `30s`.
```bash
root@768029d1c1dc:/var/www/html# grep -Rne 'max_execution_time' /usr/local/etc/php/
/usr/local/etc/php/php.ini:407:max_execution_time = 30
/usr/local/etc/php/php.ini-development:407:max_execution_time = 30
/usr/local/etc/php/php.ini-production:409:max_execution_time = 30
```

You can also set the execution timeout on a [web server](https://httpd.apache.org/docs/2.4/de/mod/core.html#timeout) basis, but we don't use apache2 for our dev setup either. However, we do set the [`FastCGI`](https://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_read_timeout) read timeout for Nginx, which I think is useless as PHP aborts the requests after `30` by itself and doesn't wait for the CGI timeout to kick in.
```bash
root@e41052c1f5f3:/# grep -Rne 'fastcgi_read_timeout' /etc/nginx
/etc/nginx/nginx.conf:27:    fastcgi_read_timeout 3600s;
```

doc/04-Upgrading.md
```diff
@@ -3,6 +3,64 @@
...
+```
+MariaDB [icingadb]> show processlist;
++------+-----+-----+----------+-----+------+---------------------------------+------------------------------------+-----+
+| Id   | ... | ... | db       | ... | Time | State                           | Info                               | ... |
++------+-----+-----+----------+-----+------+---------------------------------+------------------------------------+-----+
+| 1475 | ... | ... | icingadb | ... | 1222 | Waiting for table metadata lock | INSERT INTO "notification_history" | ... |
+| 1485 | ... | ... | icingadb | ... | 1262 | Creating sort index             | SELECT history.id, history....     | ... |
+| 1494 | ... | ... | icingadb | ... | 1224 | Waiting for table metadata lock | ALTER TABLE history ADD INDEX ...  | ... |
+| 1499 | ... | ... | icingadb | ... | 1215 | Waiting for table metadata lock | INSERT INTO "notification_history" | ... |
+| 1500 | ... | ... | icingadb | ... | 1215 | Waiting for table metadata lock | INSERT INTO "state_history" ...    | ... |
```
julianbrost, line 39: I'd add a full `| ... | ... | ...` row to make it more obvious that rows were omitted.

doc/04-Upgrading.md
```diff
@@ -3,6 +3,64 @@
...
++------+-----+-----+----------+-----+------+---------------------------------+------------------------------------+-----+
+39 rows in set (0.001 sec)
+```
+
+In the above output you can see that there are approximately `39` queries (`39 rows in set (0.001 sec)`), including the
```
julianbrost, line 44: I wouldn't focus at all on the number of rows returned, maybe even omit it from the `show processlist` output. With idle connections, that number would also be high, if there are other applications on the same MySQL server, the number may be even higher without it saying anything useful.

The relevant things to look for is a long running `SELECT` and the `ALTER TABLE` (and possibly a number of `INSERT` queries[^1]) in the "Waiting for table metadata lock" state within the Icinga DB database.

[^1]: In case no history events are currently generated, there would also be no `INSERT` queries running. Though that's probably unlikely in a setup where the `SELECT` is taking long. Nonetheless, those are just "secondary" victims of the same lock being held, they are not necessary or responsible for this situation to happen.

doc/04-Upgrading.md
```diff
@@ -3,6 +3,64 @@
...
+and all Icinga DB queries on all `*_history` tables or forcibly terminate them and let the remaining queries do their
+work. In this case, cancelling that one blocking `SELECT` query will let the upgrade script continue normally without
+blocking any other queries.
+``` 
+MariaDB [icingadb]> kill 1485;
+```
+
+In case you are insecure about which Icinga DB Web queries are blocking, you may simply cancel all long-running
+`SELECT` statements listed with `show processlist` (see column `Time`). Cancelling a `SELECT` query is usually not that
+disastrous, so feel free to abort every single one of them matching the Icinga DB database (see column `db`).
```
julianbrost, line 62: > usually not that disastrous

That wording isn't really reassuring if I'm already unsure. I mean what's the worst that could happen? The same as if the database was stopped at that moment.

julianbrost (changes requested):

doc/04-Upgrading.md
yhabteab: That's simply pointless to me! It's just a simple sample something that shouldn't really matter, how many lines I'm rendering here and it's also described immediately afterwards that lines have been cut out.

doc/04-Upgrading.md
```diff
@@ -3,6 +3,64 @@
...
+    queries. However, this may hang for a relatively considerable time, blocking all Icinga DB queries on all
+    `*_history` tables and the `history` table inclusively if there is an ongoing, long-running query on the `history`
+    table. One way of causing this to happen is if an Icinga Web user accesses the `icingadb/history` view just before
+    you are running this script. Depending on how many entries you have in the history table, Icinga DB Web may take
+    quite a long time to load, until your web servers timeout (if any) kicks in. 
+    See [unblock history tables](#unblock-history-tables) for details.
+
+### Unblock History Tables
+
+When you observe that the upgrade script has been taking an unusually long to complete, you can perform the following
```
oxzi, line 25:
```suggestion
When you observe that the upgrade script has been taking unusually long to complete, you can perform the following
```

Maybe also give a rough number so that users have an idea how long "unusually long" is.

doc/04-Upgrading.md
```diff
@@ -3,6 +3,64 @@
...
+work. In this case, cancelling that one blocking `SELECT` query will let the upgrade script continue normally without
+blocking any other queries.
+``` 
+MariaDB [icingadb]> kill 1485;
+```
+
+In case you are insecure about which Icinga DB Web queries are blocking, you may simply cancel all long-running
+`SELECT` statements listed with `show processlist` (see column `Time`). Cancelling a `SELECT` query is usually not that
+disastrous, so feel free to abort every single one of them matching the Icinga DB database (see column `db`).
+
```
oxzi, line 63: This was all very detailed, but limited to MySQL-like databases. As this also affects PostgreSQL, this should also be covered.

doc/04-Upgrading.md
yhabteab: > I mean what's the worst that could happen? The same as if the database was stopped at that moment.

Is that really that important? We already retry all database errors anyway. So, how badly could Icinga DB be damaged if it had to repeat its queries once again?

doc/04-Upgrading.md
julianbrost: Then remove the "39 rows in set (0.001 sec)" line instead. I think it gives no useful information anyways (see https://github.com/Icinga/icingadb/pull/686#discussion_r1549414038) and as it's currently given, the table just looks wrong to me.

doc/04-Upgrading.md
julianbrost: My comment is about the "usually not that disastrous" wording specifically. Isn't the whole point of this sentence to reassure users that killing a `SELECT` query is fine? But with this wording, one may ask oneself why it says "usually" and how one can be sure that this isn't an unusual situation where it is disastrous.

> I mean what's the worst that could happen? The same as if the database was stopped at that moment.

That's not what I would write into the docs. That's just my justification why I don't think we need to use very careful wording here. Killing a query is no worse than stopping the database, or even just losing the network connection. If that would be disastrous, you have other problems.

doc/04-Upgrading.md
julianbrost: Fair point. Have you tried how exactly this behaves with PostgreSQL? At least with PostgreSQL≥13 and the history view, I don't expect such slow queries (see https://github.com/Icinga/icingadb/pull/686#issuecomment-1988043508). But I don't know what exactly will happen with older versions (or other slow queries if you run them manually).

oxzi: As PostgreSQL 12 is still being [supported till end of this year](https://www.postgresql.org/support/versioning/), this might be something to at least mention. Otherwise, when there's no issue with a version greater 12, I would say to just add one paragraph mentioning potential issues with PostgreSQL 12 and advising people to upgrade, and mentioning that newer versions should be fine. Otherwise PostgreSQL users might be unsure to upgrade, as there is such a long text regarding potential issues with MySQL.

yhabteab: **Before:**
```pgsql
icingadb=# \c timing on
Timing is on.
icingadb=# SELECT COUNT(*) FROM history;
 count  
--------
 405399
(1 row)

Time: 33.537 ms

icingadb=# SELECT 1 as name FROM history ORDER BY event_time, event_type LIMIT 1;
 name 
------
    1
(1 row)

Time: 38.162 ms
```

**After:**
```pgsql
icingadb=# SELECT COUNT(*) FROM history;
 count  
--------
 414913
(1 row)

Time: 20.501 ms
icingadb=# SELECT 1 as name FROM history ORDER BY event_time, event_type LIMIT 1;
 name 
------
    1
(1 row)

Time: 0.382 ms
```

yhabteab:
```bash
root@713a54a05873:/# psql --version
psql (PostgreSQL) 12.18 (Debian 12.18-1.pgdg120+2)
```

julianbrost: That shows that the extended index indeed provides a benefit for older PostgreSQL versions. However, I think this thread is about whether a situation similar to "Waiting for table metadata lock" can happen with PostgreSQL. But so far, I don't know if anyone actually tried to verify this. It could also be an option that it handles the metadata sufficiently different than MySQL so that it doesn't have a similar locking problem.

schema/pgsql/upgrades/1.1.2.sql
```diff
@@ -137,3 +137,8 @@ BEGIN
   RETURN (100 * (total_time - problem_time)::decimal / total_time)::decimal(7, 4);
 END;
 $$;
+
+CREATE INDEX idx_history_event_time_event_type ON history(event_time, event_type);
```
julianbrost, line 141: To discuss: do we want to use [`CREATE INDEX CONCURRENTLY`](https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) here instead? It will scan the table twice but allows for concurrent modifications of the table (reading is already possibly without `CONCURRENTLY`).

doc/04-Upgrading.md
yhabteab: If we create the index that https://github.com/Icinga/icingadb/pull/686#discussion_r1549825905 way, shouldn't this discussion become irrelevant then? If PostgreSQL says that it won't hold any locks, there shouldn't be such a situation as with MySQL.

julianbrost: > If PostgreSQL says that it won't hold any locks

Does it say so? https://www.postgresql.org/docs/current/sql-createindex.html says for `CONCURRENTLY`:

> When this option is used, PostgreSQL will build the index without taking any locks that prevent concurrent inserts, updates, or deletes on the table; whereas a standard index build locks out writes (but not reads) on the table until it's done.

That doesn't say anything about metadata locking, where this glitch happens with MySQL. PostgreSQL isn't trying to reimplement MySQL, so that doesn't mean it must have the same problem. So far, the only reason I see for this actually being an issue with PostgreSQL is the claim "As this also affects PostgreSQL" at the beginning of the discussion but my question in https://github.com/Icinga/icingadb/pull/686#discussion_r1549709572 whether this was actually observed, wasn't answered so far (@oxzi).

oxzi: Let me rephrase my original comment: The update also affects PostgreSQL, not just MySQL and its derivatives. So if I, as a potential PostgreSQL user, would scroll through the update logs and see lots of details about potential problems with MySQL, but not a single word about PostgreSQL, I might be afraid to apply the schema update.

As the warning above this paragraph now (or was it already before and I just missed it?) starts with a note regarding PostgreSQL, my objections are now void.

doc/04-Upgrading.md
```diff
@@ -3,6 +3,65 @@
...
+    Please ignore this warning if you are using PostgreSQL. This applies to MySQL/MariaDB users only.
+
+    In order to fix a loading performance issue of the history view in Icinga DB Web, this upgrade script adds an
+    appropriate index on the `history` table. Creating this new index normally takes place without blocking any other
+    queries. However, this may hang for a relatively considerable time, blocking all Icinga DB queries on all
+    `*_history` tables and the `history` table inclusively if there is an ongoing, long-running query on the `history`
+    table. One way of causing this to happen is if an Icinga Web user accesses the `icingadb/history` view just before
+    you are running this script. Depending on how many entries you have in the history table, Icinga DB Web may take
+    quite a long time to load, until your web servers timeout (if any) kicks in. 
+    See [unblock history tables](#unblock-history-tables) for details.
```
julianbrost, line 23: I think this is quite lengthy for a warning block given that there's a separate section for the details already.

My suggestion for the warning block:

> With MySQL and MariaDB, a locking issue can occur if the schema upgrade is applied while the history view is accessed in Icinga DB Web. This can result in the upgrade being delayed unnecessarily and it blocking other queries. Please see unblock history tables for how to detect and resolve this situation.

All the details that are removed by this change can be moved to the section below.

doc/04-Upgrading.md
```diff
@@ -3,6 +3,65 @@
...
+
+### Unblock History Tables
```
julianbrost, line 25: So far, within this section, the only hint that this only applies to MySQL/MariaDB is the "MariaDB [icingadb]>" prompt. Maybe just add "(MySQL/MariaDB only)" to its title or add an [`!!! info` block](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types).

julianbrost (changes requested):

doc/04-Upgrading.md
yhabteab: > Please ignore this warning if you are using PostgreSQL. This applies to MySQL/MariaDB users only.

Does your suggestion include this hint? as well?

julianbrost: For me, "With MySQL and MariaDB," would be enough of a hint that this does not affect PostgreSQL. I think you added that line in response to @oxzi, so let's ask him if this suffices or whether it should be more explicit about PostgreSQL. If I remember correctly, it previously didn't mention a database type at all (apart from in the prompts and implicitly by using MySQL/MariaDB-specific commands).

oxzi (approved):

PR #692, opened 2024-03-12
Enhance HA "Taking over", "Handing over" logging

### LINKED ISSUE #688: Enhance HA take/hand over log messages
## Describe the bug

The `Handing over` and `Taking over` log message should include the reason, why it's handing/taking over. Additionally, we should also include the environment ID in these messages.

oxzi:
The reason for a switch in the HA roles was not always directly clear. This change now introduces additional logging, indicating the reasoning for either taking over or handing over the HA responsibility.

Next to additional logging messages, both the takeover and handover channel are now transporting a string to communicate the reason instead of an empty struct{}. By doing so, both the "Taking over" and "Handing over" log messages are enriched with a reason.

This also required a change in the suppressed logging handling of the HA.realize method, which got its logging enabled through the shouldLog parameter. Now, there are both recurring events, which might be suppressed, as well as state changing events, which should be logged. Therefore, and because the logTicker's functionality was not clear to me on first glance, I renamed it to routineLogTicker.

While dealing with the code, some function signature documentation were added, to ease both mine as well as the understanding of future readers.

Closes #688.

### DIFF
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -23,7 +23,10 @@ import (
 	"time"
 )
 
-var timeout = 60 * time.Second
+// peerTimeout defines the timeout for HA heartbeats, being used to detect absent nodes.
+//
+// Because this timeout relies on icingaredis.Timeout, it is icingaredis.Timeout plus a short grace period.
+const peerTimeout = icingaredis.Timeout + 5*time.Second
 
 type haState struct {
 	responsibleTsMilli int64
@@ -43,8 +46,8 @@ type HA struct {
 	heartbeat     *icingaredis.Heartbeat
 	logger        *logging.Logger
 	responsible   bool
-	handover      chan struct{}
-	takeover      chan struct{}
+	handover      chan string
+	takeover      chan string
 	done          chan struct{}
 	errOnce       sync.Once
 	errMu         sync.Mutex
@@ -64,8 +67,8 @@ func NewHA(ctx context.Context, db *DB, heartbeat *icingaredis.Heartbeat, logger
 		db:         db,
 		heartbeat:  heartbeat,
 		logger:     logger,
-		handover:   make(chan struct{}),
-		takeover:   make(chan struct{}),
+		handover:   make(chan string),
+		takeover:   make(chan string),
 		done:       make(chan struct{}),
 	}
 
@@ -107,13 +110,13 @@ func (h *HA) Err() error {
 	return h.err
 }
 
-// Handover returns a channel with which handovers are signaled.
-func (h *HA) Handover() chan struct{} {
+// Handover returns a channel with which handovers and their reasons are signaled.
+func (h *HA) Handover() chan string {
 	return h.handover
 }
 
-// Takeover returns a channel with which takeovers are signaled.
-func (h *HA) Takeover() chan struct{} {
+// Takeover returns a channel with which takeovers and their reasons are signaled.
+func (h *HA) Takeover() chan string {
 	return h.takeover
 }
 
@@ -141,9 +144,10 @@ func (h *HA) controller() {
 
 	oldInstancesRemoved := false
 
-	logTicker := time.NewTicker(time.Second * 60)
-	defer logTicker.Stop()
-	shouldLog := true
+	// Suppress recurring log messages in the realize method to be only logged this often.
+	routineLogTicker := time.NewTicker(5 * time.Minute)
+	defer routineLogTicker.Stop()
+	shouldLogRoutineEvents := true
 
 	for {
 		select {
@@ -158,9 +162,9 @@ func (h *HA) controller() {
 				if tt.After(now.Add(1 * time.Second)) {
 					h.logger.Debugw("Received heartbeat from the future", zap.Time("time", tt))
 				}
-				if tt.Before(now.Add(-1 * timeout)) {
+				if tt.Before(now.Add(-1 * peerTimeout)) {
 					h.logger.Errorw("Received heartbeat from the past", zap.Time("time", tt))
-					h.signalHandover()
+					h.signalHandover("received heartbeat from the past")
 					h.realizeLostHeartbeat()
 					continue
 				}
@@ -192,8 +196,8 @@ func (h *HA) controller() {
 				}
 
 				select {
-				case <-logTicker.C:
-					shouldLog = true
+				case <-routineLogTicker.C:
+					shouldLogRoutineEvents = true
 				default:
 				}
 
@@ -204,10 +208,10 @@ func (h *HA) controller() {
 				} else {
 					realizeCtx, cancelRealizeCtx = context.WithCancel(h.ctx)
 				}
-				err = h.realize(realizeCtx, s, t, envId, shouldLog)
+				err = h.realize(realizeCtx, s, t, envId, shouldLogRoutineEvents)
 				cancelRealizeCtx()
 				if errors.Is(err, context.DeadlineExceeded) {
-					h.signalHandover()
+					h.signalHandover("context deadline exceeded")
 					continue
 				}
 				if err != nil {
@@ -219,10 +223,10 @@ func (h *HA) controller() {
 					oldInstancesRemoved = true
 				}
 
-				shouldLog = false
+				shouldLogRoutineEvents = false
 			} else {
 				h.logger.Error("Lost heartbeat")
-				h.signalHandover()
+				h.signalHandover("lost heartbeat")
 				h.realizeLostHeartbeat()
 			}
 		case <-h.heartbeat.Done():
@@ -235,13 +239,25 @@ func (h *HA) controller() {
 	}
 }
 
-func (h *HA) realize(ctx context.Context, s *icingaredisv1.IcingaStatus, t *types.UnixMilli, envId types.Binary, shouldLog bool) error {
-	var takeover, otherResponsible bool
+// realize a HA cycle triggered by a heartbeat event.
+//
+// shouldLogRoutineEvents indicates if recurrent events should be logged.
+func (h *HA) realize(
+	ctx context.Context,
+	s *icingaredisv1.IcingaStatus,
+	t *types.UnixMilli,
+	envId types.Binary,
+	shouldLogRoutineEvents bool,
+) error {
+	var (
+		takeover         string
+		otherResponsible bool
+	)
 
 	err := retry.WithBackoff(
 		ctx,
 		func(ctx context.Context) error {
-			takeover = false
+			takeover = ""
 			otherResponsible = false
 			isoLvl := sql.LevelSerializable
 			selectLock := ""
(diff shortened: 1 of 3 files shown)


pkg/icingadb/ha.go
```diff
@@ -274,15 +287,25 @@ func (h *HA) realize(ctx context.Context, s *icingaredisv1.IcingaStatus, t *type
...
 						zap.String("instance_id", instance.Id.String()),
 						zap.String("environment", envId.String()),
-						"heartbeat", instance.Heartbeat,
+						zap.Time("heartbeat", instance.Heartbeat.Time()),
 						zap.Duration("heartbeat_age", time.Since(instance.Heartbeat.Time())))
 				}
 			case sql.ErrNoRows:
-				takeover = true
+				fields := []any{
+					zap.String("instance_id", instance.Id.String()),
```
yhabteab, line 299: When the SQL query error type matches `sql.errNoRows`, it won't scan any rows into the `instance` variable and thus `instance.Id` will always be empty.

pkg/icingadb/ha.go
```diff
@@ -274,15 +287,25 @@ func (h *HA) realize(ctx context.Context, s *icingaredisv1.IcingaStatus, t *type
...
+					zap.String("environment", envId.String()),
+					zap.Duration("heartbeat_timeout", timeout)}
+
+				if !h.responsible {
+					h.logger.Infow("Preparing to take over HA as no instance is active", fields...)
```
yhabteab, line 304: Is this even necessary when we're going to log `Taking over ...` in the main function just in the next line? I think it would be sufficient to just include the environment ID as part of the `Taking over ...` log line and then remove this completely.

pkg/icingadb/ha.go
```diff
@@ -274,15 +287,25 @@ func (h *HA) realize(ctx context.Context, s *icingaredisv1.IcingaStatus, t *type
...
+				} else if h.responsible && shouldLogRoutineEvents {
+					h.logger.Infow("Continuing being the active instance", fields...)
```
yhabteab, line 306: Is that even possible? I can't imagine a situation in which we are preparing for the takeover but at the same time being responsible/the active instance.

pkg/icingadb/ha.go
oxzi: This will look like the following:

```
icingadb-master2  | 2024-03-14T11:24:05.914Z    INFO    high-availability       Preparing to take over HA as no instance is active      {"envir
onment": "cef9c044b80d8cf516b031c0162ef41394b36498", "heartbeat_timeout": "1m0s"}
icingadb-master2  | 2024-03-14T11:24:05.919Z    INFO    icingadb        Taking over     {"reason": "no other instance is active"}
```

Yes, this could be seen as a duplicate, but in the rare case that between making the decision to take over and actually taking over something breaks, we would know on which code path Icinga DB was.

Maybe it could be reduced to debug logging?

pkg/icingadb/ha.go
oxzi: This condition will always be true if this Icinga DB instance is the active one.

The SQL query above selects all responsible instances which are not the current one.  Thus, `sql.ErrNoRows` will be returned.
https://github.com/Icinga/icingadb/blob/7d25da7815de6035e958c1bd0331f7669f95b090/pkg/icingadb/ha.go#L279-L285

That's how it looks on my current testing setup:
```
icingadb-master2  | 2024-03-14T11:25:02.914Z    INFO    high-availability       Continuing being the active instance    {"environment": "cef9c044b80d8cf516b031c0162ef41394b36498", "heartbeat_timeout": "1m0s"}
icingadb-master2  | 2024-03-14T11:26:02.913Z    INFO    high-availability       Continuing being the active instance    {"environment": "cef9c044b80d8cf516b031c0162ef41394b36498", "heartbeat_timeout": "1m0s"}
icingadb-master2  | 2024-03-14T11:27:02.914Z    INFO    high-availability       Continuing being the active instance    {"environment": "cef9c044b80d8cf516b031c0162ef41394b36498", "heartbeat_timeout": "1m0s"}
icingadb-master2  | 2024-03-14T11:28:02.913Z    INFO    high-availability       Continuing being the active instance    {"environment": "cef9c044b80d8cf516b031c0162ef41394b36498", "heartbeat_timeout": "1m0s"}
icingadb-master2  | 2024-03-14T11:29:02.915Z    INFO    high-availability       Continuing being the active instance    {"environment": "cef9c044b80d8cf516b031c0162ef41394b36498", "heartbeat_timeout": "1m0s"}
```

oxzi: Btw, by guarding this with the `shouldLogRoutineEvents` variable, this will be logged once every minute, similar to the opposite _"Another instance is active"_ message on the other node.

pkg/icingadb/ha.go
oxzi: After giving this a second thought, I would degrade this to `DEBUG`. Same as the two sibling messages "Another instance is active" and "Continuing being the active instance" as they are, by definition, only useful for debugging.

I would like to write the second part to <https://github.com/Icinga/icingadb/pull/692#discussion_r1528720666>, but GitHub does not show me a text box to comment below this comment.

pkg/icingadb/ha.go
```diff
@@ -274,15 +287,25 @@ func (h *HA) realize(ctx context.Context, s *icingaredisv1.IcingaStatus, t *type
...
-						"heartbeat", instance.Heartbeat,
+						zap.Time("heartbeat", instance.Heartbeat.Time()),
 						zap.Duration("heartbeat_age", time.Since(instance.Heartbeat.Time())))
 				}
 			case sql.ErrNoRows:
-				takeover = true
+				fields := []any{
+					zap.String("instance_id", h.instanceId.String()),
+					zap.String("environment", envId.String()),
+					zap.Duration("heartbeat_timeout", timeout)}
```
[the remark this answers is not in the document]
oxzi: As an administrator using Icinga DB, I might wanna know what timeout was exceeded as this might be a helpful debug information. Thus, as I just wrote in <https://github.com/Icinga/icingadb/pull/692#discussion_r1530070562>, this would only be `DEBUG`-logged.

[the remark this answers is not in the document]
oxzi: On second thoughts, this is not really helpful after all.

pkg/icingadb/ha.go
oxzi: I moved both down to `DEBUG`.

oxzi:
Additionally to addressing @yhabteab's comments, the error handling of the SQL query selecting Icinga DB instances was hanged slightly to also handle wrapped `sql.ErrNoRows` errors.
https://github.com/Icinga/icingadb/blob/4e4250e62aff276b7bc28b2ce2e70f5dba762682/pkg/icingadb/ha.go#L294

yhabteab (dismissed):
LFTM!  &#x2193;
```go
2024-03-19T14:31:26.547+0100    WARN    main    Handing over    {"reason": "context deadline exceeded"}
```

```diff
1c1
< 2024-03-19T12:36:45.209Z        INFO    high-availability       Another instance is active      {"instance_id": "7322e2b8658e446094c79c26b241af94", "environment": "e9698ca8a2cd336a55a948ea51903878d13869bd", "heartbeat": 1710851800743, "heartbeat_age": "4.466613808s"}
---
> 2024-03-19T13:52:55.135+0100    DEBUG   high-availability       Another instance is active      {"instance_id": "3525504452954c2eb7a2d4188c6cf01a", "environment": "e9698ca8a2cd336a55a948ea51903878d13869bd", "heartbeat": "2024-03-19T13:52:54.019+0100", "heartbeat_age": "1.116948947s"}
```

pkg/icingadb/ha.go
julianbrost: I think the idea for the "another instance is active" message to be repeated every now and then was to be some kind of heartbeat thing saying "I'm fine, it's intentional that not much is happening here". Do we still need this? Not entirely sure without trying how this behaves now.

However, if this stays at debug level, the whole `shouldLogRoutineEvents` thing can be removed. The only point of this is so that the regular log isn't spammed with these messages.

oxzi: Valid point. I removed the `shouldLog` (or as I renamed it as `shouldLogRoutineEvents`) logic.

yhabteab (approved):

lippserd (changes requested):
All in all, it's an improvement that the reasons are logged, but I don't like that the debug log now receives a lot of messages and that at least the first info message "another instance is active" is lost. Either we don't touch the levels and the timer altogether or we make proper logging for the four scenarios:

* takeover
* handover
* this is the active instance
* another instance is active

There must be at least one "another instance is active" message in the info logs. If this appears every minute, the user can quickly see that everything is fine. Also, I doubt that something like "this is the active instance" in the debug logs is helpful for debugging. Sure, if **nothing** is going on, then this at least shows that Icinga DB is not deadlocked or something, but then a timer is sufficient for that. This would be my suggestion:

* takeover: info, log reason
* handover: warn, log reason
* this is the active instance: debug, all x
* another instance is active: info, once and then all x

I don't mind to increase the timer to 5 minutes or something.

oxzi:
I guess a consensus is needed here. In an earlier review @yhabteab [commented](https://github.com/Icinga/icingadb/pull/692#discussion_r1524694282) that minutely logs are too much noise. However, I think the _"I am active"_ and the _"Someone else is active_" messages are being antagonistic and should be logged together, especially when a sysadmin wants to inspect what's going on. But I see your point, @lippserd, that it should be logged if someone else is being responsible.

Thus, I would alter the PR again to reintroduce the continuous logs with a bigger delay.

julianbrost (commented):
> By doing so, both the "Taking over" and "Handing over" log messages are enriched with a reason.

Well, not really in case of the taking over message given that there's only one call of `signalTakeover()` with the fixed reason "no other instance is active".

Most likely, that's due to the the fact that the current database query simply returns no rows in all takeover cases:

https://github.com/Icinga/icingadb/blob/2c468302ae9d196f1c2d4d4845830d2949a884aa/pkg/icingadb/ha.go#L261-L262

IIRC, one aspect we wanted to improve with the logging was to have distinct log messages based on whether there was a "graceful" takeover (i.e. the previously active instance removed its active state by itself) or a "forceful" takeover (i.e. the current instance took over because the previously active one seems to have died and its heartbeat expired).

To be able to log this, the checks on the `responsible` and `heartbeat` columns would have to be moved from the `where` clause to Go code. At first glance, this sounds feasible, so I think it's worth a try.

oxzi:
Valid point, @julianbrost. I have altered the commit and pushed a new version.

However, I don't see why the `responsible` column should be moved to the Go code. In the current logic only one row might be `responsible`, as during the takeover process an Icinga DB instances marks itself as `responsible` and all other instances as not `responsible` anymore within the same database transaction.
If I'm missing something, please show it to me.

Furthermore, I have introduced a grace period for a forceful takeover in case of an expired heartbeat, as otherwise one node might perform its hand over slightly after the other node's take over, as the following log snippet shows:

```
icingadb-master2  | 2024-03-27T11:58:59.280Z    INFO    icingadb        Taking over     {"reason": "other instance's heartbeat has expired"}
icingadb-master1  | 2024-03-27T11:58:59.304Z    WARN    icingadb        Handing over    {"reason": "lost heartbeat"}
```

oxzi:
As the old code indicated a takeover on the fact that no other instance is active, it will now additionally check if it is already being the active/responsible node. In this case, the takeover logic - which will be interrupted at a later point as the node is already responsible - can be skipped.

```diff
diff --git a/pkg/icingadb/ha.go b/pkg/icingadb/ha.go
index f3bcad1..5c62268 100644
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -302,11 +302,11 @@ func (h *HA) realize(
                                }

                        case errors.Is(errQuery, sql.ErrNoRows):
-                               takeover = "no other instance is active"
                                fields := []any{
                                        zap.String("instance_id", h.instanceId.String()),
                                        zap.String("environment", envId.String())}
                                if !h.responsible {
+                                       takeover = "no other instance is active"
                                        h.logger.Debugw("Preparing to take over HA as no instance is active", fields...)
                                } else if h.responsible && shouldLogRoutineEvents {
                                        h.logger.Debugw("Continuing being the active instance", fields...)
```

pkg/icingadb/ha.go
```diff
@@ -23,7 +23,12 @@ import (
 	"time"
 )
 
-var timeout = 60 * time.Second
+var (
+	// timeout for HA heartbeat.
+	timeout = 60 * time.Second
+	// timeoutGracePeriod is an additional grace period for the timeout to declare the other node timed out.
+	timeoutGracePeriod = 3 * time.Second
+)
```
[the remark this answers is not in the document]
oxzi: Good idea. I changed the HA timeout to be linked to the icingaredis Timeout, as you suggested. My different tests with both crashing and normally stopping components were looking good, same as with the previously introduced grace period. However, I prefer your idea as it links those two logically connected timeouts.

julianbrost (approved):

pkg/icingadb/ha.go
```diff
@@ -259,25 +275,41 @@ func (h *HA) realize(ctx context.Context, s *icingaredisv1.IcingaStatus, t *type
 			}
 
 			query := h.db.Rebind("SELECT id, heartbeat FROM icingadb_instance "+
-				"WHERE environment_id = ? AND responsible = ? AND id <> ? AND heartbeat > ?") + selectLock
```
lippserd, line 262: I haven't checked this thoroughly, but my first impression is that this can lead to unexpected rows when Icinga DB instances crash (repeatedly) in the `peerTimeout` timeframe. I think adding `ORDER BY heartbeat DESC` makes sense.

oxzi: Based on the takeover logic in Icinga DB, at most one node should be marked as `responsible` in the database. This selecting query, the upserting query setting this node's `responsible` field and the update query to mark everyone else as not `responsible` in case of a takeover is part of the same database transaction.

- https://github.com/Icinga/icingadb/blob/779afd1da35782cb164b3a296c201535bd4e170c/pkg/icingadb/ha.go#L281
- https://github.com/Icinga/icingadb/blob/779afd1da35782cb164b3a296c201535bd4e170c/pkg/icingadb/ha.go#L340
- https://github.com/Icinga/icingadb/blob/779afd1da35782cb164b3a296c201535bd4e170c/pkg/icingadb/ha.go#L346
- https://github.com/Icinga/icingadb/blob/779afd1da35782cb164b3a296c201535bd4e170c/pkg/icingadb/ha.go#L353

Thus, at most one row should be returned. Furthermore, for housekeeping there is the `removeOldInstances` method removing outdated entries.

While working on this PR, I have thought about opportunities with more then one `responsibility` Icinga DB instance and even temporary altered the code to allow working on multiple rows, but I was unable to enforce this behavior (unless I manually altered the database).

lippserd: Thanks for the extra work. You're right, it does look pretty solid.

lippserd (approved):

PR #696, opened 2024-03-14
tests/go.mod: bump github.com/icinga/icinga-testing "version"

Al2Klimov:
to include this fix:

* https://github.com/Icinga/icinga-testing/pull/29


oxzi (commented):
This change does not only bump `github.com/icinga/icinga-testing`, but almost every other package, too. However, those are not necessary being updated to their latest version. For example, `go.uber.org/zap` went from v1.21.0 to v1.26.0, while v1.27.0 is available, #680. Thus, maybe you just want to add another round of `go get -u ./...`?

Otherwise, the tests went through on my machine.

julianbrost:
What exactly does that mean? #695 will enable dependabot for `/tests`, I don't know if that will be enough, or if it can be configured to track a branch, or if that means we'd have to tag every merge commit in the `icinga-testing` repo.

Anyways, so far, I haven't merged #695 as it might result in PRs being created that would directly become obsoleted by this PR if it updates other dependencies as well.

IMHO, this PR should either update just `icinga-testing` or everything in `/tests`, not something in between. I have a slight preference towards just `icinga-testing` as that will be trivial to review and therefore should fix the GitHub Actions more quickly.

julianbrost (approved):
The version bump of icinga-testing also includes https://github.com/Icinga/icinga-testing/pull/26. Thus, this PR is actually the minimal change to do that bump as it all other module updates just reflect the changed dependencies of icinga-testing.

PR #698, opened 2024-03-18
Retry **every** database error

### LINKED ISSUE #618: Retention fails on very large databases
## Describe the bug

Bug is created from this Icingadb community thread:
https://community.icinga.com/t/icingadb-keeps-growing-to-absurd-size/12038/8

The icingadb is growing to no end, so I tried with retention for the databases:

```
# cat /etc/icingadb/config.yml |grep -v "# "

database:
  type: mysql
  host: localhost
  database: icingadb
  user: icingadb
  password: 

redis:
  host: localhost

logging:
  options:
    database: debug
    retention: debug

retention:
  history-days: 30
  sla-days: 30
  options:
    acknowledgement: 30
    comment: 30
    downtime: 30
    flapping: 30
    notification: 30
    state: 30
```

The database was to this time very large:
```
-rw-rw---- 1 mysql mysql  29G Jul 18 11:28 sla_history_state.ibd
-rw-rw---- 1 mysql mysql  34G Jul 18 11:28 state_history.ibd
-rw-rw---- 1 mysql mysql  52G Jul 18 11:28 history.ibd
[root@icinga icingadb]# ll -rSh /var/lib/mysql/icingadb/   
```

After a restart of of the icingadb.service I got from time to time this error and the icingadb service stops:

```
Jul 18 15:42:49 icinga icingadb[132922]: Error 1206: The total number of locks exceeds the lock table size
                                         can't perform "DELETE FROM sla_history_state WHERE environment_id = :environment_id AND event_time < :time\nORDER BY event_time LIMIT 5000"
                                         github.com/icinga/icingadb/pkg/icingadb.(*DB).CleanupOlderThan
                                                 github.com/icinga/icingadb/pkg/icingadb/cleanup.go:53
                                         github.com/icinga/icingadb/pkg/icingadb/history.(*Retention).Start.func1
                                                 github.com/icinga/icingadb/pkg/icingadb/history/retention.go:189
                                         github.com/icinga/icingadb/pkg/periodic.Start.func1
                                                 github.com/icinga/icingadb/pkg/periodic/periodic.go:78
                                         runtime.goexit
                                                 runtime/asm_amd64.s:1594
Jul 18 15:42:49 icinga systemd[1]: icingadb.service: Main process exited, code=exited, status=1/FAILURE
Jul 18 15:42:49 icinga systemd[1]: icingadb.service: Failed with result 'exit-code'.
```

## Your Environment

Include as many relevant details about the environment you experienced the problem in

    Icinga DB Web version: 1.0.2
    Icinga Web 2 version: 2.11.4
    Web browser: Chrome 113
    Icinga 2 version: r2.13.7-1
    Icinga DB version : v1.1.0
    PHP version used : 7.2.24
    Server operating system and version: Rocky Linux release 8.8 (Green Obsidian)

-------------------------------------------------

As a workaround I added those suggested parameters to my mariadb config:

```
# /etc/my.cnf.d/mariadb-server.cnf
[mysqld]
innodb_buffer_pool_chunk_size=536870912
innodb_buffer_pool_instances=4
innodb_buffer_pool_size=2147483648
```

### LINKED ISSUE #651: Crash after "Resource deadlock avoided"
## Describe the bug

IcingaDB daemon crashes after this error happens: "Resource deadlock avoided" (log attached).

## To Reproduce

1. Start icingadb service and wait the crash

## Expected behavior

I expect the service handle the error.

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Icinga DB version: 1.1.1-1+debian10
* Icinga 2 version: 2.14.0-1+debian10
* Operating System and version: Debian 10
* Database: MariaDB 10.3.39-0+deb10u1 (Galera cluster)

## Additional context

The database cluster has 3 nodes (the two Icinga masters, and another standby server for database redundancy). I run Icinga in this cluster with MariaDB since 2016, never had a problem with IDO. The problem started to happen as soon as I started icingadb for the first time. The error happens while the cluster is under low load.

The crashes happens on both masters.

- Error:
`icingadb[1660]: Error 1180 (HY000): Got error 35 "Resource deadlock avoided" during COMMIT#012can't perform "INSERT INTO \"history\" (\"event_time\", \"event_type\", \"id\", \"object_type\", \"host_id\", \"service_id\", \"state_history_id\", \"environment_id\", \"endpoint_id\") VALUES (:event_time,:event_type,:id,:object_type,:host_id,:service_id,:state_history_id,:environment_id,:endpoint_id) ON DUPLICATE KEY UPDATE \"id\" = VALUES(\"id\")"#012github.com/icinga/icingadb/internal.CantPerformQuery#012#011github.com/icinga/icingadb/internal/internal.go:30#012github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.1.1.1#012#011github.com/icinga/icingadb/pkg/icingadb/db.go:391#012github.com/icinga/icingadb/pkg/retry.WithBackoff#012#011github.com/icinga/icingadb/pkg/retry/retry.go:49#012github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.1.1#012#011github.com/icinga/icingadb/pkg/icingadb/db.go:386#012golang.org/x/sync/errgroup.(*Group).Go.func1#012#011golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75#012runtime.goexit#012#011runtime/asm_amd64.s:1598#012can't retry#012github.com/icinga/icingadb/pkg/retry.WithBackoff#012#011github.com/icinga/icingadb/pkg/retry/retry.go:68#012github.com/icinga/icingadb/pkg/icingadb
[icinga-log-deadlock.txt](https://github.com/Icinga/icingadb/files/12720309/icinga-log-deadlock.txt)
.(*DB).NamedBulkExec.func1.1.1#012#011github.com/icinga/icingadb/pkg/icingadb/db.go:386#012golang.org/x/sync/errgroup.(*Group).Go.func1#012#011golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75#012runtime.goexit#012#011runtime/asm_amd64.s:1598`

[icinga-log-deadlock.txt](https://github.com/Icinga/icingadb/files/12720317/icinga-log-deadlock.txt)

### LINKED ISSUE #677: Crash after "Too many connections"
## Describe the bug

The Icinga DB daemon dies if it cannot connect to the database due to "Too many connections". Manually restarting the daemon solves the issue.

While the root cause is the limited amount of connections, I believe Icinga DB should still not crash or die.

## Logs

```
Jan 26 13:39:30 localhost icingadb[1046308]: runtime-updates: Upserted 1 ServiceState items
Jan 26 13:40:29 localhost icingadb[1046308]: history-sync: Synced 1 state history items
Jan 26 13:40:30 localhost icingadb[1046308]: runtime-updates: Upserted 1 ServiceState items
Jan 29 00:03:16 localhost icingadb[1046308]: database: Can't connect to database. Retrying
Jan 29 00:03:16 localhost icingadb[1046308]: Error 1040 (08004): Too many connections
                                                            can't retry
                                                                    github.com/icinga/icingadb/pkg/retry/retry.go:68
                                                            github.com/icinga/icingadb/pkg/driver.RetryConnector.Connect
                                                                    github.com/icinga/icingadb/pkg/driver/driver.go:32
                                                            database/sql.(*DB).conn
                                                                    database/sql/sql.go:1387
                                                            database/sql.(*DB).exec
                                                                    database/sql/sql.go:1651
                                                            database/sql.(*DB).ExecContext.func1
                                                                    database/sql/sql.go:1634
                                                            database/sql.(*DB).retry
                                                                    database/sql/sql.go:1538
                                                            database/sql.(*DB).ExecContext
                                                                    database/sql/sql.go:1633
                                                            github.com/jmoiron/sqlx.NamedExecContext
                                                                    github.com/jmoiron/sqlx@v1.3.5/named_context.go:131
                                                            github.com/jmoiron/sqlx.(*DB).NamedExecContext
                                                                    github.com/jmoiron/sqlx@v1.3.5/sqlx_context.go:134
                                                                    github.com/icinga/icingadb/pkg/icingadb/db.go:389
                                                                    github.com/icinga/icingadb/pkg/retry/retry.go:49
                                                                    github.com/icinga/icingadb/pkg/icingadb/db.go:386
                                                                    golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75
                                                            runtime.goexit
                                                                    runtime/asm_amd64.s:1598
                                                            can't connect to database
                                                            github.com/icinga/icingadb/pkg/driver.RetryConnector.Connect
                                                                    github.com/icinga/icingadb/pkg/driver/driver.go:32
                                                            database/sql.(*DB).conn
                                                                    database/sql/sql.go:1387
                                                            database/sql.(*DB).exec
                                                                    database/sql/sql.go:1651
                                                            database/sql.(*DB).ExecContext.func1
                                                                    database/sql/sql.go:1634
                                                            database/sql.(*DB).retry
                                                                    database/sql/sql.go:1538
                                                            database/sql.(*DB).ExecContext
                                                                    database/sql/sql.go:1633
                                                            github.com/jmoiron/sqlx.NamedExecContext
                                                                    github.com/jmoiron/sqlx@v1.3.5/named_context.go:131
                                                            github.com/jmoiron/sqlx.(*DB).NamedExecContext
                                                                    github.com/jmoiron/sqlx@v1.3.5/sqlx_context.go:134
                                                                    github.com/icinga/icingadb/pkg/icingadb/db.go:389
                                                                    github.com/icinga/icingadb/pkg/retry/retry.go:49
                                                                    github.com/icinga/icingadb/pkg/icingadb/db.go:386
                                                                    golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75
                                                            runtime.goexit
                                                                    runtime/asm_amd64.s:1598
                                                            can't connect to database
                                                            github.com/icinga/icingadb/pkg/driver.RetryConnector.Connect
                                                                    github.com/icinga/icingadb/pkg/driver/driver.go:32
                                                            database/sql.(*DB).conn
                                                                    database/sql/sql.go:1387
                                                            database/sql.(*DB).exec
                                                                    database/sql/sql.go:1651
                                                            database/sql.(*DB).ExecContext.func1
                                                                    database/sql/sql.go:1634
                                                            database/sql.(*DB).retry
                                                                    database/sql/sql.go:1538
                                                            database/sql.(*DB).ExecContext
                                                                    database/sql/sql.go:1633
                                                            github.com/jmoiron/sqlx.NamedExecContext
                                                                    github.com/jmoiron/sqlx@v1.3.5/named_context.go:131
                                                            github.com/jmoiron/sqlx.(*DB).NamedExecContext
                                                                    github.com/jmoiron/sqlx@v1.3.5/sqlx_context.go:134
                                                                    github.com/icinga/icingadb/pkg/icingadb/db.go:389
                                                                    github.com/icinga/icingadb/pkg/retry/retry.go:49
                                                                    github.com/icinga/icingadb/pkg/icingadb/db.go:386
                                                                    golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75
                                                            runtime.goexit
                                                                    runtime/asm_amd64.s:1598
                                                            can't perform "INSERT INTO \"state_history\" (\"check_source\", \"endpoint_id\", \"object_type\", \"event_time\", \"long_output\", \"id\", \"check_attempt\", \"max_check_attempts\", \"host_id\", \"soft_state\", \"hard_state\", \"environment_id\", \"service_id\", \"state_type\", \"previous_hard_state\",>
                                                                    github.com/icinga/icingadb/pkg/icingadb/db.go:391
                                                                    github.com/icinga/icingadb/pkg/retry/retry.go:49
                                                                    github.com/icinga/icingadb/pkg/icingadb/db.go:386
                                                                    golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75
                                                            runtime.goexit
                                                                    runtime/asm_amd64.s:1598
                                                            can't retry
                                                                    github.com/icinga/icingadb/pkg/retry/retry.go:68
                                                                    github.com/icinga/icingadb/pkg/icingadb/db.go:386
                                                                    golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75
                                                            runtime.goexit
                                                                    runtime/asm_amd64.s:1598
Jan 29 00:03:16 localhost systemd[1]: icingadb.service: Main process exited, code=exited, status=1/FAILURE
Jan 29 00:03:16 localhost systemd[1]: icingadb.service: Failed with result 'exit-code'

```

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Icinga DB version: v1.1.1
* Icinga 2 version: r2.14.0-1
* Operating System and version: Ubuntu 20.04.6 LTS

lippserd:
This PR introduces several changes to our retention functionality:

* All functions except retention already retry queries, but there is no logging or timeout. Retry will now stop after 5 minutes and failed queries will be logged.
* So far we have kept a list of error codes that should be retried. This doesn't include all the errors that can be retried, and it doesn't even take into account errors that can occur in a database cluster. Instead of going through all possible error codes and checking [1] whether they should be included in the list of retryable errors, **every** database error is now simply retried. Of course, this also means retrying queries that can't be retried at all, but since we now give up after 5 minutes, that's okay.
[1] Based on a short error description from the vendor, it is difficult to tell for every error whether it can actually be retried without the context of when and how exactly such errors are triggered. There are also database clusters that send their own errors using vendor error codes.
* `DELETE` statements from retention are now also retried.

fixes #618
fixes #651
fixes #677

### DIFF
--- a/pkg/icingadb/cleanup.go
+++ b/pkg/icingadb/cleanup.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"github.com/icinga/icingadb/internal"
+	"github.com/icinga/icingadb/pkg/backoff"
 	"github.com/icinga/icingadb/pkg/com"
 	"github.com/icinga/icingadb/pkg/driver"
+	"github.com/icinga/icingadb/pkg/retry"
 	"github.com/icinga/icingadb/pkg/types"
+	"go.uber.org/zap"
 	"time"
 )
 
@@ -41,19 +44,50 @@ func (db *DB) CleanupOlderThan(
 	count uint64, olderThan time.Time, onSuccess ...OnSuccess[struct{}],
 ) (uint64, error) {
 	var counter com.Counter
-	defer db.log(ctx, stmt.Build(db.DriverName(), 0), &counter).Stop()
+	var n int64
+
+	q := db.Rebind(stmt.Build(db.DriverName(), count))
+
+	defer db.log(ctx, q, &counter).Stop()
 
 	for {
-		q := db.Rebind(stmt.Build(db.DriverName(), count))
-		rs, err := db.NamedExecContext(ctx, q, cleanupWhere{
-			EnvironmentId: envId,
-			Time:          types.UnixMilli(olderThan),
-		})
-		if err != nil {
-			return 0, internal.CantPerformQuery(err, q)
-		}
+		err := retry.WithBackoff(
+			ctx,
+			func(ctx context.Context) error {
+				rs, err := db.NamedExecContext(ctx, q, cleanupWhere{
+					EnvironmentId: envId,
+					Time:          types.UnixMilli(olderThan),
+				})
+				if err != nil {
+					return internal.CantPerformQuery(err, q)
+				}
 
-		n, err := rs.RowsAffected()
+				n, err = rs.RowsAffected()
+				if err != nil {
+					return err
+				}
+
+				return nil
+			},
+			retry.Retryable,
+			backoff.NewExponentialWithJitter(1*time.Millisecond, 1*time.Second),
+			retry.Settings{
+				Timeout: time.Minute * 5,
+				OnError: func(_ time.Duration, _ uint64, err, lastErr error) {
+					if lastErr == nil || err.Error() != lastErr.Error() {
+						db.logger.Warnw("Can't execute query. Retrying", zap.Error(err))
+					}
+				},
+				OnSuccess: func(elapsed time.Duration, attempt uint64, lastErr error) {
+					if attempt > 0 {
+						db.logger.Infow("Query retried successfully after error",
+							zap.Duration("after", elapsed),
+							zap.Uint64("attempts", attempt+1),
+							zap.NamedError("recovered_error", lastErr))
+					}
+				},
+			},
+		)
 		if err != nil {
 			return 0, err
 		}
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -8,7 +8,6 @@ import (
 	"github.com/lib/pq"
 	"github.com/pkg/errors"
 	"net"
-	"strings"
 	"syscall"
 	"time"
 )
@@ -141,43 +140,10 @@ func Retryable(err error) bool {
 		return true
 	}
 
-	var e *mysql.MySQLError
-	if errors.As(err, &e) {
-		switch e.Number {
-		case 1053, 1205, 1213, 2006:
-			// 1053: Server shutdown in progress
-			// 1205: Lock wait timeout
-			// 1213: Deadlock found when trying to get lock
-			// 2006: MySQL server has gone away
-			return true
-		default:
-			return false
-		}
-	}
-
-	var pe *pq.Error
-	if errors.As(err, &pe) {
-		switch pe.Code {
-		case "08000", // connection_exception
-			"08006", // connection_failure
-			"08001", // sqlclient_unable_to_establish_sqlconnection
-			"08004", // sqlserver_rejected_establishment_of_sqlconnection
-			"40001", // serialization_failure
-			"40P01", // deadlock_detected
-			"54000", // program_limit_exceeded
-			"55006", // object_in_use
-			"55P03", // lock_not_available
-			"57P01", // admin_shutdown
-			"57P02", // crash_shutdown
-			"57P03", // cannot_connect_now
-			"58000", // system_error
-			"58030", // io_error
-			"XX000": // internal_error
-			return true
-		default:
-			// Class 53 - Insufficient Resources
-			return strings.HasPrefix(string(pe.Code), "53")
-		}
+	var mye *mysql.MySQLError
+	var pqe *pq.Error
+	if errors.As(err, &mye) || errors.As(err, &pqe) {
+		return true
 	}
 
 	return false
--- a/pkg/icingadb/db.go
+++ b/pkg/icingadb/db.go
@@ -14,6 +14,7 @@ import (
 	"github.com/icinga/icingadb/pkg/utils"
 	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
+	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
 	"reflect"
@@ -338,7 +339,22 @@ func (db *DB) BulkExec(
 						},
 						retry.Retryable,
 						backoff.NewExponentialWithJitter(1*time.Millisecond, 1*time.Second),
-						retry.Settings{},
+						retry.Settings{
+							Timeout: time.Minute * 5,
+							OnError: func(_ time.Duration, _ uint64, err, lastErr error) {
+								if lastErr == nil || err.Error() != lastErr.Error() {
+									db.logger.Warnw("Can't execute query. Retrying", zap.Error(err))
+								}
+							},
+							OnSuccess: func(elapsed time.Duration, attempt uint64, lastErr error) {
+								if attempt > 0 {
+									db.logger.Infow("Query retried successfully after error",
+										zap.Duration("after", elapsed),
+										zap.Uint64("attempts", attempt+1),
+										zap.NamedError("recovered_error", lastErr))
+								}
+							},
+						},
 					)
 				}
 			}(b))
@@ -403,7 +419,22 @@ func (db *DB) NamedBulkExec(
 							},
 							retry.Retryable,
 							backoff.NewExponentialWithJitter(1*time.Millisecond, 1*time.Second),
-							retry.Settings{},
+							retry.Settings{
+								Timeout: time.Minute * 5,
+								OnError: func(_ time.Duration, _ uint64, err, lastErr error) {
+									if lastErr == nil || err.Error() != lastErr.Error() {
+										db.logger.Warnw("Can't execute query. Retrying", zap.Error(err))
+									}
+								},
+								OnSuccess: func(elapsed time.Duration, attempt uint64, lastErr error) {
+									if attempt > 0 {
+										db.logger.Infow("Query retried successfully after error",
+											zap.Duration("after", elapsed),
+											zap.Uint64("attempts", attempt+1),
+											zap.NamedError("recovered_error", lastErr))
+									}
+								},
+							},
 						)
 					}
 				}(b))
@@ -476,7 +507,22 @@ func (db *DB) NamedBulkExecTx(
 							},
 							retry.Retryable,
 							backoff.NewExponentialWithJitter(1*time.Millisecond, 1*time.Second),
-							retry.Settings{},
+							retry.Settings{
+								Timeout: time.Minute * 5,
+								OnError: func(_ time.Duration, _ uint64, err, lastErr error) {
+									if lastErr == nil || err.Error() != lastErr.Error() {
+										db.logger.Warnw("Can't execute query. Retrying", zap.Error(err))
+									}
+								},
+								OnSuccess: func(elapsed time.Duration, attempt uint64, lastErr error) {
+									if attempt > 0 {
+										db.logger.Infow("Query retried successfully after error",
+											zap.Duration("after", elapsed),
+											zap.Uint64("attempts", attempt+1),
+											zap.NamedError("recovered_error", lastErr))
+									}
+								},
+							},
 						)
 					}
 				}(b))


julianbrost:
It took me surprisingly long to answer a simple question: What will happen with simple configuration errors, like a wrong username for example. Answer: it will log a warning like this immediately and exit with the same error at a later point:

```
WARN    database        Can't connect to database. Retrying     {"error": "Error 1698 (28000): Access denied for user 'icingadb-test'@'localhost'"}
```

So if you did something wrong while setting it up, this PR shouldn't delay you from seeing the problem.

Why did it take me so long to test this? I wanted to test it within the docker image and the behavior totally didn't match my expectations (it logged that would retry but exited immediately afterwards). What happened? The docker image has an entrypoint which itself connects to the database and [this uses a fixed version of the Icinga DB code](https://github.com/Icinga/docker-icingadb/blob/506466c2320a0f223627da842f4a7a97f08e21ac/entrypoint/go.mod#L8). So long story short, that has to be updated as well, no matter in which way we change the error handling.

julianbrost:
I've created the release issue and added a corresponding point to the checklist there so that it won't be forgotten: #707

julianbrost:
> closes #679

I've added this to the PR description as that PR becomes obsolete if we do it this way.

> fixes #618

Not sure about that one. Yes, the issue mentions an "Lock wait timeout exceeded" error, but only after `retention.count` was lowered in the config. Before, the error was "The total number of locks exceeds the lock table size". I'm not sure if this would be fixed by retrying. I don't know what the limit is there, but if we get close to it with the default `retention.count = 5000`, maybe we should consider lowering the default.

yhabteab:
> I don't know what the limit is there, but if we get close to it with the default `retention.count = 5000`, maybe we should consider lowering the default.

I've just created an issue for this https://github.com/Icinga/icingadb/issues/717.

lippserd:
> > fixes #618
> 
> Not sure about that one. Yes, the issue mentions an "Lock wait timeout exceeded" error, but only after `retention.count` was lowered in the config. Before, the error was "The total number of locks exceeds the lock table size". I'm not sure if this would be fixed by retrying. I don't know what the limit is there, but if we get close to it with the default `retention.count = 5000`, maybe we should consider lowering the default.

Please see https://github.com/Icinga/icingadb/issues/717#issuecomment-2016951686.

lippserd:
Added commit ef34b7d6a12c1ffbf12af14c41fddd95f20097c3 to account for https://github.com/Icinga/icingadb/pull/686#issuecomment-1988441677.

PR is now ready for review.

yhabteab (commented):
The HA transaction retry handler still does not set a timeout and hangs forever without an error (if the same error is always returned and the debug log isn't used). Is it deliberate or missed to adjust it?

pkg/retry/retry.go
```diff
@@ -34,12 +36,13 @@ type Settings struct {
...
+	timeWindow, cancelTimeWindow := context.WithCancel(context.Background())
+	defer cancelTimeWindow()
 
 	if settings.Timeout > 0 {
-		var cancelCtx context.CancelFunc
-		ctx, cancelCtx = context.WithTimeout(ctx, settings.Timeout)
-		defer cancelCtx()
+		defer time.AfterFunc(settings.Timeout, func() {
+			cancelTimeWindow()
+		}).Stop()
```
Al2Klimov, line 45:
```suggestion
		defer time.AfterFunc(settings.Timeout, cancelTimeWindow).Stop()
```

lippserd:
> The HA transaction retry handler still does not set a timeout and hangs forever without an error (if the same error is always returned and the debug log isn't used). Is it deliberate or missed to adjust it?

I will have a look.

pkg/retry/retry.go
lippserd: Yep, that makes perfect sense.

pkg/retry/retry.go
```diff
@@ -34,12 +36,13 @@ type Settings struct {
...
 	}
```
julianbrost, line 46: `timeWindow` isn't really used like a context but more like a `time.Timer`. I think I'd change it to be that because otherwise, there is the question why it derives from `context.Background()` instead of `ctx`. Probably because otherwise the logging would have to be adapted to distinguish between both cancellation reasons. Using `time.Timer` instead would simply avoid the confusion.

```go
var timeout <-chan Time // receiving from nil channel blocks forever, acting like no timeout
if settings.Timeout > 0 {
	t := time.NewTimer(settings.Timeout)
	defer t.Stop()
	timeout = t.C
}

// ...
select {
// ...
case <-timeout:
```

pkg/retry/retry.go
lippserd: Done.

pkg/icingadb/cleanup.go
```diff
@@ -41,19 +44,50 @@ func (db *DB) CleanupOlderThan(
...
+					return internal.CantPerformQuery(err, q)
+				}
+
+				n, err = rs.RowsAffected()
+				if err != nil {
+					return err
+				}
 
-		n, err := rs.RowsAffected()
+				return nil
```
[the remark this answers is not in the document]
lippserd: Done.

pkg/retry/retry.go
lippserd: Done.

pkg/icingadb/cleanup.go
```diff
@@ -41,19 +44,47 @@ func (db *DB) CleanupOlderThan(
...
+
+				n, err = rs.RowsAffected()
 
-		n, err := rs.RowsAffected()
+				return err
+			},
+			retry.Retryable,
+			backoff.NewExponentialWithJitter(1*time.Millisecond, 1*time.Second),
+			retry.Settings{
+				Timeout: time.Minute * 5,
```
oxzi, line 72: The same `Timeout: time.Minute * 5` line now occurs five times within the `icingadb` package. Thus, maybe extract it into a const.

oxzi: Furthermore, it seems like the same `retry.Settings` block is used within cleanup.go and db.go.

pkg/icingadb/ha.go
```diff
@@ -337,6 +338,22 @@ func (h *HA) realize(ctx context.Context, s *icingaredisv1.IcingaStatus, t *type
 					log("Can't update or insert instance. Retrying", zap.Error(err), zap.Uint64("retry count", attempt))
 				}
 			},
+			OnSuccess: func(elapsed time.Duration, attempt uint64, lastErr error) {
+				if attempt > 0 {
+					log := h.logger.Debugw
+
+					if attempt > 3 {
+						// We log errors with severity info starting from the fourth attempt,
```
oxzi, line 346:
```suggestion
						// We log errors with severity info starting from the fifth attempt,
```

`attempt` is zero based. Otherwise, this might be the fourth retry.

pkg/icingadb/cleanup.go
```diff
@@ -41,19 +44,47 @@ func (db *DB) CleanupOlderThan(
 	count uint64, olderThan time.Time, onSuccess ...OnSuccess[struct{}],
 ) (uint64, error) {
 	var counter com.Counter
-	defer db.log(ctx, stmt.Build(db.DriverName(), 0), &counter).Stop()
+	var n int64
```
oxzi, line 47: Might be pedantic, but `n` can be moved down in the `for` loop's scope and maybe renamed to something more obvious at first glance, e.g., `currentRows`.

lippserd:
There is one more thing I have to in fix in ha:

> 2024-04-03T11:19:11.364+0200	DEBUG	high-availability	Can't update or insert instance. Retrying	{"error": "can't start transaction: can't connect to database: context deadline exceeded: dial tcp [::1]:3306: connect: connection refused"

We're using a deadline ctx for realize().

pkg/icingadb/ha.go
lippserd: The first line is about errors logs above where we log with severity info starting from the fourth attempt. Then attempt is increased and the success handler has to account that.

lippserd: And yes, it is based on zero. But when it comes to logging and writing, we start with the first attempt.

pkg/icingadb/cleanup.go
lippserd: That is true. I wanted to wait for input on timeouts and log messages before consolidating them.

julianbrost:
> We're using a deadline ctx for realize().

However, keep in mind that this deadline should probably keep existing, as it's based on the time when the heartbeat exprires:

https://github.com/Icinga/icingadb/blob/2826af4f8d99ff83fabec70ad51b3273b3bb05ca/pkg/icingadb/ha.go#L202-L206

Retrying the HA transaction with an outdated heartbeat would make no sense. And if it retries for longer than that, it may delay the forwarding of the heartbeat loss to the rest of the code.

And as I'm looking at that `if` right now, I'm not even sure if it makes sense that way. Shouldn't it use `WithDeadline` in any case, regardless of `h.responsible`? Wouldn't there otherwise be a chance that it attempts a takeover with an already timed out heartbeat if that happens to take more than a minute?

lippserd:
> However, keep in mind that this deadline should probably keep existing, as it's based on the time when the heartbeat exprires:

Yes, I didn't change the deadline but also used one for takeover scenarios. The issue was that `HA` didn't stop retrying at all.

pkg/icingadb/ha.go
```diff
@@ -228,6 +239,14 @@ func (h *HA) controller() {
...
 				h.realizeLostHeartbeat()
 			}
+
+			// Reset retry timeout so that the next iterations have the full amount of time available again.
+			// Don't be surprised by the location of the code,
+			// as it is obvious that the timer is also reset after an error that ends the loop anyway.
+			// But this is the best place to catch all scenarios where the timeout needs to be reset.
+			// And since HA needs quite a bit of refactoring anyway to e.g. return immediately after calling h.abort(),
+			// it's fine to have it here for now.
+			retryTimeout.Reset(retry.DefaultTimeout)
```
yhabteab, line 249: [`Timer#Reset()`](https://pkg.go.dev/time#Timer.Reset) emphasises that this receiver should only be called on an expired or stopped timer with drained channels, which is not the case here. Did you take these remarks into account when doing this here?

Apart from that, shouldn't the timeout be reset when entering this [code path](https://github.com/Icinga/icingadb/pull/698/files#diff-c07d29d7c4c9e56097cbdd2d4aba0dd00b2e13deac34ce4ab96d80ed63bbe821R176-R179) as well? Otherwise Icinga DB would simply crash after the timer was fired just because it *received a heartbeat from the past*.

oxzi: On first glance, it seems like the `time.Timer` can be replaced by a `time.Ticker`, which seems to be "reset-safe": <https://pkg.go.dev/time#Ticker.Reset>.

yhabteab: > Apart from that, shouldn't the timeout be reset when entering this [code path](https://github.com/Icinga/icingadb/pull/698/files#diff-c07d29d7c4c9e56097cbdd2d4aba0dd00b2e13deac34ce4ab96d80ed63bbe821R176-R179) as well? Otherwise Icinga DB would simply crash after the timer was fired just because it _received a heartbeat from the past_.

What about this part of my previous comment? Doesn't the ticker have to be reset in this case too?

lippserd: @oxzi This is the place where the timeout **must not** be reset.
@yhabteab You're right.

oxzi (commented):
While playing around with this PR, I simulated a MariaDB or network outage by blocking connections between Icinga DB and MariaDB through iptables. After less than one minute (~49 seconds) Icinga DB crashed due to a `dial tcp 172.24.0.8:3306: i/o timeout`.

```
2024-04-08T10:40:39.303Z        FATAL   icingadb        dial tcp 172.24.0.8:3306: i/o timeout
can't connect to database
github.com/icinga/icingadb/pkg/icingadb.RetryConnector.Connect
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/driver.go:42
database/sql.(*DB).conn
        /usr/local/go/src/database/sql/sql.go:1415
database/sql.(*DB).exec
        /usr/local/go/src/database/sql/sql.go:1679
database/sql.(*DB).ExecContext.func1
        /usr/local/go/src/database/sql/sql.go:1662
database/sql.(*DB).retry
        /usr/local/go/src/database/sql/sql.go:1566
database/sql.(*DB).ExecContext
        /usr/local/go/src/database/sql/sql.go:1661
github.com/jmoiron/sqlx.NamedExecContext
        /go/pkg/mod/github.com/jmoiron/sqlx@v1.3.5/named_context.go:131
github.com/jmoiron/sqlx.(*DB).NamedExecContext
        /go/pkg/mod/github.com/jmoiron/sqlx@v1.3.5/sqlx_context.go:134
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2.1
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/db.go:398
        /go/src/github.com/Icinga/icingadb/pkg/retry/retry.go:56
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/db.go:395
        /go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78
runtime.goexit
        /usr/local/go/src/runtime/asm_amd64.s:1695
        /go/src/github.com/Icinga/icingadb/internal/internal.go:30
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2.1
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/db.go:400
        /go/src/github.com/Icinga/icingadb/pkg/retry/retry.go:56
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/db.go:395
        /go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78
runtime.goexit
        /usr/local/go/src/runtime/asm_amd64.s:1695
exit status 1
```

It seems like "i/o timeout" is a bit historic in Go's net package. As it is implemented as the not exported [type `timeoutError`](https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/net/net.go;l=581-602) and is `errors.Is` equivalent to `context.DeadlineExceeded`, it gets a special handling:

https://github.com/Icinga/icingadb/blob/738e1e7e1236c32368eec537c402376fdb09518f/pkg/retry/retry.go#L68-L74

I am not quite sure, but I would expect Icinga DB to "survive" a relatively short MariaDB outage.

oxzi:
With the following improvised patch I have solved the issue mentioned in my [previous comment about "i/o timeout"](https://github.com/Icinga/icingadb/pull/698#pullrequestreview-1986074472):

```diff
diff --git a/pkg/retry/retry.go b/pkg/retry/retry.go
index 582f7c6..7974167 100644
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -8,6 +8,7 @@ import (
        "github.com/lib/pq"
        "github.com/pkg/errors"
        "net"
+       "strings"
        "syscall"
        "time"
 )
@@ -65,7 +66,10 @@ func WithBackoff(
                        settings.OnError(time.Since(start), attempt, err, prevErr)
                }

-               if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+               // Go's unexported net i/o timeout error passes as a context.DeadlineExceeded
+               // https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/net/net.go;l=581-602
+               isDeadlineExceeded := errors.Is(err, context.DeadlineExceeded) && !strings.Contains(err.Error(), "i/o timeout")
+               if isDeadlineExceeded || errors.Is(err, context.Canceled) {
                        if prevErr != nil {
                                err = errors.Wrap(err, prevErr.Error())
                        }
@@ -111,6 +115,11 @@ func Retryable(err error) bool {
                return true
        }

+       // https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/net/net.go;l=581-602
+       if errors.Is(err, context.DeadlineExceeded) && strings.Contains(err.Error(), "i/o timeout") {
+               return true
+       }
+
        var dnsError *net.DNSError
        if errors.As(err, &dnsError) {
                return true
```

julianbrost:
Not completely nice, but maybe a bit nicer than comparing error strings: `errors.Is(err, context.DeadlineExceeded) && errors.Is(ctx.Err(), context.DeadlineExceeded)`, i.e. double check that the context actually returns an deadline error as well.

lippserd:
> Not completely nice, but maybe a bit nicer than comparing error strings: `errors.Is(err, context.DeadlineExceeded) && errors.Is(ctx.Err(), context.DeadlineExceeded)`, i.e. double check that the context actually returns an deadline error as well.

What about `errors.As()`?

lippserd:
Since we pass `ctx` as is and check for context errors at this point, we could simply add `ctx.Err() != nil` or use `ctx.Err()` instead of `err`.

oxzi:
> Not completely nice, but maybe a bit nicer than comparing error strings: `errors.Is(err, context.DeadlineExceeded) && errors.Is(ctx.Err(), context.DeadlineExceeded)`, i.e. double check that the context actually returns an deadline error as well.

This approach by double checking against the `ctx` seems to be the most reasonable as there are other possible ambiguities, e.g., `net.errCanceled` as `context.Canceled`. Thus, the check within the `WithBackoff` functions boils down to check if the `ctx` is done. Or am I missing something?

I have now successfully tested the following diff:

```diff
diff --git a/pkg/retry/retry.go b/pkg/retry/retry.go
index 582f7c6..6049a7b 100644
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -65,7 +65,8 @@ func WithBackoff(
                        settings.OnError(time.Since(start), attempt, err, prevErr)
                }

-               if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+               if ctx.Err() != nil {
+                       err = ctx.Err()
                        if prevErr != nil {
                                err = errors.Wrap(err, prevErr.Error())
                        }
@@ -111,6 +112,13 @@ func Retryable(err error) bool {
                return true
        }

+       if errors.Is(err, context.DeadlineExceeded) {
+               // Errors from the standard library might be "errors.Is" equivalent to context.DeadlineExceeded.
+               // https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/net/net.go;l=581-602
+               // The used context was already checked before within WithBackoff.
+               return true
+       }
+
        var dnsError *net.DNSError
        if errors.As(err, &dnsError) {
                return true
```

I am unsure if a `context.Canceled` should also be handled as retryable. This somehow feels wrong.

yhabteab:
> I am unsure if a `context.Canceled` should also be handled as retryable. This somehow feels wrong.

Not `context#context.Canceled`, but `net#context.Canceled` should definitely be retried, as it is just a child context created by the Dialer/driver#Connector that was being cancelled there.

[1] https://github.com/Icinga/icingadb/blob/main/pkg/config/database.go#L59
[2] https://github.com/go-sql-driver/mysql/blob/master/connector.go#L93-L104
[3] https://github.com/golang/go/blob/master/src/net/dial.go#L478
[4] https://github.com/golang/go/blob/master/src/net/dial.go#L613-L614

However, I think this is not directly related to this PR and should be addressed separately in its own PR.

lippserd:
> > I am unsure if a `context.Canceled` should also be handled as retryable. This somehow feels wrong.
> 
> Not `context#context.Canceled`, but `net#context.Canceled` should definitely be retried, as it is just a child context created by the Dialer/driver#Connector that was being cancelled there.
> 
> [1] https://github.com/Icinga/icingadb/blob/main/pkg/config/database.go#L59 [2] https://github.com/go-sql-driver/mysql/blob/master/connector.go#L93-L104 [3] https://github.com/golang/go/blob/master/src/net/dial.go#L478 [4] https://github.com/golang/go/blob/master/src/net/dial.go#L613-L614
> 
> However, I think this is not directly related to this PR and should be addressed separately in its own PR.

Fixing this is subject to this PR as I've changed the context error handling with 44f18fe8e369a525461e4161a8adade15e62112e.

> **EDIT:**
> 
> Given that all errors returned by the `net` package are of type `net#OpError`, which already implements the `Unwrap()` receiver, this should be removed in the new PR [as the justification for doing so was never?/is no longer valid](https://github.com/golang/go/blob/master/src/net/net.go#L470).

That is a good change for a new PR.

I pushed two new commits, please have a look. I simply check `ctx.Err()` instead of the returned `err` and added justification. @oxzi Just for the record, in your examples you are doing `err = ctx.Err()` which overrides the returned error and shouldn't be done as `err` might be a wrapped context error.

oxzi:
Regarding @yhabteab's [comment](https://github.com/Icinga/icingadb/pull/698#issuecomment-2043035095):

> > I am unsure if a `context.Canceled` should also be handled as retryable. This somehow feels wrong.
> 
> Not `context#context.Canceled`, but `net#context.Canceled` should definitely be retried, as it is just a child context created by the Dialer/driver#Connector that was being cancelled there.

If I'm understanding you correctly, this is what @julianbrost [first suggested](https://github.com/Icinga/icingadb/pull/698#issuecomment-2042674129) and was eventually committed in <https://github.com/Icinga/icingadb/pull/698/commits/b2fd89f7de91789d4bd806d54695ffea464796d3>.

> Given that all errors returned by the `net` package are of type `net#OpError`

First, thanks for your analysis of the call stack. However, I'm not sure if this statement is true. The `net` package [exports other errors](https://pkg.go.dev/net#pkg-variables), e.g., `ErrWriteToConnected`. Furthermore, the documentation for [`OpError`](https://pkg.go.dev/net#OpError) states uncertain that it is "the error type usually returned by functions in the net package".

---

Regarding @lippserd's [comment](https://github.com/Icinga/icingadb/pull/698#issuecomment-2044225078):

> I pushed two new commits, please have a look. I simply check `ctx.Err()` instead of the returned `err` and added justification.

I am only seeing one commit, b2fd89f7de91789d4bd806d54695ffea464796d3. Am I missing something?

> @oxzi Just for the record, in your examples you are doing `err = ctx.Err()` which overrides the returned error and shouldn't be done as `err` might be a wrapped context error.

Valid point.

Regarding your last commit, b2fd89f7de91789d4bd806d54695ffea464796d3, and my objections about `net.OpError` stated above, I would suggest handling `context.DeadlineExceeded` as a valid case for `Retryable()`, as I did in my [last posted diff](https://github.com/Icinga/icingadb/pull/698#issuecomment-2042790039). This might - and only might - catch future or other nested library errors, also from other packages next to `net`.

lippserd:
> Regarding @lippserd's [comment](https://github.com/Icinga/icingadb/pull/698#issuecomment-2044225078):
> 
> > I pushed two new commits, please have a look. I simply check `ctx.Err()` instead of the returned `err` and added justification.
> 
> I am only seeing one commit, [b2fd89f](https://github.com/Icinga/icingadb/commit/b2fd89f7de91789d4bd806d54695ffea464796d3). Am I missing something?

I also pushed 2344e0a105e74236de00b8f863cdaa9227d12e29 but it's about logging.

> Regarding your last commit, [b2fd89f](https://github.com/Icinga/icingadb/commit/b2fd89f7de91789d4bd806d54695ffea464796d3), and my objections about `net.OpError` stated above, I would suggest handling `context.DeadlineExceeded` as a valid case for `Retryable()`, as I did in my [last posted diff](https://github.com/Icinga/icingadb/pull/698#issuecomment-2042790039). This might - and only might - catch future or other nested library errors, also from other packages next to `net`.

Errors that claim to be `context.DeadlineExceeded` will be retried if they implement `Tempory() == true` or `Timeout() == true`. That was also the case before. I wanted to exit early upon context errors: 44f18fe8e369a525461e4161a8adade15e62112e - please also see the commit description. But not for errors that claim to be context errors but real ones. This commit introduced the bug that the mentioned `net` errors are no longer retried. Which is fixed by checking the context for errors and not the returned error from the retryable funcion. The deadline of the context we're passing in `HA` for example **must not** be considered retryable if it passes.

pkg/retry/retry.go
```diff
@@ -54,43 +61,46 @@ func WithBackoff(
 			return
```
oxzi, line 61: This comment addresses the `for` loop header line, but GitHub doesn't allow me to click there..

I would suggest incrementing `attempt`'s initial value to `1`.
- This would ease the attempt number usage and reduce off-by-ones, e.g., here:
  https://github.com/Icinga/icingadb/blob/2344e0a105e74236de00b8f863cdaa9227d12e29/pkg/icingadb/db.go#L686
- Or here:
  https://github.com/Icinga/icingadb/blob/2344e0a105e74236de00b8f863cdaa9227d12e29/pkg/icingadb/ha.go#L404-L406
- Furthermore, this would remove the directly firing `time.After` case:
  https://github.com/Icinga/icingadb/blob/b2fd89f7de91789d4bd806d54695ffea464796d3/pkg/retry/retry.go#L88
  https://github.com/Icinga/icingadb/blob/61f8e7f915c47b10036302d709b8254ef9e622c6/pkg/backoff/backoff.go#L37-L40

yhabteab (dismissed):
Works fine!

```bash
2024-04-09T08:35:23.765Z        WARN    redis   Can't connect to Redis. Retrying        {"error": "dial tcp 45.76.93.104:6380: connect: connection refused"}
---
2024-04-09T08:36:03.675Z        WARN    redis   Can't connect to Redis. Retrying        {"error": "dial tcp 45.76.93.104:6380: i/o timeout"}
---
2024-04-09T08:36:20.054Z        WARN    telemetry       Can't update own heartbeat      {"error": "can't perform \"[xadd icingadb:telemetry:heartbeat maxlen 1 * error-since 0 ha-responsible 1 ha-responsible-ts 17...\": can't connect to Redis: dial tcp: lookup redis: i/o timeout", "errorVerbose": "dial tcp: lookup redis: i/o timeout\ncan't connect...
---
2024-04-09T08:36:20.916Z        WARN    database        Can't execute query. Retrying   {"error": "can't perform \"INSERT INTO \\\"notification_history\\\"...
---
2024-04-09T08:36:20.920Z        WARN    database        Can't connect to database. Retrying     {"error": "pq: the database system is shutting down"}
2024-04-09T08:36:21.173Z        WARN    database        Can't connect to database. Retrying     {"error": "read tcp 172.20.0.5:40236->172.20.0.6:5432: read: connection reset by peer"}
---
2024-04-09T08:40:03.426Z        INFO    database        Query retried successfully after error  {"after": "3m42.519639231s", "attempts": 2, "recovered_error": "can't perform \"INSERT INTO \\\"notification_history\\\"...
---
2024-04-09T08:40:29.145Z        WARN    database        Can't connect to database. Retrying     {"error": "dial tcp 45.76.93.104:5432: i/o timeout"}
---
```

There may, however, be some retry races as @julianbrost pointed out, that's not easy to fix, since the timeout of the different goroutines varies and IcingDB may crash despite successful retries some of the queries!

```bash
2024-04-09T08:41:22.820Z        FATAL   icingadb        dial tcp 45.76.93.104:5432: i/o timeout
retry deadline exceeded
        /go/src/github.com/Icinga/icingadb/pkg/retry/retry.go:90
github.com/icinga/icingadb/pkg/icingadb.RetryConnector.Connect
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/driver.go:42
database/sql.(*DB).conn
        /usr/local/go/src/database/sql/sql.go:1415
database/sql.(*DB).exec
        /usr/local/go/src/database/sql/sql.go:1679
database/sql.(*DB).ExecContext.func1
        /usr/local/go/src/database/sql/sql.go:1662
database/sql.(*DB).retry
        /usr/local/go/src/database/sql/sql.go:1566
database/sql.(*DB).ExecContext
        /usr/local/go/src/database/sql/sql.go:1661
github.com/jmoiron/sqlx.NamedExecContext
        /go/pkg/mod/github.com/jmoiron/sqlx@v1.3.5/named_context.go:131
github.com/jmoiron/sqlx.(*DB).NamedExecContext
        /go/pkg/mod/github.com/jmoiron/sqlx@v1.3.5/sqlx_context.go:134
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2.1
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/db.go:398
        /go/src/github.com/Icinga/icingadb/pkg/retry/retry.go:56
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/db.go:395
        /go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78
runtime.goexit
        /usr/local/go/src/runtime/asm_arm64.s:1222
can't connect to database
github.com/icinga/icingadb/pkg/icingadb.RetryConnector.Connect
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/driver.go:42
database/sql.(*DB).conn
        /usr/local/go/src/database/sql/sql.go:1415
database/sql.(*DB).exec
        /usr/local/go/src/database/sql/sql.go:1679
database/sql.(*DB).ExecContext.func1
        /usr/local/go/src/database/sql/sql.go:1662
database/sql.(*DB).retry
        /usr/local/go/src/database/sql/sql.go:1566
database/sql.(*DB).ExecContext
        /usr/local/go/src/database/sql/sql.go:1661
github.com/jmoiron/sqlx.NamedExecContext
        /go/pkg/mod/github.com/jmoiron/sqlx@v1.3.5/named_context.go:131
github.com/jmoiron/sqlx.(*DB).NamedExecContext
        /go/pkg/mod/github.com/jmoiron/sqlx@v1.3.5/sqlx_context.go:134
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2.1
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/db.go:398
        /go/src/github.com/Icinga/icingadb/pkg/retry/retry.go:56
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/db.go:395
        /go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78
runtime.goexit
        /usr/local/go/src/runtime/asm_arm64.s:1222
can't perform "INSERT INTO \"state_history\" (\"id\", \"hard_state\", \"scheduling_source\", \"environment_id\", \"object_type\", \"soft_state\", \"long_output\", \"endpoint_id\", \"event_time\", \"check_source\", \"host_id\", \"max_check_attempts\", \"service_id\", \"state_type\", \"previous_soft_state\", \"previous_hard_state\", \"check_attempt\", \"output\") VALUES (:id,:hard_state,:scheduling_source,:environment_id,:object_type,:soft_state,:long_output,:endpoint_id,:event_time,:check_source,:host_id,:max_check_attempts,:service_id,:state_type,:previous_soft_state,:previous_hard_state,:check_attempt,:output) ON CONFLICT ON CONSTRAINT pk_state_history DO UPDATE SET \"id\" = EXCLUDED.\"id\""
        /go/src/github.com/Icinga/icingadb/internal/internal.go:30
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2.1
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/db.go:400
        /go/src/github.com/Icinga/icingadb/pkg/retry/retry.go:56
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/db.go:395
        /go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78
runtime.goexit
        /usr/local/go/src/runtime/asm_arm64.s:1222
retry deadline exceeded
        /go/src/github.com/Icinga/icingadb/pkg/retry/retry.go:90
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2
        /go/src/github.com/Icinga/icingadb/pkg/icingadb/db.go:395
        /go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78
runtime.goexit
        /usr/local/go/src/runtime/asm_arm64.s:1222
exit status 1
```

oxzi (commented):
This PR works for me and looks kinda good. Thanks!

If you might consider my comment regarding the `attempt` index variable in this PR's scope, feel free to address it. Otherwise, a new PR can be opened.

pkg/retry/retry.go
lippserd: Sounds good. I will do that.

pkg/icingadb/driver.go
```diff
@@ -55,8 +55,8 @@ func (c RetryConnector) Connect(ctx context.Context) (driver.Conn, error) {
 		shouldRetry,
 		backoff.NewExponentialWithJitter(time.Millisecond*128, time.Minute*1),
```
julianbrost, line 56: With the nesting of the `retry.WithBackoff()` within those in other places that execute SQL queries and these backoff settings, some strange effects can be observed: If you stop your database for something like 4m30s, you may see some attempts recovering and others - quite some time after that - failing fatally because those exceeded the 5 minutes. I believe this happens to the maximal backoff of a minute given here, i.e. there may be a minute without an attempt and after that, both the inner and outer 5 minute timeout are exceeded.

Also, if we're adding `retry.WithBackoff()`, do we even still need it here? Just for fun, I removed all the retry from this function (so that it just does `Connect()` and `initConn()` with no loops or anything around it) and it seemed to work more reliably, only the logging became very annoying with queries and stacktraces within a single line.

lippserd: The following changes address your considerations:

`HA` now uses a `Timer` instead of a `Ticker` that is reset properly even if expired (and not yet drained): 81085c04a42a67527559a611f8129af26afbfb13.

If the timeout expires during the sleep phase between attempts one final retry attempt will be made: 3a3baafc86064f398e98972f71ff9bd709e7149f.

> only the logging became very annoying with queries and stacktraces within a single line

I have something in my stash which fixes this. `zap` adds the `errorVerbose` field if the error implements `fmt.Formatter` which is true for all our wrapped errors. Simple solution would be to replace `zap.Error()` calls with a custom implementation that returns a "silent" error, e.g.

```go
package logging

...

type stackTracer interface {
	StackTrace() errors.StackTrace
}

type errNoStackTrace struct {
	e error
}

func (e errNoStackTrace) Error() string {
	return e.e.Error()
}

func Error(e error) zap.Field {
	if _, ok := e.(stackTracer); ok {
		return zap.Error(errNoStackTrace{e})
	}

	return zap.Error(e)
}
```

julianbrost: > > only the logging became very annoying with queries and stacktraces within a single line
> 
> I have something in my stash which fixes this. `zap` adds the `errorVerbose` field if the error implements `fmt.Formatter` which is true for all our wrapped errors. Simple solution would be to replace `zap.Error()` calls with a custom implementation that returns a "silent" error, e.g.

However, the PR just keeps the nested retrying for now, thereby avoiding this issue.

julianbrost (commented):
No more complaints from my side. Compared to the others, I haven't spent much time looking into the details of this PR so far, so I think it would be good to get another approval on this as well.

yhabteab (approved):
```bash

2024-04-11T09:43:21.347+0200    INFO    database        Reconnected to database {"after": "4m50.284240542s", "attempts": 14}
2024-04-11T09:43:25.079+0200    INFO    database        Reconnected to database {"after": "4m44.25903525s", "attempts": 14}
2024-04-11T09:43:28.895+0200    INFO    database        Reconnected to database {"after": "2m45.294415s", "attempts": 12}
2024-04-11T09:43:29.917+0200    INFO    database        Reconnected to database {"after": "4m28.553121625s", "attempts": 15}
2024-04-11T09:43:31.347+0200    INFO    database        Reconnected to database {"after": "4m54.634997417s", "attempts": 15}
2024-04-11T09:43:33.639+0200    INFO    database        Reconnected to database {"after": "5m0.007767875s", "attempts": 15}
2024-04-11T09:43:34.152+0200    INFO    database        Reconnected to database {"after": "5m0.007966958s", "attempts": 15}
2024-04-11T09:43:34.820+0200    INFO    database        Reconnected to database {"after": "3m53.386925458s", "attempts": 13}
2024-04-11T09:43:34.922+0200    INFO    database        Reconnected to database {"after": "5m0.006823084s", "attempts": 15}
2024-04-11T09:43:38.775+0200    INFO    database        Reconnected to database {"after": "5m0.00952125s", "attempts": 15}
2024-04-11T09:43:39.100+0200    INFO    history-sync    Synced 622 state history items
2024-04-11T09:43:39.100+0200    INFO    history-sync    Synced 495 notification history items
2024-04-11T09:43:39.272+0200    INFO    runtime-updates Upserted 17 ServiceState items
2024-04-11T09:43:39.846+0200    INFO    database        Reconnected to database {"after": "4m53.632643916s", "attempts": 15}
2024-04-11T09:43:41.586+0200    INFO    database        Reconnected to database {"after": "3m1.067891625s", "attempts": 12}
2024-04-11T09:43:46.617+0200    INFO    database        Reconnected to database {"after": "4m20.077357709s", "attempts": 13}
2024-04-11T09:43:51.625+0200    INFO    database        Reconnected to database {"after": "4m5.054033166s", "attempts": 13}
2024-04-11T09:43:55.058+0200    INFO    database        Reconnected to database {"after": "4m18.504347416s", "attempts": 14}
```

PR #700, opened 2024-03-18
MySQL/MariaDB: Fix `icingadb_schema.timestamp` not being UNX time

lippserd:
We previously incorrectly used `CURRENT_TIMESTAMP()` an alias for `NOW()` which returns the current date and time in the format `YYYY-MM-DD hh:mm:ss` using the session time zone. Since we are using numeric context, the value is stored in the format `YYYYMMDDhhmmss`. But actually we want to set a (millisecond) UNIX timestamp here, so we need to use `UNIX_TIMESTAMP()` instead.

Since the timestamps need to be corrected, there is also a schema upgrade.

### DIFF
--- a/schema/mysql/upgrades/1.1.2.sql
+++ b/schema/mysql/upgrades/1.1.2.sql
@@ -1,0 +1,1 @@
+UPDATE icingadb_schema SET timestamp = UNIX_TIMESTAMP(timestamp / 1000) * 1000 WHERE timestamp > 20000000000000000;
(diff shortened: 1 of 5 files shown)


schema/mysql/upgrades/1.1.2.sql
```diff
@@ -0,0 +1 @@
+UPDATE icingadb_schema SET timestamp = UNIX_TIMESTAMP(timestamp / 1000) * 1000;
```
julianbrost, line 1: I'd add something like `WHERE timestamp > 20000000000000000` here for two reasons:

1. While we don't have idempotency as a requirement for schema upgrades, they probably shouldn't corrupt the data when accidentally applied twice (most other upgrades would probably bail out with a "something already exists" error).
2. Especially as you changed past schema upgrade files as well, if you'd now apply them in order, there would be broken timestamps left in this table.

That value is far enough in the future (after dividing by 1000), there won't be problems any time soon:
```console
date -ud@20000000000000.0000
Thu 08 Oct 635744 11:33:20 UTC
```

julianbrost (approved):

PR #711, opened 2024-03-19
Try setting `wsrep_sync_wait` for mysql connections

yhabteab:
This PR is an alternative version of https://github.com/Icinga/icingadb/pull/665 and utilises what is described in https://github.com/Icinga/icingadb/pull/665#discussion_r1530546783.

Using this PR avoids the need for custom driver registration and also eliminates the need for any other custom connectors/drivers that aren't of type `driver.RetryConnector`. This PR also adds an option to `driver.RetryConnector` to post initialise a connection after `Connector#Connect()`.

closes #665


pkg/driver/driver.go
```diff
@@ -43,56 +47,43 @@ func (c RetryConnector) Connect(ctx context.Context) (driver.Conn, error) {
...
-	if err != nil {
-		return nil, err
-	}
-
-	return &RetryConnector{
-		driver:    d,
-		Connector: c,
-	}, nil
+	return &ctxDriver{}
 }
```
julianbrost, line 77: The `Driver()` method doesn't look right. If it would ever be used, the result would be useless.

Maybe keep the `Driver` type, store the `InitConn` callback there and use it for every connection created from that driver.

pkg/driver/driver.go
```diff
@@ -43,56 +47,43 @@ func (c RetryConnector) Connect(ctx context.Context) (driver.Conn, error) {
...
 			},
 		},
 	), "can't connect to database")
+
+	if err == nil && c.InitConn != nil {
+		err = c.InitConn(ctx, conn)
+		if err != nil {
+			return nil, err
+		}
+	}
```
julianbrost, line 69: What happens if this returns something like a network error? Shouldn't this happen within the retry just above?

pkg/config/database.go
```diff
@@ -18,7 +22,10 @@ import (
 	"time"
 )
 
-var registerDriverOnce sync.Once
+var (
+	registerDriverOnce sync.Once
+	wsrepSyncWait      int64
```
julianbrost, line 27: Why is this a global variable? Wasn't the whole point of passing it in the DSN or using `NewConnector()` and wrapping the result into a `*sqlx.DB` to have this option scoped to the `*DB`?

julianbrost (changes requested):

pkg/driver/driver.go
yhabteab: > The `Driver()` method doesn't look right. If it would ever be used, the result would be useless.

[`PostgreSQL`](https://github.com/lib/pq/blob/master/connector.go#L36-L38) does exactly the same thing.

pkg/config/database.go
yhabteab: Do you mean scoped to `*config.Database`? Otherwise it doesn't make sense to add this to the actual `DB` type.

pkg/driver/driver.go
julianbrost: Not really the same.

`pq.Driver` has a useful implementation of the `Driver` on a zero-initialized value: https://github.com/lib/pq/blob/3d613208bca2e74f2a20e04126ed30bcb5c4cc27/conn.go#L50-L58

`ctxDriver` on the other hand just embeds two interfaces. So if a method on a zero-initialized value is called, this will just result in a nil derefence.

pkg/config/database.go
julianbrost: I wanted to say so that it's possible to have multiple `*icingadb.DB` that could theoretically use different values for `wsrep_sync_wait`. For the Icinga DB daemon, this doesn't really mater, but you just have to look at the history migration tool where you have to configure an IDO connection as well using the same config, so already there you could write config that results in strange behavior if a global variable would be used.

And fixing this wouldn't even be hard with the current code. Instead of setting a global variable, just use an anonymous function for `InitConn` that captures the value.

pkg/driver/driver.go
```diff
@@ -15,15 +14,22 @@ import (
 	"time"
 )
 
-const MySQL = "icingadb-mysql"
-const PostgreSQL = "icingadb-pgsql"
+const MySQL = "mysql"
+const PostgreSQL = "pgsql"
```
julianbrost, line 18: Previously, those were custom driver names so that we could use `sqlx.Open()`. Now, only one matches the driver name we use in the background, the other doesn't. Also note that if we simply tell sqlx that our driver is `"postgres"`, we don't even need the `sqlx.BindDriver()` call.

I'd change this to `PostgreSQL = "postgres"` (or maybe we don't need them at all, they might already exist as constants in the driver packages?).

Also, these two cases elsewhere now simply look like something was forgotten:
```go
case driver.MySQL:
case driver.PostgreSQL, "postgres":
```

julianbrost (commented):
Before I go too deep into implementation details: @lippserd Do you see any issue in removing registering the named drivers in the `sql` package?

Apart from that, just one comment of something I found strange so far.

pkg/driver/driver.go
yhabteab: > or maybe we don't need them at all, they might already exist as constants in the driver packages?

I couldn't find such constants neither in [`pq`](https://pkg.go.dev/github.com/lib/pq#pkg-constants) nor in [`mysql`](https://pkg.go.dev/github.com/go-sql-driver/mysql#pkg-constants).

lippserd: Sure they do:

`./github.com/go-sql-driver/mysql@v1.8.0/driver.go:96: sql.Register(driverName, &MySQLDriver{})`

`./github.com/lib/pq@v1.1.1/conn.go:52: sql.Register("postgres", &Driver{})`

But [pgx](https://github.com/jackc/pgx) registers as `pgx`.

lippserd: I would also try to get rid of the constants.

lippserd: Sorry, I thought you were simply talking about the driver names! Still, I think relying on `mysql` and `postgres` as strings is good enough.

yhabteab: As discussed offline, we opt to keep the constants in order to maintain the driver names in a centralized place.

lippserd:
> Before I go too deep into implementation details: @lippserd Do you see any issue in removing registering the named drivers in the `sql` package?

Nope, removing them is fine.

pkg/driver/driver.go
```diff
@@ -61,40 +75,12 @@ func (c RetryConnector) Connect(ctx context.Context) (driver.Conn, error) {
...
-		Connector: c,
-	}, nil
-}
-
-// Register makes our database Driver available under the name "icingadb-*sql".
+// Register sets the default mysql logger to the given one.
 func Register(logger *logging.Logger) {
-	sql.Register(MySQL, &Driver{ctxDriver: &mysql.MySQLDriver{}, Logger: logger})
-	sql.Register(PostgreSQL, &Driver{ctxDriver: PgSQLDriver{}, Logger: logger})
 	_ = mysql.SetLogger(mysqlLogger(func(v ...interface{}) { logger.Debug(v...) }))
```
lippserd, line 83: This could also be moved to the `mysql` case of `Database#Open()`. It's not an issue to call this method multiple times (with the same logger of course). Then we could move the constants and `RetryConnector` to the database package as well. The `driver` package really is empty now and we wouldn't have to do stuff like `icingaDriver` or `sqlDriver` and could remove `registerDriverOnce`.

pkg/config/database.go
```diff
@@ -75,7 +81,46 @@ func (d *Database) Open(logger *logging.Logger) (*icingadb.DB, error) {
...
+		// Setting "wsrep_sync_wait" for each session ensures that causality checks are performed before execution
+		// and that each statement is executed on a fully synchronized node. Doing so prevents foreign key violation
+		// when inserting into dependent tables on different MariaDB/MySQL nodes. When using MySQL single nodes, the
+		// "SET SESSION" command will fail with "Unknown system variable (1193)" and will therefore be silently dropped.
+		// https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait
+		var setGaleraOpts = func(ctx context.Context, conn driver.Conn) error {
+			const galeraOpts = "SET SESSION wsrep_sync_wait=?"
+
+			stmt, err := conn.(driver.ConnPrepareContext).PrepareContext(ctx, galeraOpts)
+			if err != nil {
```
lippserd, line 100: That whole error handling looks overdone. The prepare returns `errUnknownSysVar` not the exec.

pkg/config/database.go
```diff
@@ -75,7 +81,46 @@ func (d *Database) Open(logger *logging.Logger) (*icingadb.DB, error) {
...
+		}
+
+		wsrepSyncWait := int64(d.Options.WsrepSyncWait)
+
+		// Setting "wsrep_sync_wait" for each session ensures that causality checks are performed before execution
+		// and that each statement is executed on a fully synchronized node. Doing so prevents foreign key violation
+		// when inserting into dependent tables on different MariaDB/MySQL nodes. When using MySQL single nodes, the
+		// "SET SESSION" command will fail with "Unknown system variable (1193)" and will therefore be silently dropped.
+		// https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait
+		var setGaleraOpts = func(ctx context.Context, conn driver.Conn) error {
```
lippserd, line 96: I don't understand why `setGaleraOpts` is set here and `errUnknownSysVar` on the package level. Either make both usable, or define both here.

lippserd (changes requested):

pkg/driver/driver.go
yhabteab: > hen we could move the constants and `RetryConnector` to the database package as well.

There's no such package in Icinga DB though! `Database#Open()` is located within the `config` package and moving the connector to that package feels out of place to me.

lippserd: True. I opt for renaming `driver` to `database` then. Our go libs will have that package and we still wouldn’t need that naming foo.

pkg/config/database.go
julianbrost: I believe that's a result of my suggestion to capture in the config option in https://github.com/Icinga/icingadb/pull/711#discussion_r1531695737.

But that could also be achieved like this:

```go
connector := &icingadbDriver.RetryConnector{
	Connector: c,
	SqlDriver: &mysql.MySQLDriver{},
	Logger: logger,
	InitConn: func(ctx context.Context, conn driver.Conn) error {
		return setWsrepSyncWait(ctx, conn, wsrepSyncWait)
	}
}
```

That would probably make the code a bit nicer to read compared to the large anonymous function currently in place.

pkg/driver/driver.go
julianbrost: `pkg/icingadb/` could also be an option as that's kind of the predecessor of `icinga-go-library/database`.

Both have the `DB` type (https://github.com/Icinga/icinga-go-library/blob/main/database/db.go, https://github.com/Icinga/icingadb/blob/main/pkg/icingadb/db.go) and moving this next to it doesn't feel out of place for me.

pkg/icingadb/driver.go
```diff
@@ -33,24 +46,28 @@ func (c RetryConnector) Connect(ctx context.Context) (driver.Conn, error) {
 		ctx,
 		func(ctx context.Context) (err error) {
 			conn, err = c.Connector.Connect(ctx)
+			if err == nil && c.initConn != nil {
+				err = c.initConn(ctx, conn)
```
Al2Klimov, line 50: If c.Connector.Connect() succeeds, but c.initConn() doesn't, shouldn't we close conn explicitly?

Al2Klimov (commented):
Why did you drop my unit tests during cherry pick?

pkg/config/database.go
```diff
@@ -173,3 +182,37 @@ func (d *Database) isUnixAddr() bool {
...
+	if err != nil {
+		errUnknownSysVar := &mysql.MySQLError{Number: 1193}
+		if errors.Is(err, errUnknownSysVar) {
+			return nil
+		}
+
+		return errors.Wrap(err, "cannot prepare "+galeraOpts)
+	}
+	// This is just for an unexpected exit and can safely be ignored any returned errors and in case
+	// of the normal function exit, the stmt is closed manually, and it's error is handled gracefully.
```
julianbrost, line 205: Wrong grammar: should either be "can safely ignore any returned errors" or "any returned errors can safely be ignored". Also: "its error".

pkg/icingadb/driver.go
yhabteab: We shouldn't actually have retries around the `initConn` part, I just put it there after https://github.com/Icinga/icingadb/pull/711#discussion_r1531629225 without thinking about it thoroughly. If there are retryable errors, `c.Connector.Connect()` should have retried them and given up at that point, so there is no need for additional retries. Anyway, the connection should definitely be closed!

pkg/icingadb/db.go
```diff
@@ -54,6 +53,10 @@ type Options struct {
 	// MaxRowsPerTransaction defines the maximum number of rows per transaction.
 	// The default is 2^13, which in our tests showed the best performance in terms of execution time and parallelism.
 	MaxRowsPerTransaction int `yaml:"max_rows_per_transaction" default:"8192"`
+
+	// WsrepSyncWait defines which kinds of SQL statements catch up all pending sync between nodes first, see:
+	// https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait
+	WsrepSyncWait int `yaml:"wsrep_sync_wait" default:"7"`
```
julianbrost, line 59: It would be helpful if the comment mentioned Galera cluster outside of just the URL.

Also, this should also be documented properly. After all, wasn't the idea to have this configurable so that we use the safer option by default but in a setup where it isn't needed, this can be set to 0 if it's not necessary as connections are all routed to the same node?

yhabteab:
> Why did you drop my unit tests during cherry pick?

What does exactly cover your unit tests? It's somewhat exaggerated just to test error cases that are ultimately triggered by the underlying driver, which does not lie within the responsibility of Icinga DB.

pkg/icingadb/driver.go
```diff
@@ -1,29 +1,42 @@
...
 
-const MySQL = "icingadb-mysql"
-const PostgreSQL = "icingadb-pgsql"
+// Driver names that are automatically registered by the respective packages under exactly these names.
+// Whenever the default names in their packages are altered or should we use other driver packages of these kind,
+// they should always be registered under these names. We do not intend to change these constant names ever again.
+const (
+	MySQL      string = "mysql"
+	PostgreSQL string = "postgres"
+)
```
julianbrost, line 22: Isn't the comment just saying that the values of the constants are the driver names as registered in the `database/sql` packages by the drivers themselves?

> Whenever the default names in their packages are altered or should we use other driver packages of these kind, they should always be registered under these names.

I'm not exactly sure what "under these names" is supposed to say exactly. But that sentence can be interpreted as "if we would device to use `pgx` instead of `postgres`, we'd still have to make sure it's registered as `postgres`" which sounds strange.

Also, it sounds unlikely that a driver implementation would change the name it registers with `database/sql`, so I don't think it needs to be mentioned explicitly (it should be clear from "as registered with `database/sql`" that if that changes, these constants have to be adapted as well).

> We do not intend to change these constant names ever again.

Why? Especially why does it need to be mentioned here? Sounds a bit like "I know things will break if you change it but I won't tell you why" at the moment.

julianbrost (changes requested):

pkg/icingadb/driver.go
julianbrost: > We shouldn't actually have retries around the initConn part

If a connection is established but for some reason, `initConn()` encounters a connection error (connection reset by peer for example), why shouldn't it be retried with a new connection?

> If there are retryable errors, `c.Connector.Connect()` should have retried them and given up at that point, so there is no need for additional retries.

Isn't `c.Connector.Connect()` simply the connect function of `go-sql-driver/mysql` or `lib/pq` which doesn't have any special retry handling?

pkg/config/database.go
```diff
@@ -173,3 +182,37 @@ func (d *Database) isUnixAddr() bool {
...
+
+	stmt, err := conn.(driver.ConnPrepareContext).PrepareContext(ctx, galeraOpts)
+	if err != nil {
+		errUnknownSysVar := &mysql.MySQLError{Number: 1193}
+		if errors.Is(err, errUnknownSysVar) {
+			return nil
+		}
+
+		return errors.Wrap(err, "cannot prepare "+galeraOpts)
+	}
```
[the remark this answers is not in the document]
yhabteab: > Is this nesting of ifs only done here so that `errUnknownSysVar` isn't initialized unnecessarily in the `err == nil` case?

No. I just didn't want to construct `if ... else if ..` instead. I don't actually see anything wrong with nesting these two ifs.

pkg/icingadb/driver.go
yhabteab: > If a connection is established but for some reason, initConn() encounters a connection error (connection reset by peer for example), why shouldn't it be retried with a new connection?

If `c.Connector.Connect()` succeeds and `initConn()` fails, we would have to close every successfully established connection after each retry and waste time just because of `initConn`.

> Isn't c.Connector.Connect() simply the connect function of go-sql-driver/mysql or lib/pq which doesn't have any special retry handling?

I mean our retry handler around the `Connect()` call by that.

julianbrost: > > If a connection is established but for some reason, initConn() encounters a connection error (connection reset by peer for example), why shouldn't it be retried with a new connection?
> 
> If `c.Connector.Connect()` succeeds and `initConn()` fails, we would have to close every successfully established connection after each retry and waste time just because of `initConn`.

I think I start to understand what you're thinking here. Is it that if `initConn` would receive some kind of "not ready yet, try again later" error, which would then result in the connection being thrown away even though it would probably get into a working state later?

So yes, there are situations were connection might be closed unnecessarily.

However, consider the scenario when a new connection is established about the same time when the database server is restarted. So `Connect()` could still have worked perfectly fine, however, `initConn()` might receive some error like "server is shutting down" or "connection closed". If you don't retry `initConn()` or retry it separately from `Connect()`, this would report an error while it would have worked perfectly fine if it reattempted with a fresh connection.

Now unless you try to distinguish these situations (we got to #698, so I don't think we want to try this here), the safe option is to close the connection and try again on a new one, even if not strictly necessary in some situations.

pkg/icingadb/driver.go
yhabteab: > But that sentence can be interpreted as "if we would device to use `pgx` instead of `postgres`, we'd still have to make sure it's registered as `postgres`" which sounds strange.

Why does this feel strange? After all, no matter which drivers we use, they all have to be either of type `postgres` or `mysql` and it makes no sense to refer to a driver like `pgx` just because its maintainers have named it so. Btw. this sentence just says that if we ever use a new driver that is not yet known by `sqlx`, we need to make sure we register it using one of these constants.

julianbrost (changes requested):

config.example.yml
```diff
@@ -23,6 +23,14 @@ database:
   # Database password.
   password: CHANGEME
 
+  # List of low-level database options that can be set to influence some Icinga DB internal default behaviours.
+  options:
+    # Enforce Galera cluster nodes strict cluster-wide causality checks before executing specific SQL queries
+    # determined by the number you provided. By default, 7 is used, which includes "SELECT,DELETE,INSERT,REPLACE"
+    # query types and is usually sufficient for running Icinga DB with Galera setups and a database load balancer.
+    # When you use Galera cluster but don't have a load balancer, you can set this to 0 to disable it completely.
```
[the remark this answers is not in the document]
yhabteab: > Also, this might be worth a section in `05-Distributed-Setups.md`.

I don't think so! This is just about how an external tool has to be configured and has nothing to do with Icinga DB's own distributed setup. In any case, I would prefer `03-Configuration.md` if you want to put it in an own section, as it's dealing with some configuration techniques.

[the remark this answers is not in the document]
oxzi: > When you use Galera cluster but don't have a load balancer, you can set this to 0 to disable it completely.

Isn't it quite the opposite? If connecting to random DB nodes, one might wanna have `wsrep_sync_wait: 7`. However, when either a load balancer - routing requests to the same DB node - or a virtual IP on one DB node is configured, no special `wsrep_sync_wait` is needed.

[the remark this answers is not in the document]
julianbrost: > > Also, this might be worth a section in `05-Distributed-Setups.md`.
> 
> I don't think so! This is just about how an external tool has to be configured and has nothing to do with Icinga DB's own distributed setup. In any case, I would prefer `03-Configuration.md` if you want to put it in an own section, as it's dealing with some configuration techniques.

Currently there are three places that have some explanation of what `wsrep_sync_wait` does, were some even discourage going into it in more detail (mainly the config option table, that won't look nice in neither the source nor the rendered version with long text). So my main idea is to have a primary place for the explanation that's not very space constrained to be able to go into the necessary detail. Where exactly this should be can be discussed. For the documentation I expect as part of this PR, `03-Configuration.md` should be fine. If we were to add more specific hints on how to best set up an Galera cluster for Icinga DB (like load balancer yes/no?, virtual/failover IP address yes/no?, but I'm not sure if we actually want to make such recommendations), that might not really fit into the Icinga DB configuration reference.

> When you use Galera cluster but don't have a load balancer, you can set this to 0 to disable it completely.

FYI: my comment was supposed for this sentence in particular, but if you select only a single line, GitHub shows some extra context making this less clear.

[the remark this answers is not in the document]
oxzi: @yhabteab and I just talked about my prior comment afk. In a nutshell, when referring to a load balancer, everyone thinks of something else. A load balancer might, well, balance the load by distributing the queries to the nodes or route the queries to a distinct node, until this one fails. @yhabteab thought about the first case, I about the latter.

The issue is, as @julianbrost pointed out, more how the network is being built. When, for example, a round robin DNS record points to all Galera nodes, the `wsrep_sync_wait` value is necessary. Same goes for a randomly distributing load balancer or an anycast IP address. However, when using a load balancer routing to a distinct node or assigning a virtual (unicast) IP address, no further options are necessary.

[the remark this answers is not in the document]
yhabteab: > Regarding `05-Distributed-Setups.md`: I like the idea of showing users one supported way how to realize a HA database.

Seriously 😆? There's no way I would tell some system administrators who are experts in their field how to better set up an HA database, nor is it our job to do so.

> However, the SQL database - even when being a crucial part of the setup - is just being assumed to be HA

You can also use Icinga without having to set up a database cluster.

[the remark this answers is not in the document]
oxzi: > There's no way we would tell some system administrators out there who are experts in their field

But then there are still the other sysadmins :upside_down_face:

doc/03-Configuration.md
```diff
@@ -91,3 +98,19 @@ allowing to keep this information for longer with a smaller storage footprint.
...
 A duration string is a sequence of decimal numbers and a unit suffix, such as `"20s"`.
 Valid units are `"ms"`, `"s"`, `"m"` and `"h"`.
+
+### Galera Cluster
+
+Icinga DB might not function properly or even crash occasionally when it is used with [Galera cluster](https://mariadb.com/kb/en/what-is-mariadb-galera-cluster/) setups.
+To avoid such unexpected crashes and reading/writing inconsistent data, Icinga DB sets the [wsrep_sync_wait](https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait)
+system variable for all its database connections. Consequently, strict cluster-wide causality checks are enforced before
+executing specific SQL queries, which are determined by the value set in the `"wsrep_sync_wait"` system variable. By default,
+Icinga DB sets this to `7`, which includes `READ, DELETE, INSERT, REPLACE` query types and is usually sufficient.
```
oxzi, line 108:
```suggestion
Icinga DB sets this to `7`, which includes `READ, UPDATE, DELETE, INSERT, REPLACE` query types and is usually sufficient.
```

doc/03-Configuration.md
```diff
@@ -91,3 +98,19 @@ allowing to keep this information for longer with a smaller storage footprint.
...
+Unfortunately, this also has the downside that every single Icinga DB query will be blocked until the cluster nodes
+resynchronise their states after each executed query, and may result in degraded performance.
+
+However, this does not necessarily have to be the case if, for instance, Icinga DB is only allowed to connect to a single
+cluster node at a time. This is the case when a load balancer does not randomly route connections to all the nodes evenly,
+but always to the same node until it fails, or if your database cluster does not implement an IP address fail over. In such
```
oxzi, line 114: > does not implement an IP address fail over

If a virtual IP address is being used, which is only changed for failover situations - e.g., by keepalived - no `wsrep_sync_wait` should be necessary. I would change this sentence to:

> […], or if your database cluster nodes have a virtual failover IP address assigned.

config.example.yml
```diff
@@ -23,6 +23,14 @@ database:
   # Database password.
   password: CHANGEME
 
+  # List of low-level database options that can be set to influence some Icinga DB internal default behaviours.
+  options:
+    # Enforce Galera cluster nodes strict cluster-wide causality checks before executing
```
oxzi, line 28:
```suggestion
    # Enforce Galera cluster nodes to perform strict cluster-wide causality checks before executing
```

Or some other predicate.

doc/03-Configuration.md
```diff
@@ -29,19 +29,26 @@ This is also the database used in
...
+| ca       | **Optional.** Path to TLS CA certificate.                                                                                                                                        |
+| insecure | **Optional.** Whether not to verify the peer.                                                                                                                                    |
+| options  | **Optional.** List of low-level database options that can be set to influence some Icinga DB internal default behaviours. See [database options](#database-options) for details. |
+
+### Database Options
+
+| Option          | Description                                                                                                          |
+|-----------------|----------------------------------------------------------------------------------------------------------------------|
+| wsrep_sync_wait | **Optional.** Enforce [Galera cluster](#galera-cluster) nodes strict cluster-wide causality checks. Defaults to `7`. |
```
oxzi, line 51: Same as above.

oxzi (commented):
Btw, are we using British or American English? I've seen both so far, but your text seems to be British, e.g., in "behaviours" or "resynchronise".

yhabteab:
> Btw, are we using British or American English? I've seen both so far, but your text seems to be British, e.g., in "behaviours" or "resynchronise".

I didn't think about it that much, but I've also changed "synchronization" -> "synchronisation" now 😆.

doc/03-Configuration.md
```diff
@@ -93,3 +105,19 @@ allowing to keep this information for longer with a smaller storage footprint.
 
 A duration string is a sequence of decimal numbers and a unit suffix, such as `"20s"`.
 Valid units are `"ms"`, `"s"`, `"m"` and `"h"`.
+
+### Galera Cluster
+
+Icinga DB might not function properly or even crash occasionally when it is used with [Galera cluster](https://mariadb.com/kb/en/what-is-mariadb-galera-cluster/) setups.
+To avoid such unexpected crashes and reading/writing inconsistent data, Icinga DB sets the [wsrep_sync_wait](https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait)
+system variable for all its database connections. Consequently, strict cluster-wide causality checks are enforced before
```
julianbrost, line 113: That's a strange way of starting this section given that this PR is supposed to fix exactly that. The documentation should primarily describe the behavior of the new version and this still reads like it doesn't work properly.

I'd start something like "Icinga DB expects a more consistent behavior from its database than a Galera cluster provides by default. To accommodate for this, Icinga DB sets Icinga DB sets the `wsrep_sync_wait` system variable ..."

julianbrost (changes requested):

julianbrost:
Otherwise, the implementation should be fine. I tested a few things:

- Reconnect still works fine, both when the database is started after Icinga DB and when the database goes down for a bit while Icinga DB is still running.
- Additionally ran the integration tests locally with MariaDB 10.1, MariaDB latest (11.3), and MySQL 5.5 (`mysql:latest` is tested in GitHub Actions) to check that there are no compatibility issues and also the simplified error handling (https://github.com/Icinga/icingadb/pull/711#discussion_r1536898564) works consistently.

@oxzi So if you you're happy with the code as well, now would be a good time to give it another test against your Galera cluster.

oxzi:
As I just talked with @yhabteab about the handling of an invalid `wsrep_sync_wait` in the Icinga DB config, here's the output:

```
icingadb-master1  | 2024-03-27T14:37:30.565Z	INFO	icingadb	Starting Icinga DB daemon (1.1.1)
icingadb-master1  | 2024-03-27T14:37:30.565Z	INFO	icingadb	Connecting to database at 'mysql:3306'
icingadb-master1  | 2024-03-27T14:37:30.568Z	WARN	database	Can't connect to database. Retrying	{"error": "cannot execute SET SESSION wsrep_sync_wait=?: Error 1231 (42000): Variable 'wsrep_sync_wait' can't be set to the value of '-1'", "errorVerbose": "Error 1231 (42000): Variable 'wsrep_sync_wait' can't be set to the value of '-1'\ncannot execute SET SESSION wsrep_sync_wait=?\ngithub.com/icinga/icingadb/pkg/config.setGaleraOpts\n\t/go/src/github.com/Icinga/icingadb/pkg/config/database.go:209\ngithub.com/icinga/icingadb/pkg/config.(*Database).Open.func1\n\t/go/src/github.com/Icinga/icingadb/pkg/config/database.go:83\ngithub.com/icinga/icingadb/pkg/icingadb.RetryConnector.Connect.func1\n\t/go/src/github.com/Icinga/icingadb/pkg/icingadb/driver.go:47\ngithub.com/icinga/icingadb/pkg/retry.WithBackoff\n\t/go/src/github.com/Icinga/icingadb/pkg/retry/retry.go:49\ngithub.com/icinga/icingadb/pkg/icingadb.RetryConnector.Connect\n\t/go/src/github.com/Icinga/icingadb/pkg/icingadb/driver.go:42\ndatabase/sql.(*DB).conn\n\t/usr/local/go/src/database/sql/sql.go:1415\ndatabase/sql.(*DB).PingContext.func1\n\t/usr/local/go/src/database/sql/sql.go:883\ndatabase/sql.(*DB).retry\n\t/usr/local/go/src/database/sql/sql.go:1566\ndatabase/sql.(*DB).PingContext\n\t/usr/local/go/src/database/sql/sql.go:882\ndatabase/sql.(*DB).Ping\n\t/usr/local/go/src/database/sql/sql.go:900\nmain.run\n\t/go/src/github.com/Icinga/icingadb/cmd/icingadb/main.go:68\nmain.main\n\t/go/src/github.com/Icinga/icingadb/cmd/icingadb/main.go:37\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:271\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1695"}
icingadb-master1  | 2024-03-27T14:37:30.568Z	FATAL	icingadb	Error 1231 (42000): Variable 'wsrep_sync_wait' can't be set to the value of '-1'
icingadb-master1  | cannot execute SET SESSION wsrep_sync_wait=?
icingadb-master1  | github.com/icinga/icingadb/pkg/config.setGaleraOpts
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/pkg/config/database.go:209
icingadb-master1  | github.com/icinga/icingadb/pkg/config.(*Database).Open.func1
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/pkg/config/database.go:83
icingadb-master1  | github.com/icinga/icingadb/pkg/icingadb.RetryConnector.Connect.func1
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/pkg/icingadb/driver.go:47
icingadb-master1  | github.com/icinga/icingadb/pkg/retry.WithBackoff
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/pkg/retry/retry.go:49
icingadb-master1  | github.com/icinga/icingadb/pkg/icingadb.RetryConnector.Connect
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/pkg/icingadb/driver.go:42
icingadb-master1  | database/sql.(*DB).conn
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:1415
icingadb-master1  | database/sql.(*DB).PingContext.func1
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:883
icingadb-master1  | database/sql.(*DB).retry
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:1566
icingadb-master1  | database/sql.(*DB).PingContext
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:882
icingadb-master1  | database/sql.(*DB).Ping
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:900
icingadb-master1  | main.run
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/cmd/icingadb/main.go:68
icingadb-master1  | main.main
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/cmd/icingadb/main.go:37
icingadb-master1  | runtime.main
icingadb-master1  | 	/usr/local/go/src/runtime/proc.go:271
icingadb-master1  | runtime.goexit
icingadb-master1  | 	/usr/local/go/src/runtime/asm_amd64.s:1695
icingadb-master1  | can't retry
icingadb-master1  | github.com/icinga/icingadb/pkg/retry.WithBackoff
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/pkg/retry/retry.go:68
icingadb-master1  | github.com/icinga/icingadb/pkg/icingadb.RetryConnector.Connect
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/pkg/icingadb/driver.go:42
icingadb-master1  | database/sql.(*DB).conn
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:1415
icingadb-master1  | database/sql.(*DB).PingContext.func1
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:883
icingadb-master1  | database/sql.(*DB).retry
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:1566
icingadb-master1  | database/sql.(*DB).PingContext
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:882
icingadb-master1  | database/sql.(*DB).Ping
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:900
icingadb-master1  | main.run
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/cmd/icingadb/main.go:68
icingadb-master1  | main.main
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/cmd/icingadb/main.go:37
icingadb-master1  | runtime.main
icingadb-master1  | 	/usr/local/go/src/runtime/proc.go:271
icingadb-master1  | runtime.goexit
icingadb-master1  | 	/usr/local/go/src/runtime/asm_amd64.s:1695
icingadb-master1  | can't connect to database
icingadb-master1  | github.com/icinga/icingadb/pkg/icingadb.RetryConnector.Connect
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/pkg/icingadb/driver.go:42
icingadb-master1  | database/sql.(*DB).conn
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:1415
icingadb-master1  | database/sql.(*DB).PingContext.func1
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:883
icingadb-master1  | database/sql.(*DB).retry
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:1566
icingadb-master1  | database/sql.(*DB).PingContext
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:882
icingadb-master1  | database/sql.(*DB).Ping
icingadb-master1  | 	/usr/local/go/src/database/sql/sql.go:900
icingadb-master1  | main.run
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/cmd/icingadb/main.go:68
icingadb-master1  | main.main
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/cmd/icingadb/main.go:37
icingadb-master1  | runtime.main
icingadb-master1  | 	/usr/local/go/src/runtime/proc.go:271
icingadb-master1  | runtime.goexit
icingadb-master1  | 	/usr/local/go/src/runtime/asm_amd64.s:1695
icingadb-master1  | can't connect to database
icingadb-master1  | main.run
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/cmd/icingadb/main.go:70
icingadb-master1  | main.main
icingadb-master1  | 	/go/src/github.com/Icinga/icingadb/cmd/icingadb/main.go:37
icingadb-master1  | runtime.main
icingadb-master1  | 	/usr/local/go/src/runtime/proc.go:271
icingadb-master1  | runtime.goexit
icingadb-master1  | 	/usr/local/go/src/runtime/asm_amd64.s:1695
icingadb-master1  | exit status 1
```

Thus, the SQL warning results in a Go error. However, it might be good to check if the value is within `[0, 15]` and report an even more human-friendly error message.

doc/03-Configuration.md
```diff
@@ -93,3 +105,18 @@ allowing to keep this information for longer with a smaller storage footprint.
 
 A duration string is a sequence of decimal numbers and a unit suffix, such as `"20s"`.
 Valid units are `"ms"`, `"s"`, `"m"` and `"h"`.
+
+### Galera Cluster
+
+Icinga DB expects a more consistent behaviour from its database than a [Galera cluster](https://mariadb.com/kb/en/what-is-mariadb-galera-cluster/) provides by default.
+To accommodate this, Icinga DB sets the [wsrep_sync_wait](https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait) system variable for all its database connections. Consequently,
```
julianbrost, line 112: Now the lines are getting really long here. You might want to reflow the sections you added to 120 characters length (that's what most[^1] lines in this file outside of tables adhere to).

[^1]: There's one line with 121 characters. :(

julianbrost (changes requested):

julianbrost (dismissed):

yhabteab:
I just have rebased this and resolved some conflicts!

julianbrost (approved):

Al2Klimov:
> > Why did you drop my unit tests during cherry pick?
> 
> What does exactly cover your unit tests? It's somewhat exaggerated just to test error cases that are ultimately triggered by the underlying driver, which does not lie within the responsibility of Icinga DB.

It _is_ our responsibility to swallow the correct error.

julianbrost:
> > > Why did you drop my unit tests during cherry pick?
> > 
> > 
> > What does exactly cover your unit tests? It's somewhat exaggerated just to test error cases that are ultimately triggered by the underlying driver, which does not lie within the responsibility of Icinga DB.
> 
> It _is_ our responsibility to swallow the correct error.

You're probably referring to this test file from #665: https://github.com/Icinga/icingadb/blob/error-1452-foreign-key-577/pkg/driver/mysql_test.go

To be honest, I haven't looked at it in detail and it's not immediately obvious to me what this actually tests. So please explain them.

The current implementation isn't really ideal for testability. Maybe we could improve that by replacing `setGaleraOpts(ctx, conn, wsrepSyncWait)` with a more generic `setSessionVariableIfExists(ctx, conn, variable, value)` function that could then be tested against an actual database server in [`tests/sql/`](https://github.com/Icinga/icingadb/tree/main/tests/sql) with other variables. Like setting a random variable name to test that the error is ignored, setting a variable present on all MySQL/MariaDB versions to an invalid value to test the other error handling, etc.

julianbrost:
> that could then be tested against an actual database server in [`tests/sql/`](https://github.com/Icinga/icingadb/tree/main/tests/sql?rgh-link-date=2024-04-04T08%3A43%3A55Z)

When writing this, I missed that this would add dependencies between `icingadb/go.mod` and `icingadb/tests/go.mod` that we want to avoid as the integration tests are supposed to test the functionality of a compiled Icinga DB binary and therefore not import its code.

As we plan to move that database code to `icinga-go-library` anyways, I'd rather take the opportunity there to add proper tests for more of the database code. @yhabteab already started a PR and I wrote a comment where I explained a bit how I imagine this: https://github.com/Icinga/icinga-go-library/pull/10#pullrequestreview-1983085171

The tests proposed in#665 seem to mainly test that one error handling condition is there with quite a bit of complexity in the test cases that seems unnecessary for what it's supposed to test. Restructuring the tests within this project when the corresponding implementation is supposed to be moved to another project soon anyways, feels like unnecessary work, so I'd actually postpone the proper automatic testing of this function until it has reached its new home. Apart from that, the existing integration tests already cover that this PR isn't fundamentally breaking anything and automatically testing that Icinga DB works with a Galera cluster as part of this PR is out of scope in any way.

PR #715, opened 2024-03-22
Upgrade `go-redis` to `v9`

### LINKED ISSUE #708: Update to `github.com/redis/go-redis/v9`
I just noticed that we're using `github.com/go-redis/redis/v8` while [`v9`](https://github.com/redis/go-redis/releases/tag/v9.0.0) exists for over a year and `v8` doesn't seem to get any further updates. Looks like this change isn't picked up by dependabot.

lippserd:
Closes  #708.


oxzi:
I pushed a commit referencing this <https://github.com/Icinga/icinga-testing/pull/30> `icinga-testing` version.

After that PR got merged, this one needs to update its `icinga-testing` dependency again, afterwards it should be fine to be merged.

julianbrost:
The [changelog for v9](https://github.com/redis/go-redis/blob/master/CHANGELOG.md#v9-2023-01-30) only mentions that pipelines are no longer thread-safe. The two places where we use pipelines ([overdue sync](https://github.com/Icinga/icingadb/blob/e31b101f4f41941793adc7b430a0db8bef71b8c3/pkg/icingadb/overdue/sync.go#L95-L113), [runtime updates](https://github.com/Icinga/icingadb/blob/e31b101f4f41941793adc7b430a0db8bef71b8c3/pkg/icingadb/runtime_updates.go#L248-L281)) don't do any parallelization where they use these pipelines, so that should be just fine.

julianbrost (approved):
Tests pass, I gave it a quick run locally, seems to work just fine.

PR #716, opened 2024-03-22
GitHub Actions: Notify Go Module Major Version Updates

oxzi:
When Go modules mature and their major version has exceeded one, the [module path requires a version suffix](https://go.dev/ref/mod#major-version-suffixes), such as `/v2`. Unfortunately, the Dependabot is unable to detect such a Go module major version bump for suffixed major versions, resulting in stalling dependencies - <https://github.com/dependabot/dependabot-core/issues/2213>.

By introducing another GitHub Actions Workflow utilizing [gomajor](https://github.com/icholy/gomajor), those major version updates for Go modules can be identified. As a scheduled workflow, running each Monday morning, it will either create or update an open GitHub issue with all currently available major version updates for referenced Go modules.

As a playground, I have forked this repository to another GitHub repository. Please feel free to take a look at its [issue tracker](https://github.com/oxzi/icingadb-go-mod-major-test/issues?q=is%3Aissue). Issue <https://github.com/oxzi/icingadb-go-mod-major-test/issues/1>, for example, shows how a new issue was first created and later re-used while it was still open. After it was closed, a new issue <https://github.com/oxzi/icingadb-go-mod-major-test/issues/2> was created and used afterwards, until this one was also closed.

An open issue will look like the following:

> ## Go Module Major Version Updates
> There are major version updates available for used Go modules.
> 
> ### Updates in `.`
> 1. `github.com/go-redis/redis/v8: v8.11.5 [latest v9.5.1]`
> 2. `github.com/vbauerster/mpb/v6: v6.0.4 [latest v8.7.2]`
> 
> ### Updates in `tests`
> 1. `github.com/go-redis/redis/v8: v8.11.5 [latest v9.5.1]`

If this idea suits you, I can also submit this PR against our other Go projects.


.github/workflows/go-mod-major.yml
```diff
@@ -0,0 +1,25 @@
+name: Go Module Major Update Check
+on:
+  schedule:
+    - cron: '30 6 * * 1' # at 06:30 every Monday morning
```
Al2Klimov, line 4: I've used this trick [myself](https://github.com/Al2Klimov/check_rungrep/blob/master/.github/workflows/rust-update.yml) and according to GH's eMails they're auto-suspending these jobs after the last code change and an mount of time I don't remember.

PR #718, opened 2024-03-25
Document `retention.count` & `retention.interval` options

### LINKED ISSUE #717: Document `retention.count` & `retention.interval` configs
## Describe the bug

Currently the default retention delete counter happens to be `5000` and occasionally causes some strange SQL errors (**Error 1206: The total number of locks exceeds the size of the lock table**). We should reduce these numbers to a much smaller number like `500` or maybe even smaller. [`MySQL`](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_lock_table_full) also suggests splitting such large operations into smaller chunks instead of tweaking system variables like the `innodb_buffer_pool_size` to get around this problem.

PS: We should also document the `retention.count` option [here](https://icinga.com/docs/icinga-db/latest/doc/03-Configuration/#retention).

refs https://github.com/Icinga/icingadb/issues/618

yhabteab:
fixes #717

### DIFF
--- a/config.example.yml
+++ b/config.example.yml
@@ -79,6 +79,16 @@ retention:
   # Number of days to retain historical data for SLA reporting. By default, it is retained forever.
 #  sla-days:
 
+  # Interval for periodically cleaning up the historical data, defined as a duration string.
+  # A duration string is a sequence of decimal numbers and a unit suffix, such as "20s".
+  # Valid units are "ms", "s", "m", "h".
+  # Defaults to "1h".
+#  interval: 1h
+
+  # Number of old historical entries a single query can delete in a "DELETE FROM ... LIMIT count" manner.
+  # By default, this is limited to 5000 entries.
+#  count: 5000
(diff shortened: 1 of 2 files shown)


config.example.yml
```diff
@@ -79,6 +79,15 @@ retention:
   # Number of days to retain historical data for SLA reporting. By default, it is retained forever.
 #  sla-days:
 
+  # Interval for periodically cleaning up the historical data, defined as a duration string.
+  # A duration string is a sequence of decimal numbers and a unit suffix, such as "20s".
+  # Valid units are "ms", "s", "m", "h". Defaults to "1h".
```
oxzi, line 84: Totally pedantic, but could you please add a newline between the list of valid units and the default to make the default value stand out more.

oxzi (approved):

Al2Klimov (approved):

PR #719, opened 2024-03-26
retry: distinguish non-retryable errors from timeouts

oxzi:
The WithBackoff function returns a wrapped error with the message "can't retry" both for errors marked as non-retryable by IsRetryable as well as when the context is done. The latter might happen when a positive settings.Timeout is specified, as this results in a wrapped context with a timeout.

This behavior resulted in me debugging the IsRetryable function, when just the Timeout has exceeded. To save others from having this experience, I have adapted the messages.

### DIFF
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -3,6 +3,7 @@ package retry
 import (
 	"context"
 	"database/sql/driver"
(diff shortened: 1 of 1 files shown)


oxzi:
#698 addresses the same issue.

PR #723, opened 2024-03-27
Document `database.options` properly

yhabteab:
These options are user configurable but are not yet documented anywhere visible to the end user. This PR should hopefully spare us future questions like *how many connections Icinga DB is allowed to open?* or *Icinga DB opens abnormally too many connections to the database?*. Yes, these questions are serious and were also asked by our support customers.

### DIFF
--- a/config.example.yml
+++ b/config.example.yml
@@ -23,6 +23,33 @@ database:
   # Database password.
   password: CHANGEME
 
+  # List of low-level database options that can be set to influence some Icinga DB internal default behaviours.
+  # Do not change the defaults if you don't have to!
+  options:
+    # Maximum number of connections Icinga DB is allowed to open in parallel to the database.
+    # By default, Icinga DB is allowed to open up to "16" connections whenever necessary.
+    # Setting this to a number less than or equal to "-1" allows Icinga DB to open an unlimited number of connections.
+    # However, it is not possible to set this option to "0".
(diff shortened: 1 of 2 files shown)


config.example.yml
```diff
@@ -23,6 +23,28 @@ database:
   # Database password.
   password: CHANGEME
 
+  # List of low-level database options that can be set to influence some Icinga DB internal default behaviours.
+  # Do not change the defaults if you don't have to!
+  options:
+    # Maximum number of connections Icinga DB is allowed to open in parallel to the database.
+    # By default, Icinga DB is allowed to open up to "16" connections whenever necessary.
```
oxzi, line 30: Also note that `-1` can be used for no connection limit, but 0 is not allowed.

config.example.yml
```diff
@@ -23,6 +23,28 @@ database:
...
+#    max_connections: 16
+
+    # Maximum number of queries allowed to connect to a single database table simultaneously.
+    # By default, Icinga DB is allowed to execute up to "8" queries of any kind, e.g. INSERT,UPDATE,DELETE
+    # concurrently on a given table.
```
oxzi, line 35: Please note for this and the following options that a positive integer is expected.

oxzi (approved):

Al2Klimov (approved):

PR #732, opened 2024-04-08
Register mysql default logger prior to `mysql#NewConfig()`

yhabteab:
**Before:**
```bash
2024-04-08T13:55:58.621+0200    WARN    database        Can't connect to database. Retrying     {"error": "driver: bad connection"}
2024-04-08T13:56:00.294+0200    WARN    database        Can't connect to database. Retrying     {"error": "dial tcp [::1]:3306: connect: connection refused"}
2024-04-08T13:56:40.352+0200    INFO    database        Reconnected to database {"after": "41.737475334s", "attempts": 10}
[mysql] 2024/04/08 14:03:19 connection.go:49: closing bad idle connection: EOF
[mysql] 2024/04/08 14:03:19 connection.go:49: closing bad idle connection: EOF
```

**After:**
```bash
2024-04-08T14:41:53.032+0200    DEBUG   database        [unexpected EOF]
2024-04-08T14:41:53.272+0200    DEBUG   database        [unexpected EOF]
2024-04-08T14:41:53.272+0200    WARN    database        Can't connect to database. Retrying     {"error": "driver: bad connection"}
2024-04-08T14:41:53.353+0200    DEBUG   database        [unexpected EOF]
2024-04-08T14:41:53.581+0200    DEBUG   database        [unexpected EOF]
2024-04-08T14:41:53.878+0200    DEBUG   database        [unexpected EOF]
...
2024-04-08T14:46:12.751+0200    DEBUG   database        [closing bad idle connection:  EOF]
2024-04-08T14:46:12.751+0200    DEBUG   database        [closing bad idle connection:  EOF]
2024-04-08T14:46:12.751+0200    DEBUG   database        [closing bad idle connection:  EOF]
2024-04-08T14:46:12.751+0200    DEBUG   database        [closing bad idle connection:  EOF]
2024-04-08T14:46:12.753+0200    DEBUG   database        [closing bad idle connection:  EOF]
```

### DIFF
--- a/pkg/config/database.go
+++ b/pkg/config/database.go
@@ -42,6 +42,7 @@ func (d *Database) Open(logger *logging.Logger) (*icingadb.DB, error) {
 
 		config.User = d.User
 		config.Passwd = d.Password
+		config.Logger = icingadb.MysqlFuncLogger(logger.Debug)
 
 		if d.isUnixAddr() {
 			config.Net = "unix"
@@ -70,8 +71,6 @@ func (d *Database) Open(logger *logging.Logger) (*icingadb.DB, error) {
 				return nil, errors.Wrap(err, "can't register TLS config")
 			}
 		}
-
-		_ = mysql.SetLogger(icingadb.MysqlFuncLogger(logger.Debug))
 
 		c, err := mysql.NewConnector(config)
 		if err != nil {


oxzi (approved):

julianbrost:
Do we even care about the default logger? Currently, all it does is to set the public `Logger` attribute on the `Config` struct:

https://github.com/go-sql-driver/mysql/blob/4395c45fd098a81c5251667cda111f94c693ab14/errors.go#L53-L61
https://github.com/go-sql-driver/mysql/blob/4395c45fd098a81c5251667cda111f94c693ab14/dsn.go#L89
https://github.com/go-sql-driver/mysql/blob/4395c45fd098a81c5251667cda111f94c693ab14/dsn.go#L202-L204

Unfortunately, [the documentation for `SetLogger()`](https://pkg.go.dev/github.com/go-sql-driver/mysql#SetLogger) doesn't explicitly state that there will never be logging outside of a connection:

> SetLogger is used to set the default logger for critical errors. The initial logger is os.Stderr.

However, I think it's not totally unreasonable to assume that it stays this way (default logger kind of suggests that it just serves as the default for `Config` and even if it would change at some point in the future, it would be a bit annoying, but that's it).

I have the feeling that the default logger was only set as before #711, when the database was opened from a DSN, there was no other way to set a logger. So maybe we just want to change it and simply set `config.Logger` instead?

pkg/config/database.go
```diff
@@ -42,6 +42,7 @@ func (d *Database) Open(logger *logging.Logger) (*icingadb.DB, error) {
 
 		config.User = d.User
 		config.Passwd = d.Password
+		config.Logger = icingadb.MysqlFuncLogger(logger.Debug)
```
oxzi, line 45: I like the [documentation](https://pkg.go.dev/github.com/go-sql-driver/mysql@v1.8.1#Config) :smiling_face_with_tear:
```
Logger               Logger            // Logger
```

oxzi (approved):

julianbrost (approved):

PR #735, opened 2024-04-10
Increment go.mod Go version to 1.22

oxzi:
With the approaching release, the go version within go.mod can be incremented to a current Go version.

There were some small changes regarding the go.mod's go semantic since Go version 1.21, mainly enforcing compilation with a compatible version[^0]. As other modules' go.mod files are now also being included, there are additional entries in go.sum.

[^0]: https://go.dev/doc/modules/gomod-ref#go

@lippserd: As we talked about bumping the number to 1.21 or 1.22, I thought going up directly to the latest release would be more future proof. However, if you would prefer 1.21, I would alter this PR.


lippserd:
> @lippserd: As we talked about bumping the number to 1.21 or 1.22, I thought going up directly to the latest release would be more future proof. However, if you would prefer 1.21, I would alter this PR.

`1.22` is perfectly fine.

PR #743, opened 2024-04-18
doc/03-Configuration.md: Logging Fields in Journald

oxzi:
While working on an issue where the additional logging information provided by zap fields was missing, I realized that our documentation did not indicate how to view them. So a new documentation section has been added explaining how to view the fields when using the `systemd-journald` logging output.

Also removes one trailing space.

### DIFF
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -68,11 +68,11 @@ manual adjustments.
 Configuration of the logging component used by Icinga DB.
 
 | Option   | Description                                                                                                                                                                                              |
-|----------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| level    | **Optional.** Specifies the default logging level. Can be set to `fatal`, `error`, `warn`, `info` or `debug`. Defaults to `info`.                                                                        |
-| output   | **Optional.** Configures the logging output. Can be set to `console` (stderr) or `systemd-journald`. If not set, logs to systemd-journald when running under systemd, otherwise stderr.                  |
-| interval | **Optional.** Interval for periodic logging defined as [duration string](#duration-string). Defaults to `"20s"`.                                                                                         |
-| options  | **Optional.** Map of component name to logging level in order to set a different logging level for each component instead of the default one. See [logging components](#logging-components) for details. |
+|----------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| level    | **Optional.** Specifies the default logging level. Can be set to `fatal`, `error`, `warn`, `info` or `debug`. Defaults to `info`.                                                                                                                     |
+| output   | **Optional.** Configures the logging output. Can be set to `console` (stderr) or `systemd-journald`. Defaults to systemd-journald when running under systemd, otherwise to console. See notes below for [systemd-journald](#systemd-journald-fields). |
+| interval | **Optional.** Interval for periodic logging defined as [duration string](#duration-string). Defaults to `"20s"`.                                                                                                                                      |
+| options  | **Optional.** Map of component name to logging level in order to set a different logging level for each component instead of the default one. See [logging components](#logging-components) for details.                                              |
 
 ### Logging Components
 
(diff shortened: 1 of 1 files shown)


doc/03-Configuration.md
```diff
@@ -128,6 +128,16 @@ the cluster nodes resynchronise their states after each executed query, and may
...
 nodes evenly, but always to the same node until it fails, or if your database cluster nodes have a virtual IP address
-fail over assigned. In such situations, you can set the `wsrep_sync_wait` system variable to `0` in the 
+fail over assigned. In such situations, you can set the `wsrep_sync_wait` system variable to `0` in the
 `/etc/icingadb/config.yml` file to disable it entirely, as Icinga DB doesn't have to wait for cluster
 synchronisation then.
+
+### Systemd Journald Fields
+
+Using the `systemd-journald` logging output results in sending structured log entries to journald,
+including fields with additional information that would be included in the `console` logging output mode.
```
julianbrost, line 138: I'm not sure about the level of detail here. I don't know if the average user would understand what "including fields with additional information that would be included in the `console` logging output mode" is referring to. I think I'd omit this paragraph and extend the following one a bit mentioning that the messages logged to journald can contain extra `ICINGADB_*` fields that are not shown by default.

oxzi: I tried to rephrase this section. Please take another look.

julianbrost (approved):

doc/03-Configuration.md
```diff
@@ -128,6 +128,14 @@ the cluster nodes resynchronise their states after each executed query, and may
...
+fail over assigned. In such situations, you can set the `wsrep_sync_wait` system variable to `0` in the
 `/etc/icingadb/config.yml` file to disable it entirely, as Icinga DB doesn't have to wait for cluster
 synchronisation then.
+
+### Systemd Journald Fields
+
+When examining the journal with `journalctl`, fields containing additional information are hidden by default.
+Setting an appropriate
+[`--output` option](https://www.freedesktop.org/software/systemd/man/latest/journalctl.html#Output%20Options)
+will include them, such as: `journalctl --output verbose` or `journalctl --output json`.
```
yhabteab, line 140: I would also add `--unit 'icingadb.service'` to those example commands given that this is all about Icinga DB.

oxzi: I get your point. However, including the `--unit` argument for both examples makes this sentence very long, hiding the highlighted `--output` argument.

> Setting an appropriate [--output option](https://www.freedesktop.org/software/systemd/man/latest/journalctl.html#Output%20Options) will include them, such as: `journalctl --unit icingadb.service --output verbose` or `journalctl --unit icingadb.service --output json`.

Thus, I would prefer to drop `journalctl` and just show the two `--output` arguments.


> Setting an appropriate [--output option](https://www.freedesktop.org/software/systemd/man/latest/journalctl.html#Output%20Options) will include them, such as: `--output verbose` or `--output json`.

yhabteab: > However, including the `--unit` argument for both examples makes this sentence very long, hiding the highlighted `--output` argument.

But I don't quite get your point. Isn't this docs snippet supposed to help someone find the Icinga DB fields? I mean, OK, you want to highlight the `--output` argument even more, but in the end the user just needs a copy-pasteable one without having to construct the full command on his own. Personally, this just doesn't seem like a good approach to only partially document something.

oxzi: My assumption is that users have a basic understanding of journalctl, as otherwise they simply cannot operate it. Using it without any further parameter would result in logs since last log vacuum, which might have never happened. Thus, they might want to use `-f` or `-n 100` or `--since today` or something else..

I don't see any one-command-fits-it-all. My intention was just to point out that the necessary details are hidden, which is a non-default behavior, compared to most other software out there.

I have added a line showing one generic copy-pastable journalctl command.

doc/03-Configuration.md
```diff
@@ -128,6 +128,18 @@ the cluster nodes resynchronise their states after each executed query, and may
...
+
+When examining the journal with `journalctl`, fields containing additional information are hidden by default.
+Setting an appropriate
+[`--output` option](https://www.freedesktop.org/software/systemd/man/latest/journalctl.html#Output%20Options)
+will include them, such as: `--output verbose` or `--output json`.
+All Icinga DB fields are prefixed with `ICINGADB_`, e.g., `ICINGADB_ERROR` for error messages.
+
+```
+journalctl --unit icingadb.service --output verbose
+```
```
julianbrost, line 145: I agree with https://github.com/Icinga/icingadb/pull/743#discussion_r1703623929 in that I'd be happy with just `--output verbose` (i.e. no full command line example).

However, if you add an example, please try to somewhat incorporate it into the text, currently it looks a bit lost at the end. Maybe something like this:

````suggestion
When examining the journal with `journalctl`, fields containing additional information are hidden by default.
Setting an appropriate
[`--output` option](https://www.freedesktop.org/software/systemd/man/latest/journalctl.html#Output%20Options)
such as `--output verbose` or `--output json` will include them, for example:

```
journalctl --unit icingadb.service --output verbose
```

All Icinga DB fields are prefixed with `ICINGADB_`, e.g., `ICINGADB_ERROR` for error messages.
````

yhabteab (approved):

PR #747, opened 2024-04-19
Library split

lippserd:
Closes #654


pkg/icingadb/db.go
```diff
@@ -542,7 +542,7 @@ func (db *DB) CreateIgnoreStreamed(
 	ctx context.Context, entities <-chan database.Entity, onSuccess ...OnSuccess[database.Entity],
 ) error {
 	first, forward, err := com.CopyFirst(ctx, entities)
-	if first == nil {
+	if err != nil {
 		return errors.Wrap(err, "can't copy first entity")
 	}
```
julianbrost: Changing the error handling in this way sounds like a good idea (given that the function returns `error` already), but then the following case within the function should probably be changed to also return an error in this case:

https://github.com/Icinga/icingadb/blob/5d182aad64891e7007b7d90320bccce0c933d437/pkg/com/com.go#L52-L54

pkg/database/db.go
```diff
@@ -566,19 +649,19 @@ func (db *DB) YieldAll(ctx context.Context, factoryFunc contracts.EntityFactoryF
...
 // Entities for which the query ran successfully will be passed to onSuccess.
 func (db *DB) CreateStreamed(
-	ctx context.Context, entities <-chan contracts.Entity, onSuccess ...OnSuccess[contracts.Entity],
+	ctx context.Context, entities <-chan Entity, onSuccess ...OnSuccess[Entity],
 ) error {
 	first, forward, err := com.CopyFirst(ctx, entities)
-	if first == nil {
+	if err != nil {
 		return errors.Wrap(err, "can't copy first entity")
 	}
```
julianbrost, line 657: Changing the error handling in this way sounds like a good idea (given that the function returns `error` already), but then the following case within the function should probably be changed to also return an error in this case:

https://github.com/Icinga/icingadb/blob/5d182aad64891e7007b7d90320bccce0c933d437/pkg/com/com.go#L52-L54

pkg/icingadb/db.go
lippserd: True, fixed.

pkg/types/unix_milli.go
```diff
@@ -30,32 +29,32 @@ func (t UnixMilli) MarshalJSON() ([]byte, error) {
...
+// MarshalText implements the encoding.TextMarshaler interface.
+func (t UnixMilli) MarshalText() ([]byte, error) {
+	if time.Time(t).IsZero() {
+		return []byte{}, nil
 	}
-	tt := utils.FromUnixMilli(int64(ms))
-	*t = UnixMilli(tt)
 
-	return nil
+	return []byte(strconv.FormatInt(time.Time(t).UnixMilli(), 10)), nil
```
julianbrost, line 48: Isn't the `Time()` method there to make this cast more convenient, i.e. `t.Time()` instead of `time.Time(t)`?

pkg/types/unix_milli_test.go
```diff
@@ -28,3 +28,69 @@ func TestUnixMilli_MarshalJSON(t *testing.T) {
...
+			require.Equal(t, st.input, s)
+		})
+	}
+}
+
+func TestUnixMilli_MarshalText(t *testing.T) {
+	subtests := []struct {
+		name   string
+		input  UnixMilli
+		output string
```
julianbrost, line 58: The input/output variables are mixed up here. 

Also, the pretty much same test data is repeated multiple times here. Maybe create a global variable for it that's something like []struct{name, time, json, text}` (if JSON and text representation even differ anywhere, otherwise a common field for both).

pkg/icingadb/history/sla.go
```diff
@@ -1,14 +1,22 @@
...
 
-var slaStateStructify = structify.MakeMapStructifier(reflect.TypeOf((*history.SlaHistoryState)(nil)).Elem(), "json")
+var slaStateStructify = structify.MakeMapStructifier(
+	reflect.TypeOf((*history.SlaHistoryState)(nil)).Elem(),
+	"json",
+	func(a any) {
+		if initer, ok := a.(contracts.Initer); ok {
+			initer.Init()
+		}
+	})
```
julianbrost, line 19: This inline function appears quite a few times. Define it once in a common place?

pkg/strcase/strcase.go
```diff
@@ -0,0 +1,70 @@
...
+					n.WriteRune(d)
+				}
+			} else if isNumber || wasLower && unicode.IsUpper(r) {
+				n.WriteRune(d)
+			}
+
+			n.WriteRune(unicode.To(_case, r))
+		} else if delimiter != d {
+			n.WriteRune(d)
+		}
```
julianbrost, line 61: I'd just go for a `case ' ', '_', '-', '.':` instead of the `delimiters` map here.

Also, the value type of that map is `any`, thus the local variable `delimiter` is of type any and is either `nil` or `struct{}{}`. `d` on the other hand is of type `rune`, so `else if delimiter != d` will never be true.

pkg/logging/logging.go
```diff
@@ -117,15 +118,19 @@ func (l *Logging) GetLogger() *Logger {
...
+//	}
+func Fatal(err error) {
+	// NewDevelopment to use the console encoder.
+	l, _ := zap.NewDevelopment()
+	defer func(l *zap.Logger) {
+		_ = l.Sync()
+	}(l)
+
+	l.Sugar().Fatal(err)
 }
```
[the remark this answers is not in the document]
lippserd: 1) Not using `panic()`.
2) Printing timestamp.
3) Printing stacktraces.

For 3) i'll push a fix as it should be `Fatalf("%+v", err)`.

pkg/strcase/strcase.go
lippserd: > I'd just go for a `case ' ', '_', '-', '.':` instead of the `delimiters` map here.

That's much simpler, yes. But I had in mind that at some point it might be necessary for the caller to extend the delimiter list. The current implementation would be prepared for this (at least a little bit) but I'll simplify it.

> ..., so `else if delimiter != d` will never be true.

It will never be `false`. This code path overrides considered delimiters with the desired one. So this should be just `else` or the code for `case ' ', '_', '-', '.':` in the simplified version.

julianbrost: > It will never be `false`.

Err yes, it was `!=`.

> But I had in mind that at some point it might be necessary for the caller to extend the delimiter list.

The implementation is fixed to letters and numbers anyways, could simply treating everything else as a separator be an option?

pkg/strcase/strcase.go
```diff
@@ -0,0 +1,70 @@
...
+
+	for _, r := range s[pos:] {
+		if unicode.IsUpper(r) && (unicode.IsNumber(prevRune) || unicode.IsLower(prevRune)) {
+			n.WriteRune(d)
+		}
+
+		n.WriteRune(unicode.To(_case, r))
+
+		prevRune = r
+	}
```
julianbrost, line 67: This can be simplified into a single loop (and there's `rune` as an alias for `int32` which is better fitting here):

```suggestion
	var prevRune rune

	for i, r := range s {
		if i > 0 && unicode.IsUpper(r) && (unicode.IsNumber(prevRune) || unicode.IsLower(prevRune)) {
			n.WriteRune(d)
		}

		n.WriteRune(unicode.To(_case, r))

		prevRune = r
	}
```

The `i > 0` check wouldn't even be necessary as `(unicode.IsNumber(0) || unicode.IsLower(0)) == false`, but I don't think it hurts to make this a bit more explicit.

Bonus: this fixes a panic with input `"\xff"`:
```
panic: runtime error: slice bounds out of range [3:1]
```

If the input starts with invalid UTF-8, `for ... range` will iterate over it as `r = '�'` which is 3 bytes long, so `pos += utf8.RuneLen(r)` will add 3 which may be beyond the end of the string.

This requires the current `{"\xff\xfe\xfd", "�"}` test case to `{"\xff\xfe\xfd", "���"}` instead. This looks fine to me, at least I don't see a reason why these bytes should be treated as a single character. Treating it as three characters also matches the Go iteration behavior (well, that's basically the reason why it becomes three unicode replacement characters).

julianbrost (changes requested):

pkg/strcase/strcase.go
lippserd: Makes sense.

pkg/logging/logging.go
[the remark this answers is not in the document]
julianbrost: Is printing stack traces even necessary here? The use of the function is limited to the early startup (as also stated by its doc comment), so wouldn't they be something like "main calls whatever it says in the error message" all the time, so that there isn't really any additional information in the stack trace?

For me, seeing a stack trace always screams "the software is broken", so it should only print one in unexpected cases. During the initialization phase, when config is loaded etc., errors due to wrong configuration are to be expected and throwing a stack trace at the user just distracts from the problem. I don't think that the following change in behavior is an improvement:

### main branch
```console
$ ./icingadb --doesnotexist
unknown flag `doesnotexist'
```

### This PR
```console
$ ./icingadb --doesnotexist
unknown flag `doesnotexist'
2024-05-08T12:42:06.441+0200	FATAL	logging/logging.go:135	unknown flag `doesnotexist'
can't parse CLI flags
github.com/icinga/icingadb/pkg/config.ParseFlags
	/home/jbrost/src/icingadb/pkg/config/config.go:63
github.com/icinga/icingadb/internal/command.New
	/home/jbrost/src/icingadb/internal/command/command.go:27
main.run
	/home/jbrost/src/icingadb/cmd/icingadb/main.go:41
main.main
	/home/jbrost/src/icingadb/cmd/icingadb/main.go:37
runtime.main
	/usr/lib/go/src/runtime/proc.go:271
runtime.goexit
	/usr/lib/go/src/runtime/asm_amd64.s:1695
github.com/icinga/icingadb/pkg/logging.Fatal
	/home/jbrost/src/icingadb/pkg/logging/logging.go:135
main.run
	/home/jbrost/src/icingadb/cmd/icingadb/main.go:43
main.main
	/home/jbrost/src/icingadb/cmd/icingadb/main.go:37
runtime.main
	/usr/lib/go/src/runtime/proc.go:271
```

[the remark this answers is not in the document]
lippserd: Yes, a stack trace is not necessary in this case. This is an oversight. But if `defaults.Set()` in `FromYAMLFile()` fails, or if the argument passed to `FromYAMLFile()`/`ParseFlags()` is invalid, I would like to see one. In such cases, we could also just use `panic()`. What do you think?

For printing all the other errors we have two options:

1) ```can't parse CLI flags: unknown flag `doesnotexist'```

2) ```2024-05-08T13:55:41.738+0200    ERROR    logging/logging.go:138    can't parse CLI flags: unknown flag `doesnotexist'```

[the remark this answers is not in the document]
julianbrost: I prefer 1. If this function is used, it's most likely the only thing that will be printed by this process (logging wasn't initialized, so no previous log messages where you could compare times with). If this happens when the process was manually started in the foreground, you see the error immediately, if it happens from within a systemd unit, journald will track the time of the message already.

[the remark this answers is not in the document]
lippserd: Done.

julianbrost (changes requested):
Just a few small things I noticed. I think we can start doing the actual move soon. What's the plan here? There seems to be no clear pattern like `pkg/` will be moved as a whole, is each sub-package planned to either be moved or kept here (hence also the one "is this supposed to be moved?" question)?

julianbrost (commented):
Looks fine for me now. (Not formally clicking approve here as you wanted to still clean up the commit history. And maybe even do the switchover to the library in this PR?)

internal/config/config.go
```diff
@@ -0,0 +1,77 @@
...
+	Redis     redis.Config    `yaml:"redis"`
+	Logging   logging.Config  `yaml:"logging"`
+	Retention RetentionConfig `yaml:"retention"`
+}
+
+func (c *Config) SetDefaults() {
+	// Since SetDefaults() is called after the default values of the struct's fields have been evaluated,
+	// setting the default port only works here because
+	// the embedded Redis config struct itself does not provide a default value.
+	if defaults.CanUpdate(c.Redis.Port) {
```
yhabteab, line 25: `defaults.CanUpdate()` uses some `reflect` magic to determine this, but since that field is of type `int`, you'll never be able to distinguish whether it's set to `0` explicitly, since its zero initialised value is also `0`. So I suggest replacing this with a simple `c.Redis.Port == 0` instead. Apart from that, can you please define the default port for the database type as well, so that we no longer get [such confusing log messages](https://github.com/Icinga/icinga-go-library/issues/11).

pkg/redis/client.go
```diff
@@ -26,43 +29,53 @@ type Client struct {
...
-		return errors.New("xread_count must be at least 1")
+
+	if utils.IsUnixAddr(c.Host) {
+		options.Network = "unix"
+		options.Addr = c.Host
+	} else {
+		port := c.Port
+		if port == 0 {
+			port = 6379
+		}
```
yhabteab, line 68: The default value does not match the one set in `config#SetDefaults()`, and that variable is never used afterwards. However, do you even need this check at all, given that `config#SetDefaults()` takes care of it? It might also be reasonable to let Icinga DB fail in `Redis#Config#Validate()` if the user explicitly set this to `0`, as this is not a valid port.

internal/config/database.go
```diff
@@ -77,12 +88,11 @@ func (d *Database) Open(logger *logging.Logger) (*icingadb.DB, error) {
 			return nil, errors.Wrap(err, "can't open mysql database")
 		}
 
-		wsrepSyncWait := int64(d.Options.WsrepSyncWait)
-		setWsrepSyncWait := func(ctx context.Context, conn driver.Conn) error {
-			return setGaleraOpts(ctx, conn, wsrepSyncWait)
+		connectorCallbacks.OnInitConn = func(ctx context.Context, conn driver.Conn) error {
+			return setGaleraOpts(ctx, conn, int64(d.Options.WsrepSyncWait))
```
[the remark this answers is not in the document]
lippserd: We use almost all options concurrently. Though it makes sense to make `Options` private.

internal/config/config.go
lippserd: `internal` won't be moved to the lib so a change here won't help for that specific issue as you would have to copy `SetDefaults()` to your code. I would rather introduce something like `redis#Client.GetAddr()` and `database#DB.GetAddr()`.

internal/config/database.go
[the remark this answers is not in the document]
yhabteab: > Though it makes sense to make Options private.

By "private" do you mean unexported? If so, the Yaml parser will not be able to decode config into it and usages like `db.Options.MaxPlaceholdersPerStatement` outside the database package are not going to work as well.

internal/config/config.go
yhabteab: > I would rather introduce something like redis#Client.GetAddr() and database#DB.GetAddr().

That sounds good 👍!

internal/config/database.go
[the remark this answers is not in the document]
lippserd: > If so, the Yaml parser will not be able to decode config into it

I did not refer to the `Database#Config` struct. The YAML parser would not be a problem.

> and usages like `db.Options.MaxPlaceholdersPerStatement` outside the database package are not going to work as well.

That's right, and that also applies to the Redis options. So that's not the solution. But again, to be honest, I wouldn't worry too much about using options concurrently, as we basically do this for every option.

internal/config/config.go
lippserd: `redis#Client` already provides the `String()` method which exposes the addr. I also implemented `database#DB.String()` and use them for logging the initial connections:

> ...	Connecting to Database<localhost:3306 db:icingadb driver:mysql>
> ... Connecting to Redis<localhost:6380 db:0>

julianbrost: IMHO that format looks strange. It seems to be chosen because `redis.Client.String()` returns `"Redis<localhost:6380 db:0>"`? Hint: You can actually get just the address, it's just that our struct definition shadows the required method, but `rc.Client.Options().Addr` should do the trick.

For a more structured logging approach, something like https://github.com/Icinga/icinga-notifications/pull/198/commits/eb18f970de441954d202f72c0641319deda27c3b could be an option.

julianbrost (changes requested):

pkg/database/db.go
```diff
@@ -83,61 +92,138 @@ func (o *Options) Validate() error {
...
 
-	if version != expectedDbSchemaVersion {
-		// Since these error messages are trivial and mostly caused by users, we don't need
-		// to print a stack trace here. However, since errors.Errorf() does this automatically,
-		// we need to use fmt instead.
-		return fmt.Errorf(
-			"unexpected database schema version: v%d (expected v%d), please make sure you have applied all database"+
-				" migrations after upgrading Icinga DB", version, expectedDbSchemaVersion,
-		)
+			query["port"] = []string{strconv.FormatInt(int64(c.Port), 10)}
```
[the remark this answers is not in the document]
lippserd: > You probably want to use `port` instead of `c.Port` here.

🤦

> Also, with PostgreSQL, the port setting is relevant even when using unix domain sockets.

Thanks, will be fixed.

> What you configure as the socket path everywhere is actually just the directory, the actual socket path is something like ...

So for PostgreSQL we will only log the directory.

cmd/icingadb/main.go
```diff
@@ -64,14 +60,14 @@ func run() int {
 	}
 	defer db.Close()
 	{
-		logger.Infof("Connecting to database at '%s'", utils.JoinHostPort(cmd.Config.Database.Host, cmd.Config.Database.Port))
+		logger.Infof("Connecting to Database at '%s'", db.GetAddr())
```
julianbrost, line 63: Not sure if this is upper-cased because it was forgotten from the previous iteration or for consistency with Redis being written in upper-case. I'd keep it lower-cased here as database is not a proper name.

julianbrost (commented):
Now same level of approval from my side as in https://github.com/Icinga/icingadb/pull/747#pullrequestreview-2055264405:
> Looks fine for me now. (Not formally clicking approve here as you wanted to still clean up the commit history. And maybe even do the switchover to the library in this PR?)

internal/config/config.go
yhabteab: > defaults.CanUpdate() uses some reflect magic to determine this, but since that field is of type int, you'll never be able to distinguish whether it's set to 0 explicitly, since its zero initialised value is also 0. So I suggest replacing this with a simple c.Redis.Port == 0 instead.

No reaction to this at all?

lippserd: `defaults.CanUpdate()` and `defaults.CanUpdate()` in combination with `SetDefaults()` is the logical choice for handling default values as both is package functionality. If we ever replace this, we can look for such usages. Also, at first glance, it's pretty obvious what's happening, rather than having to recognize the initial value check. I have the impression that you are not happy with the reflection usage. But that's how the package works anyway. And if there were more such checks with different types, would you replace them with individual zero checks as well? No, because that's what this method is for.

yhabteab:
> Now same level of approval from my side as in [#747 (review)](https://github.com/Icinga/icingadb/pull/747#pullrequestreview-2055264405):
> 
> > Looks fine for me now. (Not formally clicking approve here as you wanted to still clean up the commit history. And maybe even do the switchover to the library in this PR?)

Same! I have nothing to complain about anymore!

lippserd:
I cleaned up the commit history and switched to using `icinga-go-library` in this PR.
Note that the new lib is [here](https://github.com/Icinga/icinga-go-library/tree/new) at the moment. Once you've approved, I'll force the push to main and tag it as version 1.0.0 and change `go.mod` here accordingly.

yhabteab:
Looks like something went wrong with the rebasing? 1ddaa70, 1d9dc74 and 563045c seem to appear twice in the commit history.

yhabteab:
> Looks like something went wrong with the rebasing? [1ddaa70](https://github.com/Icinga/icingadb/commit/1ddaa70aabe1b28371c12b6df279bd06941ca87d), [1d9dc74](https://github.com/Icinga/icingadb/commit/1d9dc74391a1cde2310b174a760395aa3a4cebfc) and [563045c](https://github.com/Icinga/icingadb/commit/563045cc5a0afe3f35d3fcd46170d14d7f2da790) seem to appear twice in the commit history.

Ohh 😧! Never mind! It's GitHub's fault!

pkg/config/config.go
```diff
@@ -39,16 +47,30 @@ func FromYAMLFile[T any, P interface {
...
+// If either option is specified on the command line,
+// ParseFlags prints the help message to [os.Stdout] and exits.
+func ParseFlags(v any) error {
+	rv := reflect.ValueOf(v)
+	if rv.Kind() != reflect.Pointer || rv.IsNil() {
+		return errors.Wrapf(ErrInvalidArgument, "non-nil pointer expected, got %T", v)
+	}
 
-	parser := flags.NewParser(&f, flags.Default)
+	parser := flags.NewParser(v, flags.Default^flags.PrintErrors)
```
yhabteab: It would be nice if you could provide some comments why you are clearing the `flags.PrintErrors` bits. If I understand it correctly, you are doing this so that the help message is not printed twice since you are explicitly printing it. However, wouldn't it be sufficient to rely on the library and just exit down below if the error is of type `flags.ErrHelp` without having to clear that flag?

internal/config/config.go
```diff
@@ -38,3 +41,26 @@ type Flags struct {
...
+}
+
+// Validate checks constraints in the supplied retention configuration and
+// returns an error if they are violated.
+func (r *RetentionConfig) Validate() error {
+	if r.Interval <= 0 {
+		return errors.New("retention interval must be positive")
+	}
+
+	if r.Count == 0 {
```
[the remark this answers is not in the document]
lippserd: Can you open an issue for that please? I don't think this PR is the right place to fix that.

pkg/config/config.go
lippserd: The function returns either `ErrInvalidArgument` or a flag-related error. In the latter case, the library would print the error and the function would then return this error, which we would then print again. Therefore, the actual error handling is the responsibility of the caller. I can mention this briefly in the documentation.

julianbrost (commented):
Fine for me now. At least on the "does not seem to break anything" level. And I wouldn't go into more clean this up side-quests here unless it fixes something that's worse than the current state in Icinga DB. This PR is huge and cumbersome to work with already.

> Once you've approved, I'll force the push to main and tag it as version 1.0.0 and change `go.mod` here accordingly.

If in doubt, it could also be tagged as v0.1.0, then do some more few PRs in the new repo and then make it a 1.0.

internal/config/config.go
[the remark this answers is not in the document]
yhabteab: Done https://github.com/Icinga/icingadb/issues/755

lippserd:
> > Once you've approved, I'll force the push to main and tag it as version 1.0.0 and change `go.mod` here accordingly.
> 
> If in doubt, it could also be tagged as v0.1.0, then do some more few PRs in the new repo and then make it a 1.0.

Yeah, I started with `v0.1.0`.

julianbrost (approved):

yhabteab (approved):
I found one last thing in the [new library](https://github.com/Icinga/icinga-go-library/blob/main/version/version.go#L83) that should be changed, otherwise this LFTM!

julianbrost:
Not really relevant for Icinga DB, will be fixed and become a v0.2.0.

Issue #750, opened 2024-04-23
Setting wsrep_sync_wait via prepared statement is incompatible with ProxySQL


glbyers:
## Describe the bug

With the release of Icinga DB 1.2.0, new functionality was added to always attempt to set the `wsrep_sync_wait` session variable on connection. This is done as a prepared statement, which is incompatible with ProxySQL; [Currently ProxySQL doesn't support the parsing and tracking of SET statements using the binary protocol (prepared statements)](https://github.com/sysown/proxysql/issues/3622#issuecomment-921167155).

With ProxySQL in front of Galera, this doesn't achieve the desired outcome of transaction consistency; `wsrep_sync_wait` never gets set. We can't simply rewrite this query with ProxySQL query rewrite feature either, as it's a prepared statement.

Aside from this not achieving the desired outcome, ProxySQL logs become flooded with errors. ie;

`2024-04-23 22:05:13 MySQL_Session.cpp:7064:handler___status_WAITING_CLIENT_DATA___STATE_SLEEP___MYSQL_COM_QUERY_qpo(): [ERROR] Unable to parse query. If correct, report it as a bug: SET SESSION wsrep_sync_wait=?`

## To Reproduce

To reproduce, you can run ProxySQL in front of a single MariaDB/MySQL database & point Icinga DB to ProxySQL.

## Expected behavior

I'm expecting that whatever we set `database.options.wsrep_sync_wait` to (or the default of 7) is reflected as the value of the session variable. If this cannot be achieved by replacing the prepared statement with a standard `ExecContext` type query, we hope that it may be possible to set `database.options.wsrep_sync_wait` to a value that tells Icinga DB to ignore the setting and never try to set it (we can take care of it instead via mysql client configuration). ie.

```yaml
database:
  options:
    ## Ignore setting. We'll deal with this in /etc/my.cnf.d/client.cnf instead
    wsrep_sync_wait: -1
```

## Your Environment

Icinga DB 1.2.0
Icinga 2.1.4
RHEL 8.9

## Additional context

Functionality to add setting of `wsrep_sync_wait` added in [this commit](https://github.com/Icinga/icingadb/commit/4d0b58cfb4bfabcdcba0063aed6f4e492c9dee50).

I understand this is a limitation of ProxySQL and not IcingaDB, but given this is not likely to be addressed in ProxySQL in the short term, i'm hopping for a workaround in IcingaDB.

glbyers:
FYI, worked around with the following for now;

```diff
[...]

PR #751, opened 2024-04-24
setGaleraOpts(): execute SET SESSION wsrep_sync_wait directly

Al2Klimov:
to save extra roundtrips for preparing and closing a statement.

* Idea (c) @glbyers #750

## Test

```
000001F4  1e 00 00 00 03 53 45 54  20 53 45 53 53 49 4f 4e   .....SET  SESSION
00000204  20 77 73 72 65 70 5f 73  79 6e 63 5f 77 61 69 74    wsrep_s ync_wait
00000214  3d 37                                              =7
    00000231  32 00 00 01 ff a9 04 23  48 59 30 30 30 55 6e 6b   2......# HY000Unk
    00000241  6e 6f 77 6e 20 73 79 73  74 65 6d 20 76 61 72 69   nown sys tem vari
    00000251  61 62 6c 65 20 27 77 73  72 65 70 5f 73 79 6e 63   able 'ws rep_sync
    00000261  5f 77 61 69 74 27                                  _wait'
00000216  01 00 00 00 0e                                     .....
    00000267  07 00 00 01 00 00 00 02  00 01 00                  ........ ...
0000021B  3d 00 00 00 03 53 45 4c  45 43 54 20 76 65 72 73   =....SEL ECT vers
0000022B  69 6f 6e 20 46 52 4f 4d  20 69 63 69 6e 67 61 64   ion FROM  icingad
0000023B  62 5f 73 63 68 65 6d 61  20 4f 52 44 45 52 20 42   b_schema  ORDER B
0000024B  59 20 69 64 20 44 45 53  43 20 4c 49 4d 49 54 20   Y id DES C LIMIT 
0000025B  31                                                 1
```

👍

### DIFF
--- a/pkg/config/database.go
+++ b/pkg/config/database.go
@@ -189,27 +189,15 @@ func unknownDbType(t string) error {
 //
 // https://mariadb.com/kb/en/galera-cluster-system-variables/#wsrep_sync_wait
 func setGaleraOpts(ctx context.Context, conn driver.Conn, wsrepSyncWait int64) error {
-	const galeraOpts = "SET SESSION wsrep_sync_wait=?"
+	galeraOpts := "SET SESSION wsrep_sync_wait=" + strconv.FormatInt(wsrepSyncWait, 10)
 
-	stmt, err := conn.(driver.ConnPrepareContext).PrepareContext(ctx, galeraOpts)
+	_, err := conn.(driver.ExecerContext).ExecContext(ctx, galeraOpts, nil)
 	if err != nil {
 		if errors.Is(err, &mysql.MySQLError{Number: 1193}) { // Unknown system variable
 			return nil
 		}
 
-		return errors.Wrap(err, "cannot prepare "+galeraOpts)
-	}
-	// This is just for an unexpected exit and any returned error can safely be ignored and in case
-	// of the normal function exit, the stmt is closed manually, and its error is handled gracefully.
-	defer func() { _ = stmt.Close() }()
-
-	_, err = stmt.(driver.StmtExecContext).ExecContext(ctx, []driver.NamedValue{{Value: wsrepSyncWait}})
-	if err != nil {
 		return errors.Wrap(err, "cannot execute "+galeraOpts)
-	}
-
-	if err = stmt.Close(); err != nil {
-		return errors.Wrap(err, "cannot close prepared statement "+galeraOpts)
 	}
 
 	return nil


pkg/config/database.go
```diff
@@ -189,28 +189,16 @@ func unknownDbType(t string) error {
...
 func setGaleraOpts(ctx context.Context, conn driver.Conn, wsrepSyncWait int64) error {
-	const galeraOpts = "SET SESSION wsrep_sync_wait=?"
+	galeraOpts := "SET SESSION wsrep_sync_wait=" + strconv.FormatInt(wsrepSyncWait, 10)
 
-	stmt, err := conn.(driver.ConnPrepareContext).PrepareContext(ctx, galeraOpts)
+	_, err := conn.(driver.ExecerContext).ExecContext(ctx, galeraOpts, nil)
 	if err != nil {
 		if errors.Is(err, &mysql.MySQLError{Number: 1193}) { // Unknown system variable
 			return nil
 		}
```
oxzi, line 198: @glbyers' [patch](https://github.com/Icinga/icingadb/issues/750#issuecomment-2073765560) also swallowed [`driver.ErrSkip`](https://pkg.go.dev/database/sql/driver#ErrSkip). Even as the [`driver.ExecerContext` documentation](https://pkg.go.dev/database/sql/driver#ExecerContext) states that it might return this error, I am not sure why it was included in the patch. Maybe @glbyers can enlighten us?

oxzi (commented):
While I don't have anything against this patch and it even addresses a specific issue, I am unsure about the _framing_ to "save extra roundtrips". The issue at hand is that ProxySQL is not feature complete with MySQL and not that setting `wsrep_sync_wait` is slowing anything down.

pkg/config/database.go
Al2Klimov: Precisely speaking, it's this error that patch had swallowed: https://github.com/go-sql-driver/mysql/blob/d7ddb8b9e324830b1ede89c5fea090c824497c51/connection.go#L318

I also have no idea why to swallow this error because this makes the query a complete no-op.

Al2Klimov:
The issue was just an inspiration, not even something I claim to have fixed. See also https://github.com/go-sql-driver/mysql/blob/d7ddb8b9e324830b1ede89c5fea090c824497c51/connection.go#L320

I mean, yes, my patch allows ProxySQL to rewrite the SET to something else... 🤷‍♂️

pkg/config/database.go
glbyers: > @glbyers' [patch](https://github.com/Icinga/icingadb/issues/750#issuecomment-2073765560) also swallowed [`driver.ErrSkip`](https://pkg.go.dev/database/sql/driver#ErrSkip). Even as the [`driver.ExecerContext` documentation](https://pkg.go.dev/database/sql/driver#ExecerContext) states that it might return this error, I am not sure why it was included in the patch. Maybe @glbyers can enlighten us?

Right. I'd read that ErrSkip may be raised if interpolation wasn't available in the driver, but wasn't aware that it would result in a complete noop. In our situation, we're not concerned that `wsrep_sync_wait` isn't set at all as we control writes to a single node via proxysql (so wsrep_sync_wait=0 or not at all set suits our use case). Apologies for overlooking that, but as stated, that works just fine for us. I'll test your patch with interpolation on Monday when i'm back at work.

glbyers: OK. I have applied your patch and tested. I can confirm this works as expected. ProxySQL events log captures the query;

```json
{"client":"127.0.0.1:37024","digest":"0x4935002AA513D03A","duration_us":0,"endtime":"2024-04-28 21:45:44.090140","endtime_timestamp_us":1714340744090140,"event":"COM_QUERY","hostgroup_id":-1,"query":"SET SESSION wsrep_sync_wait=0","rows_affected":0,"rows_sent":0,"schemaname":"icingadb","starttime":"2024-04-28 21:45:44.090140","starttime_timestamp_us":1714340744090140,"thread_id":7,"username":"icingadb"}
```

glbyers:
@oxzi , what's required to get this one merged? Looks like there's now merge conflicts, unfotunately.

oxzi:
After the library split, the code in question was moved to the icinga-go-library. I have just recreated this PR over there as <https://github.com/Icinga/icinga-go-library/pull/51>.

PR #756, opened 2024-05-23
cmd/icingadb: if --config-from-env given, load config from env vars

Al2Klimov:


pkg/config/database.go
```diff
@@ -22,14 +22,14 @@ import (
...
-	Password   string           `yaml:"password"`
+	Type       string           `yaml:"type" env:"TYPE" default:"mysql"`
+	Host       string           `yaml:"host" env:"HOST"`
+	Port       int              `yaml:"port" env:"PORT"`
+	Database   string           `yaml:"database" env:"DATABASE"`
+	User       string           `yaml:"user" env:"USER"`
+	Password   string           `yaml:"password" env:"PASSWORD"`
 	TlsOptions TLS              `yaml:",inline"`
-	Options    icingadb.Options `yaml:"options"`
+	Options    icingadb.Options `yaml:"options" envPrefix:"OPTIONS_"`
```
Al2Klimov, line 32: @lippserd Can you live with all those struct tags?

PR #768, opened 2024-07-09
icingadb-migrate: Verify env ID length

oxzi:
An env ID with the wrong length, either due to a copy-paste error or human error during testing, results in a SQL CHECK CONSTRAINT violation that is retried multiple times until it finally fails.


Al2Klimov (approved):
Test: https://github.com/Icinga/icingadb/pull/769#issue-2398153664

PR #769, opened 2024-07-09
cmd/icingadb-migrate: replace magic number 20 with sha1.Size

Al2Klimov:
## Test

```
bad env ID: must be 20 bytes long, has 21 bytes
```

👍


Al2Klimov:
I'd let GH rebase-and-merge btw.

PR #770, opened 2024-07-09
cmd/icingadb: automatically upgrade SQL schema

Al2Klimov:


cmd/icingadb/main.go
```diff
@@ -67,8 +72,38 @@ func run() int {
...
+								{Version: "3", DDL: []string{mysql.Upgrade100}},
+								{Version: "4", DDL: []string{mysql.Upgrade111}},
+								{Version: "5", DDL: []string{mysql.Upgrade120}},
+							},
+						},
+						database.PostgreSQL: {
+							Schema: []string{pgsql.Schema},
+							Upgrades: []database.SchemaUpgrade{
+								{Version: "2", DDL: []string{pgsql.Upgrade111}},
+								{Version: "3", DDL: []string{pgsql.Upgrade120}},
```
Al2Klimov, line 94: * @lippserd This is a raw\* usage example of https://github.com/Icinga/icinga-go-library/pull/15

Do you consider the design fine?

\* One has to maintain such schema file lists. A non-raw usage requires the specific project to have a function which auto-calculates schema upgrade lists from the embedded tree of schema/.

schema/mysql/embed.go
oxzi, line 1: Generally speaking, I am against embedding both the schema files and all schema upgrades in Icinga DB as they are being installed as part of the package. This only bloats up the binary.

cmd/icingadb/main.go
```diff
@@ -67,8 +72,38 @@ func run() int {
...
+							},
+						},
+					},
+				)
+			},
+			retry.Retryable,
+			backoff.NewExponentialWithJitter(128*time.Millisecond, time.Minute),
+			db.GetDefaultRetrySettings(),
```
oxzi, line 102: Please correct me if I'm wrong, but this would force (multiple) schema upgrades to be performed in five minutes. If this takes more time, the user ends up in a very messy situation, potentially don't even knowing how they got there.

There were already long running schema upgrades in the past, <https://icinga.com/docs/icinga-db/latest/doc/04-Upgrading/#upgrading-to-icinga-db-v120>.

cmd/icingadb/main.go
```diff
@@ -67,8 +72,38 @@ func run() int {
 		}
 	}
 
-	if err := icingadb.CheckSchema(context.Background(), db); err != nil {
-		logger.Fatalf("%+v", err)
+	{
+		err := retry.WithBackoff(
+			context.Background(),
+			func(ctx context.Context) (err error) {
+				return database.AutoUpgradeSchema(
```
oxzi, line 79: I would love to see some logging here. The `AutoUpgradeSchema` function just starts upgrading the schema without any notice.

From a user's point of view, Icinga DB just "blocked" and Icinga Web will hint that Icinga DB might have stopped. Thus, even when a user have consulted the logs and sees nothing, they will likely restart Icinga DB and crash the migration.

schema/mysql/embed.go
Al2Klimov: OK, then let's don't ship the separate files. 👍

PR #771, opened 2024-07-12
GHA: also test MySQL 8

Al2Klimov:


oxzi:
Why was this labeled as a _bug_?

julianbrost (approved):
That change feels a bit arbitrary as it doesn't make the list complete/up-to-date again, but it doesn't hurt either, so why not?

oxzi (approved):

PR #777, opened 2024-07-22
Test icingaredis.CreateEntities()

Al2Klimov:


pkg/icingaredis/utils_test.go
```diff
@@ -0,0 +1,159 @@
...
+						if st.error && i == len(st.output)-1 {
+							errs = nil
+						}
+
+						select {
+						case actual, ok := <-entities:
+							if !ok {
+								require.Fail(t, "entities channel should not be closed, yet")
+							}
+
```
oxzi, line 116: You can shorten all the `if !ok { ... }` cases like this:

```suggestion
							require.True(t, ok, "entities channel should not be closed, yet")
```

oxzi (commented):
Thanks for creating this PR. I have looked over the test code and tried it. Seems to work!

PR #797, opened 2024-09-04
Make structifyStream() channel types more explicit, e.g. input only

Al2Klimov:
This enforces which side (caller/function) can read and especially    close/write to a particular channel at compile time.    That leaves less room for bugs (which had to be detected by tests otherwise).

### DIFF
--- a/pkg/icingadb/runtime_updates.go
+++ b/pkg/icingadb/runtime_updates.go
@@ -301,17 +301,24 @@ func (r *RuntimeUpdates) xRead(ctx context.Context, updateMessagesByKey map[stri
 // those messages into Icinga DB entities (contracts.Entity) using the provided structifier.
 // Converted entities are inserted into the upsertEntities or deleteIds channel depending on the "runtime_type" message field.
 func structifyStream(
-	ctx context.Context, updateMessages <-chan redis.XMessage, upsertEntities, upserted chan database.Entity,
-	deleteIds, deleted chan interface{}, structifier structify.MapStructifier,
+	ctx context.Context,
+	updateMessages <-chan redis.XMessage,
+	upsertEntities chan<- database.Entity,
+	upserted <-chan database.Entity,
+	deleteIds chan<- any,
+	deleted <-chan any,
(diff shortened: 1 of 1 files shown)


pkg/icingadb/runtime_updates.go
```diff
@@ -301,17 +301,19 @@ func (r *RuntimeUpdates) xRead(ctx context.Context, updateMessagesByKey map[stri
 // those messages into Icinga DB entities (contracts.Entity) using the provided structifier.
 // Converted entities are inserted into the upsertEntities or deleteIds channel depending on the "runtime_type" message field.
 func structifyStream(
-	ctx context.Context, updateMessages <-chan redis.XMessage, upsertEntities, upserted chan database.Entity,
-	deleteIds, deleted chan interface{}, structifier structify.MapStructifier,
+	ctx context.Context, updateMessages <-chan redis.XMessage, upsertEntities chan<- database.Entity,
+	upserted <-chan database.Entity, deleteIds chan<- any, deleted <-chan any, structifier structify.MapStructifier,
```
oxzi, line 305: The parameter list got quite long and is imo hard to read in its current format. Would you mind reformatting it, having one parameter each line.

oxzi (approved):

yhabteab:
I don't understand what this change is supposed to achieve! Previously, nil channels were initialised and immediately closed, and with PR nil channels are assigned to a new closed one. So, which issue is this supposed to fix?

julianbrost:
> I don't understand what this change is supposed to achieve!

Makes it more clear in the function signature, which channels are used for input and which for output.

> Previously, nil channels were initialised and immediately closed, and with PR nil channels are assigned to a new closed one. So, which issue is this supposed to fix?

That probably became necessary as otherwise, once you assign a channel to `upserted <-chan database.Entity` for example, you can no longer close it using that `<-chan` reference. Therefore, it's assigned to a temporary variable, closed, then assigned to the final variable that no longer allows closing.

lippserd (changes requested):
Commit and PR description missing.

PR #799, opened 2024-09-05
Test icingaredis.SetChecksums()

Al2Klimov:


pkg/icingaredis/utils_test.go
```diff
@@ -147,3 +149,201 @@ func TestCreateEntities(t *testing.T) {
...
+							input := make(chan database.Entity, 1)
+							go func() {
+								defer close(input)
+
+								for _, v := range st.input {
+									if l.latency > 0 {
+										select {
+										case <-time.After(l.latency):
+										case <-ctx.Done():
+											return
```
oxzi, line 256: Indentation depth eleven. We're hitting indentation levels that shouldn't even be possible :>

oxzi (commented):
For both functions, tests against a prematurely canceled context are missing.

Al2Klimov:
If the input channel blocks since its creation forever and the context has always been canceled, do you expect the output channel to get closed immediately?

```patch
--- pkg/icingaredis/utils_test.go
+++ pkg/icingaredis/utils_test.go
@@ -346,4 +346,29 @@ func TestSetChecksums(t *testing.T) {
                        }
                })
        }
+
+       t.Run("cancel-ctx", func(t *testing.T) {
+               ctx, cancel := context.WithCancel(context.Background())
+               cancel()
+
+               output, errs := SetChecksums(ctx, make(chan database.Entity), map[string]database.Entity{}, 1)
+
+               require.NotNil(t, output, "output channel should not be nil")
+               require.NotNil(t, errs, "error channel should not be nil")
+
+               select {
+               case v, ok := <-output:
+                       require.False(t, ok, "output channel should be closed, got %#v", v)
+               case <-time.After(time.Second):
+                       require.Fail(t, "output channel should not block")
+               }
+
+               select {
+               case err, ok := <-errs:
+                       require.True(t, ok, "error channel should not be closed, yet")
+                       require.Error(t, err)
+               case <-time.After(time.Second):
+                       require.Fail(t, "error channel should not block")
+               }
+       })
 }
```

oxzi:
> If the input channel blocks since its creation forever and the context has always been canceled, do you expect the output channel to get closed immediately?

I guess so, yes. However, the current implementation should skip closing the output channel, as it happens deferred in the `errgroup` goroutine.

PR #800, opened 2024-09-05
Abort HA Realization Logic After Timeout

### LINKED ISSUE #787: Competing HA takeover results in both instances becoming active
## Describe the bug

Apparently there is a race condition in our HA realize handlings that can occur from time to time, causing both Icinga DB instances to become active/responsible at the same time. It is possible for the following to happen if both instances access the `icingadb_instance` table at the same time.

Since both instances are trying to update that table quite often[^1], it is to be expected that they will deadlock each other, as can be seen in the logs.
```json
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:57:43",
    "MESSAGE": "high-availability: Can't update or insert instance. Retrying",
    "ICINGADB_ERROR": "can't perform \"INSERT INTO \\\"icingadb_instance\\\"...: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction"
}
```

However, this retry mechanism also prevents new heartbeat events from being consumed from the channel and results in such weird situation.
```json
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:57:41",
    "ICINGADB_PREVIOUS": "2024-07-25 11:57:38.433795293 +0200 CEST",
    "ICINGADB_CURRENT": "2024-07-25 11:57:41.433328734 +0200 CEST",
    "MESSAGE": "heartbeat: Previous heartbeat not read from channel",
}
```

In order to prevent it from being retried endlessly, though, we have a deadline[^2] for the entire HA realize thing.
https://github.com/Icinga/icingadb/blob/f9ce7a57bcb6a5b6b2ebfafafd8f91367e56bbf5/pkg/icingadb/ha.go#L220-L221

Unfortunately, if all this retry error mess then lasts for exactly one minute, and barely manages to survive the context deadline, it becomes in the end a wasted time, as after this delay the heartbeat of a given instance is already outdated, but has nevertheless been written to the database. Obviously, the current instance remains responsible, it just has written an outdated heartbeat timestamp to the database because of the deadlocks. When the passive instance finally reads that timestamp, it thinks that the other instance is no longer available and takes over the HA responsibility.
https://github.com/Icinga/icingadb/blob/f9ce7a57bcb6a5b6b2ebfafafd8f91367e56bbf5/pkg/icingadb/ha.go#L319-L321

Having two active instances, however, does not produce any good results and brings down one of the two services.

<details><summary>Active IcingaDB Instance</summary>

```json
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:58:32",
    "ICINGADB_PREVIOUS": "2024-07-25 11:58:29.433879058 +0200 CEST",
    "ICINGADB_CURRENT": "2024-07-25 11:58:32.433682676 +0200 CEST",
    "MESSAGE": "heartbeat: Previous heartbeat not read from channel",
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:58:35",
    "ICINGADB_PREVIOUS": "2024-07-25 11:58:32.433682676 +0200 CEST",
    "ICINGADB_CURRENT": "2024-07-25 11:58:35.433895608 +0200 CEST",
    "MESSAGE": "heartbeat: Previous heartbeat not read from channel",
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:58:38",
    "ICINGADB_PREVIOUS": "2024-07-25 11:58:35.433895608 +0200 CEST",
    "ICINGADB_CURRENT": "2024-07-25 11:58:38.432917241 +0200 CEST",
    "MESSAGE": "heartbeat: Previous heartbeat not read from channel",
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:58:41",
    "MESSAGE": "high-availability: Instance updated/inserted successfully after error",
    "ICINGADB_AFTER": "1m9.144891965s",
    "ICINGADB_RECOVERED_ERROR": "can't perform \"INSERT INTO \\\"icingadb_instance\\\"...: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction",
}
```

Take a closer look at the `ICINGADB_AFTER` field of the above log entry `"ICINGADB_AFTER": "1m9.144891965s"` originated from this part of the code.
https://github.com/Icinga/icingadb/blob/f9ce7a57bcb6a5b6b2ebfafafd8f91367e56bbf5/pkg/icingadb/ha.go#L410-L413

The other(passive) Instance uses the heartbeat timestamp of this Instance from the database to determine that this currently active Instance is absent, by verifying that our timestamp does not exceed the `1m` + grace period of **5s** limit. However, this instance has spent over a minute and 5 seconds (to be precise, it took `1m9.144891965s`) to record the heartbeat, so that it immediately exceeded the aforementioned limit right after the upsert.
https://github.com/Icinga/icingadb/blob/f9ce7a57bcb6a5b6b2ebfafafd8f91367e56bbf5/pkg/icingadb/ha.go#L319 https://github.com/Icinga/icingadb/blob/f9ce7a57bcb6a5b6b2ebfafafd8f91367e56bbf5/pkg/icingadb/ha.go#L25-L28

Subsequently, the other Instance takes over the HA responsibility, even though this one is actually still responsible.

</details> 

<details><summary>Passive IcingaDB Instance</summary>

```json
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:58:33",
    "MESSAGE": "high-availability: Can't update or insert instance. Retrying",
    "ICINGADB_ERROR": "can't commit transaction: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction"
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:58:41",
    "ICINGADB_PREVIOUS": "2024-07-25 11:58:38.302404535 +0200 CEST",
    "ICINGADB_CURRENT": "2024-07-25 11:58:41.299007643 +0200 CEST",
    "MESSAGE": "heartbeat: Previous heartbeat not read from channel",
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:58:43",
    "ICINGADB_HEARTBEAT": "2024-07-25 11:57:32.259000062 +0200 CEST",
    "ICINGADB_HEARTBEAT_AGE": "1m11.394978989s",
    "MESSAGE": "high-availability: Preparing to take over HA as other instance's heartbeat has expired",
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:58:44",
    "ICINGADB_PREVIOUS": "2024-07-25 11:58:41.299007643 +0200 CEST",
    "ICINGADB_CURRENT": "2024-07-25 11:58:44.298509894 +0200 CEST",
    "MESSAGE": "heartbeat: Previous heartbeat not read from channel",
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:59:15",
    "MESSAGE": "high-availability: Instance updated/inserted successfully after error",
    "ICINGADB_AFTER": "1m3.721497393s",
    "ICINGADB_RECOVERED_ERROR": "can't commit transaction: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction",
},
```

Afterwards, the passive instance also becomes active and starts its initial config sync and encounters numerous duplicate key and deadlock errors until one of the queries exceeds the retry limit of `5m` and crashes the Icinga DB process.

```json
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 11:59:20",
    "MESSAGE": "dump-signals: Received dump signal from Redis",
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 12:00:20",
    "MESSAGE": "database: Can't execute query. Retrying",
    "ICINGADB_ERROR": "can't perform \"INSERT INTO \\\"host_customvar\\\"...: Error 1062 (23000): Duplicate entry '\\xDE\\xF7i\\xF9\\x9A\\xFEz\\xF6\\x8B\\x107\\xAAE\\xF0\\xBE\\xFF\\xA1Z\\xD3...' for key 'PRIMARY'"
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 12:00:20",
    "MESSAGE": "database: Can't execute query. Retrying",
    "ICINGADB_ERROR": "can't perform \"INSERT INTO \\\"notification_customvar\\\"...: Error 1062 (23000): Duplicate entry '~\\x1E\\xEB\\x8B\\x8D\\xC2g\\x07\\xE4\\x8A\\xA0^\\x13\\x14\\xE0l\\xD0\\x0F\\...' for key 'PRIMARY'"
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 12:00:20",
    "MESSAGE": "database: Can't execute query. Retrying",
    "ICINGADB_ERROR": "can't perform \"INSERT INTO \\\"state_history\\\"...: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction"
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 12:00:20",
    "MESSAGE": "database: Query retried successfully after error",
    "ICINGADB_AFTER": "833.661486ms",
    "ICINGADB_RECOVERED_ERROR": "can't perform \"INSERT INTO \\\"state_history\\\"...: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction",
},
{
    "_SOURCE_REALTIME_TIMESTAMP": "2024-07-25 12:04:22",
    "MESSAGE": "Error 1062 (23000): Duplicate entry '\\xE9\\xF6\\xC3\\x12\\xA1\\xBE\\xBCo8`h\\xBCA3\\xD72rED\\xE0' for key 'PRIMARY'\ncan't perform \"INSERT INTO \\\"host_customvar\\\" (\\\"host_id\\\", \\\"customvar_id\\\", \\\"environment_id\\\", \\\"id\\\") VALUES (:host_id, :customvar_id, :environment_id, :id)\"\ngithub.com/icinga/icingadb/internal.CantPerformQuery\n\tgithub.com/icinga/icingadb/internal/internal.go:30\ngithub.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2.1\n\tgithub.com/icinga/icingadb/pkg/icingadb/db.go:412\ngithub.com/icinga/icingadb/pkg/retry.WithBackoff\n\tgithub.com/icinga/icingadb/pkg/retry/retry.go:60\ngithub.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2\n\tgithub.com/icinga/icingadb/pkg/icingadb/db.go:407\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\tgolang.org/x/sync@v0.7.0/errgroup/errgroup.go:78\nruntime.goexit\n\truntime/asm_amd64.s:1695\nretry deadline exceeded\ngithub.com/icinga/icingadb/pkg/retry.WithBackoff\n\tgithub.com/icinga/icingadb/pkg/retry/retry.go:95\ngithub.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2\n\tgithub.com/icinga/icingadb/pkg/icingadb/db.go:407\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\tgolang.org/x/sync@v0.7.0/errgroup/errgroup.go:78\nruntime.goexit\n\truntime/asm_amd64.s:1695",
}
```

</details> 

ref/IP/48748

## Additional context

```bash
Error 1062 (23000): Duplicate entry '\\xE9\\xF6\\xC3\\x12\\xA1\\xBE\\xBCo8`h\\xBCA3\\xD72rED\\xE0' for key 'PRIMARY'
can't perform \"INSERT INTO \"host_customvar\" (\"host_id\", \"customvar_id\", \"environment_id\", \"id\") VALUES (:host_id, :customvar_id, :environment_id, :id)\"
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2.1
    github.com/icinga/icingadb/pkg/icingadb/db.go:412
    github.com/icinga/icingadb/pkg/retry/retry.go:60
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2
    github.com/icinga/icingadb/pkg/icingadb/db.go:407
    golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78
runtime.goexit
    runtime/asm_amd64.s:1695
retry deadline exceeded
    github.com/icinga/icingadb/pkg/retry/retry.go:95
github.com/icinga/icingadb/pkg/icingadb.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2
    github.com/icinga/icingadb/pkg/icingadb/db.go:407
    golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78
runtime.goexit
    runtime/asm_amd64.s:1695
```

[^1]:  It's a well-known fact that the Icinga 2 heartbeats arrives in Redis every 1 second, but we only read them every `3` seconds and these 3 seconds are the pacesetter for HA realize thing.
[^2]: The overall Redis heartbeat timeout is set to `1m`, i.e. if we do not receive an Icinga 2 heartbeat for more than one minute, we hand over the HA responsibility. Seemingly, that very same timeout is also used as a deadline for the HA transaction retry's.

oxzi:
- **icingadb: Unify select cases for derived contexts**
  The main loop select cases for hactx.Done() and ctx.Done() were unified, as hactx is a derived ctx. A closed ctx case may be lost as the hactx case could have been chosen.
- **HA: Increase log level for heartbeats from the future**
  Timing issues may be the root of future failures. Thus, it is important to be aware if the timing seems to be out of sync.
- **HA: Deferred SQL Transaction Rollback**
  Each transaction is created within the retryable function, but this function may be exited prematurely before committing. A deferred rollback ensures that the transaction will be rolled back and cleaned up in this case, or will be a noop when performed after the commit.
- **HA: Insert environment within retryable function**
  The HA.insertEnvironment() method was inlined into the retryable function to use the deadlined context. Otherwise, this might block afterwards, as it was used within HA.realize(), but without the passed context.
- **HA/Heartbeat: Use last message's timestamp**
  Since the retryable HA function may be executed a few times before succeeding, the inserted heartbeat value will be directly outdated. The heartbeat logic was slightly altered to always use the latest heartbeat time value.
- **HA: Abort Transaction Commit after Timeout**
  A strange HA behavior was reported in #787, resulting in both instances being active.
  The logs contained an entry of the previous active instance exiting the HA.realize() method successfully after 1m9s. This, however, should not be possible as the method's context is deadlined to a minute after the heartbeat was received.
  However, as it turns out, executing COMMIT on a database transaction is not bound to the transaction's context, allowing to survive longer. To mitigate this, another context watch was introduced. Doing so allows directly handing over, while the other instance can now take over due to the expired heartbeat in the database.

Closes #787.

### DIFF
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -221,7 +221,7 @@ func (h *HA) controller() {
 
 				// Ensure that updating/inserting the instance row is completed by the current heartbeat's expiry time.
 				realizeCtx, cancelRealizeCtx := context.WithDeadline(h.ctx, m.ExpiryTime())
-				err = h.realize(realizeCtx, s, t, envId, shouldLogRoutineEvents)
+				err = h.realize(realizeCtx, s, envId, shouldLogRoutineEvents)
 				cancelRealizeCtx()
 				if errors.Is(err, context.DeadlineExceeded) {
 					h.signalHandover("instance update/insert deadline exceeded heartbeat expiry time")
@@ -268,10 +268,12 @@ func (h *HA) controller() {
 // realize a HA cycle triggered by a heartbeat event.
 //
 // shouldLogRoutineEvents indicates if recurrent events should be logged.
+//
+// The internal, retryable function always fetches the last received heartbeat's timestamp instead of reusing the one
+// from the calling controller loop. Doing so results in inserting a more accurate timestamp if a retry happens.
 func (h *HA) realize(
 	ctx context.Context,
 	s *icingaredisv1.IcingaStatus,
-	t *types.UnixMilli,
 	envId types.Binary,
 	shouldLogRoutineEvents bool,
 ) error {
@@ -354,7 +356,7 @@ func (h *HA) realize(
 				EnvironmentMeta: v1.EnvironmentMeta{
 					EnvironmentId: envId,
 				},
-				Heartbeat:                         *t,
+				Heartbeat:                         types.UnixMilli(time.UnixMilli(h.heartbeat.LastMessageTime())),
 				Responsible:                       types.Bool{Bool: takeover != "" || h.responsible, Valid: true},
 				EndpointId:                        s.EndpointId,
 				Icinga2Version:                    s.Version,
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -303,6 +303,7 @@ func (h *HA) realize(
 			if errBegin != nil {
 				return errors.Wrap(errBegin, "can't start transaction")
 			}
+			defer func() { _ = tx.Rollback() }()
 
 			query := h.db.Rebind("SELECT id, heartbeat FROM icingadb_instance "+
 				"WHERE environment_id = ? AND responsible = ? AND id <> ?") + selectLock
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -374,9 +374,14 @@ func (h *HA) realize(
 
 			if takeover != "" {
 				stmt := h.db.Rebind("UPDATE icingadb_instance SET responsible = ? WHERE environment_id = ? AND id <> ?")
-				_, err := tx.ExecContext(ctx, stmt, "n", envId, h.instanceId)
+				if _, err := tx.ExecContext(ctx, stmt, "n", envId, h.instanceId); err != nil {
+					return database.CantPerformQuery(err, stmt)
+				}
 
-				if err != nil {
+				// Insert the environment after each heartbeat takeover if it does not already exist in the database
+				// as the environment may have changed, although this is likely to happen very rarely.
+				stmt, _ = h.db.BuildInsertIgnoreStmt(h.environment)
+				if _, err := h.db.NamedExecContext(ctx, stmt, h.environment); err != nil {
 					return database.CantPerformQuery(err, stmt)
 				}
 			}
@@ -424,12 +429,6 @@ func (h *HA) realize(
 	}
 
 	if takeover != "" {
-		// Insert the environment after each heartbeat takeover if it does not already exist in the database
-		// as the environment may have changed, although this is likely to happen very rarely.
-		if err := h.insertEnvironment(); err != nil {
-			return errors.Wrap(err, "can't insert environment")
-		}
-
 		h.signalTakeover(takeover)
 	} else if otherResponsible {
 		if state := h.state.Load(); !state.otherResponsible {
@@ -450,18 +449,6 @@ func (h *HA) realizeLostHeartbeat() {
 	if _, err := h.db.ExecContext(h.ctx, stmt, "n", h.instanceId); err != nil && !utils.IsContextCanceled(err) {
 		h.logger.Warnw("Can't update instance", zap.Error(database.CantPerformQuery(err, stmt)))
 	}
-}
-
-// insertEnvironment inserts the environment from the specified state into the database if it does not already exist.
-func (h *HA) insertEnvironment() error {
-	// Instead of checking whether the environment already exists, use an INSERT statement that does nothing if it does.
-	stmt, _ := h.db.BuildInsertIgnoreStmt(h.environment)
-
-	if _, err := h.db.NamedExecContext(h.ctx, stmt, h.environment); err != nil {
-		return database.CantPerformQuery(err, stmt)
-	}
-
-	return nil
 }
 
 func (h *HA) removeInstance(ctx context.Context) {
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -173,7 +173,7 @@ func (h *HA) controller() {
 				}
 				tt := t.Time()
 				if tt.After(now.Add(1 * time.Second)) {
-					h.logger.Debugw("Received heartbeat from the future", zap.Time("time", tt))
+					h.logger.Warnw("Received heartbeat from the future", zap.Time("time", tt))
 				}
 				if tt.Before(now.Add(-1 * peerTimeout)) {
 					h.logger.Errorw("Received heartbeat from the past", zap.Time("time", tt))
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -267,6 +267,9 @@ func (h *HA) controller() {
 
 // realize a HA cycle triggered by a heartbeat event.
 //
+// The context passed is expected to have a deadline, otherwise the method will panic. This deadline is strictly
+// enforced to abort the realization logic the moment the context expires.
+//
 // shouldLogRoutineEvents indicates if recurrent events should be logged.
 //
 // The internal, retryable function always fetches the last received heartbeat's timestamp instead of reusing the one
@@ -388,8 +391,39 @@ func (h *HA) realize(
 				}
 			}
 
-			if err := tx.Commit(); err != nil {
-				return errors.Wrap(err, "can't commit transaction")
+			// In general, cancellation does not work for COMMIT and ROLLBACK. Some database drivers may support a
+			// context-based abort, but only if the DBMS allows it. This was also discussed in the initial issue about
+			// context support to Go's sql package: https://github.com/golang/go/issues/15123#issuecomment-245882486
+			//
+			// This paragraph is implementation knowledge, not covered by the API specification. Go's sql.Tx.Commit() -
+			// which is not being overridden by sqlx.Tx - performs a preflight check on the context before handing over
+			// to the driver's Commit() method. Drivers may behave differently. For example, the used
+			// github.com/go-sql-driver/mysql package calls its internal exec() method with a COMMIT query, writing and
+			// reading packets without honoring the context.
+			//
+			// In a nutshell, one cannot expect a Tx.Commit() call to be covered by the transaction context. For this
+			// reason, the following Commit() call has been moved to its own goroutine, which communicates back via a
+			// channel selected along with the context. If the context ends before Commit(), this retryable function
+			// returns with a non-retryable error.
+			//
+			// However, while the COMMIT continues in the background, it may still succeed. In this case, the state of
+			// the database does not match the state of Icinga DB, specifically the database says that this instance is
+			// active while this instance thinks otherwise. Fortunately, this mismatch is not critical because when this
+			// function is re-entered, the initial SELECT query would be empty for this Icinga DB node and imply the
+			// presence of another active instance for the other node. Effectively, this could result in a single HA
+			// cycle with no active node. Afterwards, either this instance takes over due to the false impression that
+			// no other node is active, or the other instances does so as the inserted heartbeat has already expired.
+			// Not great, not terrible.
+			commitErrCh := make(chan error, 1)
+			go func() { commitErrCh <- tx.Commit() }()
+
+			select {
+			case err := <-commitErrCh:
+				if err != nil {
+					return errors.Wrap(err, "can't commit transaction")
+				}
+			case <-ctx.Done():
+				return ctx.Err()
 			}
 
 			return nil
--- a/pkg/icingaredis/heartbeat.go
+++ b/pkg/icingaredis/heartbeat.go
@@ -25,6 +25,7 @@ type Heartbeat struct {
 	active         bool
 	events         chan *HeartbeatMessage
 	lastReceivedMs atomic.Int64
+	lastMessageMs  atomic.Int64
 	cancelCtx      context.CancelFunc
 	client         *redis.Client
 	done           chan struct{}
@@ -60,6 +61,11 @@ func (h *Heartbeat) Events() <-chan *HeartbeatMessage {
 // LastReceived returns the last heartbeat's receive time in ms.
 func (h *Heartbeat) LastReceived() int64 {
 	return h.lastReceivedMs.Load()
+}
+
+// LastMessageTime returns the last message's time in ms.
+func (h *Heartbeat) LastMessageTime() int64 {
+	return h.lastMessageMs.Load()
 }
 
 // Close stops the heartbeat controller loop, waits for it to finish, and returns an error if any.
@@ -139,6 +145,15 @@ func (h *Heartbeat) controller(ctx context.Context) {
 				}
 
 				h.lastReceivedMs.Store(m.received.UnixMilli())
+
+				statsT, err := m.stats.Time()
+				if err != nil {
+					h.logger.Warnw("Received Icinga heartbeat with invalid stats time", zap.Error(err))
+					h.lastMessageMs.Store(0)
+				} else {
+					h.lastMessageMs.Store(statsT.Time().UnixMilli())
+				}
+
 				h.sendEvent(m)
 			case <-time.After(Timeout):
 				if h.active {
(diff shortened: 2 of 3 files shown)


pkg/icingadb/ha.go
```diff
@@ -264,6 +264,9 @@ func (h *HA) controller() {
 
 // realize a HA cycle triggered by a heartbeat event.
 //
+// The given context MUST have a deadline, as the method will otherwise panic. This deadline is hardly enforced to
+// cancel the logic if it has expired.
```
julianbrost, line 268: Hint: look up the meaning of "hardly", this sounds like it's not what you'd expect :sweat_smile:

oxzi: A hard working person or a hardly working person. There was a difference, I remember. Oops.

oxzi:
After going over this with @lippserd, I reworked this PR. The changes are, in a nutshell, that the potentially blocking `COMMIT` query will be guarded with a `select` watching at the context and that always the latest heartbeat timestamp will be used, as otherwise an already heartbeat will be retried. The description in the initial post was updated accordingly.

pkg/icingadb/ha.go
```diff
@@ -370,23 +376,48 @@ func (h *HA) realize(
...
+				return ctx.Err()
 			}
 
 			return nil
 		},
 		retry.Retryable,
 		backoff.NewExponentialWithJitter(256*time.Millisecond, 3*time.Second),
 		retry.Settings{
-			// Intentionally no timeout is set, as we use a context with a deadline.
+			// Intentionally, no timeout is set because a context with a deadline is used and QuickContextExit is set.
```
lippserd, line 420: QuickContextExit is no longer there ;).

pkg/icingadb/ha.go
```diff
@@ -370,23 +376,48 @@ func (h *HA) realize(
...
+			// context support to sql: https://github.com/golang/go/issues/15123#issuecomment-245882486
+			//
+			// The following is implementation knowledge, not covered by the API specification. Go's sql.Tx.Commit() -
+			// which is not being overridden by sqlx.Tx - performs a preflight check on the context before handing over
+			// to the driver's Commit method. The drivers may behave differently. For example, the used
+			// github.com/go-sql-driver/mysql package calls its internal exec method with a "COMMIT" query, writing and
+			// reading packets without honoring the context.
+			//
+			// In a nutshell, one cannot expect a Commit() call to be covered by the transaction context. Thus, an
+			// additional check wraps the call, ensuring that this function will be left when the context is done.
```
lippserd, line 402: Can you please also add a comment that it is theoretically no problem to call realize again while the COMMIT is still running in the go routine.

lippserd (changes requested):
Please rebase and add a comment about `COMMIT` possibly not respecting context deadlines in `ExecTx()` and `NamedExecTx()`.

pkg/icingadb/ha.go
oxzi: I wrote more prose. Please take another look.

lippserd (approved):

pkg/icingadb/ha.go
lippserd: Looks good, ty! 👍

pkg/icingaredis/heartbeat.go
```diff
@@ -25,6 +25,7 @@ type Heartbeat struct {
 	active         bool
 	events         chan *HeartbeatMessage
 	lastReceivedMs int64
+	lastMessageMs  int64
```
yhabteab, line 28: Am I missing something? Why can't you make this of type `atomic.Int64` and use its `Load()` and `Store()` methods?

pkg/icingaredis/heartbeat.go
```diff
@@ -62,6 +63,11 @@ func (h *Heartbeat) LastReceived() int64 {
 	return atomic.LoadInt64(&h.lastReceivedMs)
 }
 
+// LastMessage returns the last message's time in ms.
+func (h *Heartbeat) LastMessage() int64 {
```
yhabteab, line 67: Since this does not return the actual heartbeat message, it should be named according to what it actually returns.

pkg/icingaredis/heartbeat.go
oxzi: It was implemented this way to be symmetrical to `lastReceivedMs`, especially in the later calls.

yhabteab: The previous implementation was designed that way because `atomic.Int64` was not available prior to [Go 1.19](https://go.dev/doc/go1.19#atomic_types).

pkg/icingaredis/heartbeat.go
oxzi: This method name tries to follow `LastReceived`, returning the last received timestamp while `LastMessage` returns the last message's timestamp. Thus, what would you suggest then? Especially when fitting with its twin method.

pkg/icingaredis/heartbeat.go
oxzi: Okay, but it works, right? There is no necessity to change it, which would blow up this change, btw. And having these two implemented differently would just look odd.

yhabteab: > There is no necessity to change it, which would blow up this change, btw. And having these two implemented differently would just look odd.

I was referring to the new type you introduced and did not mean for you to change the existing code. Personally, we should be more focused on using new features rather than maintaining symmetry with the existing code :), anyway I've changed the existing usages with #829.

pkg/icingaredis/heartbeat.go
yhabteab: > while `LastMessage` returns the last message's timestamp. Thus, what would you suggest then?

Please don't take this the wrong way, and I'm not trying to impose my own ideals on you. I just think that `LastMessage` is not the same as `last message timestamp`, and I apologize for not providing an alternative name from the beginning that I'm thinking of. Thus, my first suggestion would be to simply call it `LastMessageTime[stamp]`.

pkg/icingaredis/heartbeat.go
oxzi: Personally, I don't think always using the latest features is something to strive for, unless it adds real value. In that case, the API is more foolproof. I am for changing it, as you did, but then it should be done consequentially. I will update this PR as well.

yhabteab (commented):
Apart from the conflict, everything looks fine to me now. Sorry for that :)

lippserd (changes requested):
Although I approved earlier, I now have something that I had previously forgotten. It's not about the code itself, but about the commit(s). The changes here are subject to multiple commits:

* Changes in main
* Increasing the "heartbeat from future" log to warning
* Explicit rollback
* Inserting the environment inside the transaction
* Introduction and use of LastMessageTime() (plus the corresponding godoc paragraph of realize)
* Executing commit in a separate Go routine (plus its godoc paragraph of realize)

lippserd (approved):
Perfect, thanks a lot @oxzi!

yhabteab (approved):

yhabteab:
Shouldn't this pull request close issue #787?

oxzi:
> Shouldn't this pull request close issue #787?

This was the initial motivation, yes. However, as I was unable to reproduce the reported behavior, I cannot verify this. My intention was to make the observed bug impossible with this change, but maybe there is still another bug lurking around the corner.

However, with a bit more trust in this change, I have linked this PR to the issue. Otherwise, we can reopen it.

PR #801, opened 2024-09-05
Shorten icingadb/v1.AddressBin#Value()

Al2Klimov:


oxzi (approved):
I like the symmetry with `Address6Bin.Value`.

PR #804, opened 2024-09-06
Test icingadb.EntitiesById

Al2Klimov:


pkg/icingadb/entitiesbyid_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+			select {
+			case v, ok := <-ch:
+				require.False(t, ok, "channel should be closed, got %#v", v)
+			case <-time.After(time.Second):
+				require.Fail(t, "channel should not block")
+			}
+		})
+	}
+}
```
oxzi, line 132: There is one case not tested in `EntitiesById.Entities`:
https://github.com/Icinga/icingadb/blob/7c068d4adf169181b4d68dcd6222a18b6990b583/pkg/icingadb/entitiesbyid.go#L40-L41

Thus, please add another test case with a prematurely closed context.

oxzi (changes requested):
Please rebase to please the CI.

oxzi (approved):

PR #805, opened 2024-09-06
Shorten CommentType#UnmarshalText()

Al2Klimov:
by just calling CommentType#UnmarshalJSON().    The implementations behave identically. (Just like for AcknowledgementState,    NotificationStates, NotificationTypes and StateType.)

### DIFF
--- a/pkg/icingadb/types/comment_type.go
+++ b/pkg/icingadb/types/comment_type.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"github.com/icinga/icinga-go-library/types"
 	"github.com/pkg/errors"
-	"strconv"
 )
 
 // CommentType specifies a comment's origin's kind.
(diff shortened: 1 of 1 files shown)


pkg/icingadb/types/comment_type.go
```diff
@@ -30,25 +29,7 @@ func (ct *CommentType) UnmarshalJSON(data []byte) error {
...
-	i, err := strconv.ParseUint(s, 10, 64)
-	if err != nil {
-		return types.CantParseUint64(err, s)
-	}
-
-	c := CommentType(i)
-	if uint64(c) != i {
-		// Truncated due to above cast, obviously too high
-		return badCommentType(s)
-	}
```
oxzi, line 44: This check is missing in `CommentType.UnmarshalJSON`.

oxzi (commented):
While the comment message is technically correct, it does not explain anything. Please just add at least one sentence stating that both implementations were _almost_ identical. Otherwise, the first look at this diff leaves one puzzled.

pkg/icingadb/types/comment_type.go
Al2Klimov: But that method already uses uint8 which implies this check.

oxzi: Fair enough.

oxzi (approved):

lippserd:
@Al2Klimov Please rebase.

PR #809, opened 2024-09-09
GHA: Remove -gcflags="-m" from go build

oxzi:
According to "go tool compile", the "-m" flag results in "print[ing] optimization decisions". Those are shown on GitHub PRs as errors[^1].

In my opinion, this behavior is misleading and has no real benefit.

[^1]: For example: https://github.com/Icinga/icingadb/pull/808/files#diff-a15436f81f31f0b364e028b51108a6f988c775aa76390aebfb857d9b29771187


lippserd (approved):

lippserd:
I originally added this so that with the escape analysis we could point out possible resource leaks, but to be honest, I never looked at it closely either, so we can just remove the noise.

PR #812, opened 2024-09-18
Use `sync/atomic#Pointer` instead of our own wrapper

lippserd:
Go 1.19 introduced `sync/atomic#Pointer` among other things, so we no longer need to use the Atomic wrapper from our Icinga Go library.

### DIFF
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -7,7 +7,6 @@ import (
 	"encoding/hex"
 	"github.com/google/uuid"
 	"github.com/icinga/icinga-go-library/backoff"
-	"github.com/icinga/icinga-go-library/com"
 	"github.com/icinga/icinga-go-library/database"
 	"github.com/icinga/icinga-go-library/logging"
 	"github.com/icinga/icinga-go-library/retry"
@@ -19,6 +18,7 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -35,7 +35,7 @@ type haState struct {
 
 // HA provides high availability and indicates whether a Takeover or Handover must be made.
 type HA struct {
-	state         com.Atomic[haState]
+	state         atomic.Pointer[haState]
 	ctx           context.Context
 	cancelCtx     context.CancelFunc
 	instanceId    types.Binary
@@ -70,6 +70,8 @@ func NewHA(ctx context.Context, db *database.DB, heartbeat *icingaredis.Heartbea
 		takeover:   make(chan string),
 		done:       make(chan struct{}),
 	}
+
+	ha.state.Store(&haState{})
 
 	go ha.controller()
 
@@ -121,7 +123,8 @@ func (h *HA) Takeover() chan string {
 
 // State returns the status quo.
 func (h *HA) State() (responsibleTsMilli int64, responsible, otherResponsible bool) {
-	state, _ := h.state.Load()
+	state := h.state.Load()
+
 	return state.responsibleTsMilli, state.responsible, state.otherResponsible
 }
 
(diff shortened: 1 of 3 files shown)


pkg/icingadb/ha.go
```diff
@@ -514,7 +515,7 @@ func (h *HA) signalHandover(reason string) {
 // signalTakeover claims HA.responsible and notifies the HA.Takeover chan.
 func (h *HA) signalTakeover(reason string) {
 	if !h.responsible {
-		h.state.Store(haState{
+		h.state.Store(&haState{
```
Al2Klimov, line 518: This is fine.

pkg/icingadb/ha.go
```diff
@@ -121,7 +121,8 @@ func (h *HA) Takeover() chan string {
 
 // State returns the status quo.
 func (h *HA) State() (responsibleTsMilli int64, responsible, otherResponsible bool) {
-	state, _ := h.state.Load()
+	state := h.state.Load()
+
```
Al2Klimov, line 125: This has been a zero value if uninitialized, now it's a nil ptr in this case. Now we have to make sure HA#State() don't get a nil ptr.

pkg/icingadb/ha.go
```diff
@@ -428,7 +429,7 @@ func (h *HA) realize(
 
 		h.signalTakeover(takeover)
 	} else if otherResponsible {
-		if state, _ := h.state.Load(); !state.otherResponsible {
+		if state := h.state.Load(); !state.otherResponsible {
```
Al2Klimov, line 432: As we're now loading by-ref, `state.otherResponsible = true` below affects all readers. But this shouldn't be too bad as that's just a bool and we're storing it anyway right away.

pkg/icingadb/ha.go
```diff
@@ -428,9 +433,17 @@ func (h *HA) realize(
...
-		if state, _ := h.state.Load(); !state.otherResponsible {
+		var state haState
+		statePtr := h.state.Load()
+		if statePtr != nil {
+			// Dereference pointer to create a copy of the value it points to.
+			// Ensures that any modifications do not directly affect the original data unless explicitly stored back.
+			state = *statePtr
+		}
+
+		if !state.otherResponsible {
```
yhabteab, line 444: You don't change the state if you don't enter this `if` branch anyway, so I would handle this the same way as before.
```suggestion
		if state := h.state.Load(); state != nil && !state.otherResponsible {
			// Dereference pointer to create a copy of the value it points to.
			// Ensures that any modifications do not directly affect the original data unless explicitly stored back.
			newState := *state
			newState.otherResponsible = true
			h.state.Store(&newState)
		}
```

yhabteab: Oh, never mind, I get it now why you did it that way 🙈

yhabteab (approved):

pkg/icingadb/ha.go
lippserd: Done.

pkg/icingadb/ha.go
lippserd: Fixed.

Al2Klimov (commented):
There's still

https://github.com/Icinga/icingadb/blob/34844764fc949e3b4b6607c700d052da3d9e7203/pkg/icingaredis/telemetry/heartbeat.go#L84

btw.

lippserd:
@Al2Klimov @yhabteab I also changed the type in `telemetry` and adjusted the implementation so that the `atomic.Pointer` usages initially hold the zero values of their types. In addition, copies of the pointer values are now only created if they are actually needed.

yhabteab (approved):

lippserd:
@yhabteab I'm sorry,  I had to push another change 😆.

Al2Klimov (approved):

yhabteab (approved):

PR #813, opened 2024-09-18
Update GHA dependencies via dependabot

yhabteab:
Dependabot supports to automatically refresh GitHub actions as well, so we should use it to avoid such errors:
https://github.com/Icinga/icingadb/actions/runs/10903916830/job/30259089610

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,3 +12,8 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: daily


.github/dependabot.yml
```diff
@@ -12,3 +12,9 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
```
lippserd, line 20: Why the PR limit? (I know it was already there still I'm wondering why it's necessary).

Al2Klimov (approved):

yhabteab:
You have to force merge this PR though, so that it can start upgrading the dependencies from within the main branch.

.github/dependabot.yml
yhabteab: I guess to limit the PRs it can create at a time, but I can also remove it, I just wanted it to be identical to the other config.

Al2Klimov: > By default, Dependabot opens a maximum of five pull requests for version updates.

– https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

Al2Klimov (approved):

.github/dependabot.yml
yhabteab: Then 5 PRs a day is more than enough for the GitHub Actions, I would say.

Al2Klimov: > 5 PRs a day

This is not a _rate_ limit:

> Once there are five open pull requests from Dependabot, Dependabot will not open any new requests until some of those open requests are merged or closed.

– https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

yhabteab: Nevertheless, the default limit is more than enough for the GA.

PR #818, opened 2024-09-20
Clarify that our Debian installation docs also apply to the Raspberry Pi OS

Al2Klimov:
* CC https://github.com/Icinga/packages.icinga.com/pull/46

### DIFF
--- a/doc/02-Installation.md
+++ b/doc/02-Installation.md
@@ -75,7 +75,7 @@ wget https://packages.icinga.com/centos/ICINGA-release.repo -O /etc/yum.repos.d/
 <!-- {% if debian %} -->
 <!-- {% if not icingaDocs %} -->
 
-### Debian Repository
+### Debian and Raspberry Pi OS Repository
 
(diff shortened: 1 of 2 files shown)


julianbrost:
Note that the Icinga DB installation instructions are given for fewer operating systems than for example Icinga 2, compare the following two:

* https://github.com/Icinga/icingadb/tree/main/doc/02-Installation.md.d
* https://github.com/Icinga/icinga2/tree/master/doc/02-installation.md.d

You'll note that Fedora is also missing for example. This is because on https://icinga.com/subscriptions/support-matrix/, we make a difference between support for agents vs. for a full master setup. So I'm not sure if makes much sense to add instructions for Raspberry Pi OS in that context.

lippserd:
Icinga Web also lists other instructions, but ultimately all projects should include the instructions from here:

https://github.com/Icinga/package-installation-docs/tree/main

This is different from the support matrix because that is only for support, not for operating systems for which we have packages.

PR #819, opened 2024-09-25
pkg/icingaredis/: check context termination while looping over channel

Al2Klimov:
A `for x := range c` will block as long as c blocks, ignoring `<-ctx.Done()`. The latter has to be checked via `select` instead, together with c.

* Discovered during a discussion about the unit tests: https://github.com/Icinga/icingadb/pull/799#issuecomment-2355907794

Also, _hide whitespace_ under _files changed_ is your friend. 😅


oxzi (approved):
Basst scho.

PR #821, opened 2024-10-04
Introduce golangci-lint and fix findings

oxzi:
First, the same GitHub Action used in the icinga-go-library was added to the Go workflow.

Then, the findings were addressed. In general, these were mostly missing return check - when being unused anyway -, potential overflows for integer type conversions, and SHA1 warnings.

If applicable, the code was slightly modified to perform the necessary checks. Sometimes the linter has detected the checks. Sometimes not and a linter comment was added.

### DIFF
--- a/cmd/icingadb-migrate/convert.go
+++ b/cmd/icingadb-migrate/convert.go
@@ -11,6 +11,7 @@ import (
 	"github.com/icinga/icingadb/pkg/icingadb/v1/history"
 	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
+	"math"
 	"strconv"
 	"strings"
 	"time"
@@ -282,6 +283,14 @@ func convertDowntimeRows(
 			cancelTime = actualEnd
 		}
 
+		// The IDO duration is of type bigint, representing seconds, while Icinga DB's FlexibleDuration is an unsigned
+		// bigint, representing milliseconds. In theory, there should be no negative value in the IDO and multiplied
+		// with factor 1,000 should not overflow anything. In theory, at least. To make sure, invalid values are capped.
+		var flexibleDuration uint64
+		if durationSec := row.Duration; durationSec >= 0 && durationSec < math.MaxUint64/1_000 {
+			flexibleDuration = uint64(durationSec) * 1_000
+		}
+
 		downtimeHistory = append(downtimeHistory, &history.DowntimeHistory{
 			DowntimeHistoryEntity: history.DowntimeHistoryEntity{DowntimeId: id},
 			HistoryTableMeta: history.HistoryTableMeta{
@@ -299,7 +308,7 @@ func convertDowntimeRows(
 			Author:             row.AuthorName,
 			Comment:            row.CommentData,
 			IsFlexible:         types.Bool{Bool: row.IsFixed == 0, Valid: true},
-			FlexibleDuration:   uint64(row.Duration) * 1000,
+			FlexibleDuration:   flexibleDuration,
 			ScheduledStartTime: scheduledStart,
 			ScheduledEndTime:   scheduledEnd,
 			StartTime:          startTime,
--- a/pkg/icingadb/cleanup.go
+++ b/pkg/icingadb/cleanup.go
@@ -32,7 +32,7 @@ func (stmt *CleanupStmt) CleanupOlderThan(
 	defer db.Log(ctx, q, &counter).Stop()
 
 	for {
-		var rowsDeleted int64
+		var rowsDeleted uint64
 
 		err := retry.WithBackoff(
 			ctx,
@@ -45,7 +45,10 @@ func (stmt *CleanupStmt) CleanupOlderThan(
 					return database.CantPerformQuery(err, q)
 				}
 
-				rowsDeleted, err = rs.RowsAffected()
+				i, err := rs.RowsAffected()
+				if err == nil && i >= 0 {
+					rowsDeleted = uint64(i)
+				}
 
 				return err
 			},
@@ -57,7 +60,7 @@ func (stmt *CleanupStmt) CleanupOlderThan(
 			return 0, err
 		}
 
-		counter.Add(uint64(rowsDeleted))
+		counter.Add(rowsDeleted)
 
 		for _, onSuccess := range onSuccess {
 			if err := onSuccess(ctx, make([]struct{}, rowsDeleted)); err != nil {
@@ -65,7 +68,7 @@ func (stmt *CleanupStmt) CleanupOlderThan(
 			}
 		}
 
-		if rowsDeleted < int64(count) {
+		if rowsDeleted < count {
 			break
 		}
 	}
--- a/pkg/icingadb/types/notification_type.go
+++ b/pkg/icingadb/types/notification_type.go
@@ -3,7 +3,6 @@ package types
 import (
 	"database/sql/driver"
 	"encoding"
-	"github.com/icinga/icinga-go-library/types"
 	"github.com/pkg/errors"
 	"strconv"
 )
@@ -15,17 +14,12 @@ type NotificationType uint16
 func (nt *NotificationType) UnmarshalText(text []byte) error {
 	s := string(text)
 
-	i, err := strconv.ParseUint(s, 10, 64)
+	i, err := strconv.ParseUint(s, 10, 16)
 	if err != nil {
-		return types.CantParseUint64(err, s)
+		return errors.Wrapf(err, "can't parse %q into uint16", s)
 	}
 
 	n := NotificationType(i)
-	if uint64(n) != i {
-		// Truncated due to above cast, obviously too high
-		return badNotificationType(s)
-	}
-
 	if _, ok := notificationTypes[n]; !ok {
 		return badNotificationType(s)
 	}
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -38,6 +38,14 @@ jobs:
         with:
           install-go: false
 
+      - uses: golangci/golangci-lint-action@v6
+        with:
+          version: latest
+          only-new-issues: true
+
+          # Enable the gosec linter w/o having to create a .golangci.yml config
+          args: -E gosec
+
   vet:
     runs-on: ubuntu-latest
     steps:
--- a/cmd/icingadb-migrate/main.go
+++ b/cmd/icingadb-migrate/main.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"context"
-	"crypto/sha1"
+	"crypto/sha1" // #nosec G505 -- used as a non-cryptographic hash function to hash IDs
 	"database/sql"
 	_ "embed"
 	"encoding/hex"
--- a/cmd/icingadb-migrate/misc.go
+++ b/cmd/icingadb-migrate/misc.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"context"
-	"crypto/sha1"
+	"crypto/sha1" // #nosec G505 -- used as a non-cryptographic hash function to hash IDs
 	"github.com/icinga/icinga-go-library/database"
 	"github.com/icinga/icinga-go-library/objectpacker"
 	"github.com/icinga/icinga-go-library/types"
@@ -54,7 +54,7 @@ var objectTypes = map[uint8]string{1: "host", 2: "service"}
 
 // hashAny combines objectpacker.PackAny and SHA1 hashing.
 func hashAny(in interface{}) []byte {
-	hash := sha1.New()
+	hash := sha1.New() // #nosec G401 -- used as a non-cryptographic hash function to hash IDs
 	if err := objectpacker.PackAny(in, hash); err != nil {
 		panic(err)
 	}
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -50,7 +50,7 @@ func run() int {
 	_ = sdnotify.Ready()
 
 	logger := logs.GetLogger()
-	defer logger.Sync()
+	defer func() { _ = logger.Sync() }()
 
 	logger.Infof("Starting Icinga DB daemon (%s)", internal.Version.Version)
 
@@ -58,7 +58,7 @@ func run() int {
 	if err != nil {
 		logger.Fatalf("%+v", errors.Wrap(err, "can't create database connection pool from config"))
 	}
-	defer db.Close()
+	defer func() { _ = db.Close() }()
 	{
 		logger.Infof("Connecting to database at '%s'", db.GetAddr())
 		err := db.Ping()
@@ -112,7 +112,7 @@ func run() int {
 		if err != nil {
 			logger.Fatalf("%+v", errors.Wrap(err, "can't create database connection pool from config"))
 		}
-		defer db.Close()
+		defer func() { _ = db.Close() }()
 		ha = icingadb.NewHA(ctx, db, heartbeat, logs.GetChildLogger("high-availability"))
 
 		telemetryLogger := logs.GetChildLogger("telemetry")
@@ -125,7 +125,7 @@ func run() int {
 		// Give up after 3s, not 5m (default) not to hang for 5m if DB is down.
 		ctx, cancelCtx := context.WithTimeout(context.Background(), 3*time.Second)
 
-		ha.Close(ctx)
+		_ = ha.Close(ctx)
 		cancelCtx()
 	}()
 	s := icingadb.NewSync(db, rc, logs.GetChildLogger("config-sync"))
--- a/pkg/icingadb/delta_test.go
+++ b/pkg/icingadb/delta_test.go
@@ -208,14 +208,14 @@ func testDeltaVerifyResult(t *testing.T, name string, expected map[uint64]uint64
 }
 
 func BenchmarkDelta(b *testing.B) {
-	for n := 1 << 10; n <= 1<<20; n <<= 1 {
-		b.Run(strconv.Itoa(n), func(b *testing.B) {
+	for n := uint64(1 << 10); n <= 1<<20; n <<= 1 {
+		b.Run(strconv.FormatUint(n, 10), func(b *testing.B) {
 			benchmarkDelta(b, n)
 		})
 	}
 }
 
-func benchmarkDelta(b *testing.B, numEntities int) {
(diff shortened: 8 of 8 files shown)


pkg/icingadb/history/retention.go
```diff
@@ -180,9 +181,14 @@ func (r *Retention) Start(ctx context.Context) error {
 			zap.Uint64("retention-days", days),
 		)
 
+		if days > -math.MinInt {
```
lippserd, line 184: What if we switch from `uint64` to `uint32`?

lippserd (changes requested):
In addition to the requested changes, I would split the commits, e.g.
* GitHub Action
* Ignore errors explicitly
* Weak SHA1
* Int conversions

pkg/icingadb/types/comment_type.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"github.com/icinga/icinga-go-library/types"
 	"github.com/pkg/errors"
+	"math"
```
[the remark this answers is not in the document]
oxzi: Indeed! Thanks for merging it. This shortens this PR.

pkg/icingadb/history/retention.go
oxzi: The (non) issue is that the size of `int` dependents on your platform, either being 32 or 64 bits. In addition, the min values is always one less in the negative direction than the max value in the positive (-9223372036854775808 and 9223372036854775807 for `int` on my amd64).

Thus, there may be the case that days is within `(-math.MinInt, math.MaxUint64]`, which this [playground example](https://go.dev/play/p/YF8TiQPPZoN) should illustrate.

However, this is not a real issue when being put into context. The value of `days` will be used for the retention and if a user decides to overflow (or is it underflow now?) `math.MinInt`, we are dealing with over `2.526951242973911e+16` years. Thus, this is just a sanity check and may prevent us from potential crashes if a user decides to put a fantasy-grade number there.

oxzi:
@lippserd: Thanks for your comment. I have addressed the changes or wrote rationales above.

The commit was also split, as you suggested, just with the GitHub Action commit being the last, as otherwise - when checking the CI for each commit independently - there would be failures before these were addressed.

pkg/icingadb/history/retention.go
julianbrost: Are the `int` bounds actually the relevant thing here? Like is the `time` package even able to represent the result of `tick.Time.AddDate(0, 0, -int(days))` with `days` being $2^{31}-1$ or even $2^{63}-1$?

oxzi: Not really as `time.Time` may overflow already. 

The following snippet demonstrates this:
```go
	t := time.Now()
	fmt.Printf("%v\n%v\n", t, t.AddDate(0, 0, -math.MaxUint32))
```
Resulting in jumping back one day:
```
2024-10-08 10:32:27.911974323 +0200 CEST m=+0.000032968
2024-10-09 10:32:27.911974323 +0200 CEST
```

This is known behavior, as shown in https://github.com/golang/go/issues/20678 and other issues.

But even adding "just" `math.MaxUint32` does already overflow the time. Thus, I think there is not much we can do here and if a user decides to write astrophysical space time numbers, than let it be.

I moved the check up and decide to bail out of the retention if the user has configured more than $2^{16}$ days. This is totally arbitrary, I know, but may safe a setup from purging new data, as shown in my example above.

julianbrost: > I moved the check up and decide to bail out of the retention if the user has configured more than $2^{16}$ days. This is totally arbitrary, I know, but may safe a setup from purging new data, as shown in my example above.

Wouldn't it make sense to just change the type in the config structs to `uint16` instead? Then it would directly reject the (probably wrong/unintended) config instead of just doing nothing with a warning. Also, that's about 180 years, that sounds like more then enough for regular use and 180 years in the past should be fine for the `time` package, so wouldn't that solve all overflow issues? Maybe this will even work without any linter comments.

julianbrost: > and 180 years in the past should be fine for the `time` package, so wouldn't that solve all overflow issues?

Addendum: Probably not all. The database schema uses unsigned Unix timestamps, so anything before 1970 (which is less than 180 years ago, alternatively this would fix itself in about 126 years :see_no_evil:) would probably still overflow somewhere. `uint8`, i.e. 255 days would be too small, so maybe we want an extra check somewhere after all. Though that could just be a comparison with Unix time 0, if it's before that, just skipping the deletion (that would even result in perfectly correct behavior, there can't be anything earlier than that in the database).

oxzi: Good point. Thus I altered the if-guard to ensure that the value will neither overflow int64 nor result in a negative timestamp when being used in subtraction.

https://github.com/Icinga/icingadb/compare/1cd597b0d8887737edf033f034f366375f35efc1..f3fdc1644a79dfa12402b61de03576afdabab668

lippserd (approved):

pkg/icingadb/types/notification_type.go
```diff
@@ -19,13 +20,11 @@ func (nt *NotificationType) UnmarshalText(text []byte) error {
 	if err != nil {
 		return types.CantParseUint64(err, s)
 	}
-
-	n := NotificationType(i)
-	if uint64(n) != i {
-		// Truncated due to above cast, obviously too high
+	if i > math.MaxUint16 {
```
julianbrost, line 23: One line above the context shown by GitHub: `strconv.ParseUint(s, 10, 64)`

Shouldn't changing that 64 to 16 do an appropriate bounds check? :thinking: (Other question would still be if the linter was also happy with that).

julianbrost (changes requested):

pkg/icingadb/history/retention.go
```diff
@@ -172,6 +173,20 @@ func (r *Retention) Start(ctx context.Context) error {
 			r.logger.Debugf("Skipping history retention for category %s", stmt.Category)
 			continue
 		}
+		// #nosec G115 -- First if-clause checks int overflow
+		if days > math.MaxInt64 || time.Now().AddDate(0, 0, -int(days)).Before(time.Unix(0, 0)) {
```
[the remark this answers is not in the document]
julianbrost: > but after plugging `MaxUint16` into my playbook example, I'm confused as it says 1830 and I'd have expected something like 1845, is it an overflow or some weird calendar thing?

D'oh! I forget that `time.Now()` is always 2009 in the Go playground...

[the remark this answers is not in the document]
lippserd: I wouldn't care too much about 32bit builds but reducing days to `uint16` sounds good to me too. Though I would do that in a separate PR.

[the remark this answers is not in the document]
julianbrost: > I wouldn't care too much about 32bit builds

To fix that part, all the change that's needed would be to use `math.MaxInt` instead of `math.MaxInt64`, I wouldn't say that's too much. But that part will likely become irrelevant anyway due to `time.Time` overflowing.

cmd/icingadb-migrate/convert.go
```diff
@@ -299,7 +299,7 @@ func convertDowntimeRows(
 			Author:             row.AuthorName,
 			Comment:            row.CommentData,
 			IsFlexible:         types.Bool{Bool: row.IsFixed == 0, Valid: true},
-			FlexibleDuration:   uint64(row.Duration) * 1000,
+			FlexibleDuration:   uint64(row.Duration) * 1000, // #nosec G115 -- flexible_duration is unsigned in Icinga DB's schema
```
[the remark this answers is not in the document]
oxzi: In IDO it is a `BIGINT`, in Icinga DB it's an `unsigned int`. I have specified the comment and made ensured that no negative number gets incorrectly casted to an `uint64`.

pkg/icingadb/types/notification_type.go
oxzi: Good idea and it even worked for the linter :exploding_head:

pkg/icingadb/history/retention.go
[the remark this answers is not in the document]
oxzi: Changed for int32 and did some further testing: https://go.dev/play/p/92-1ROGJBOl

pkg/icingadb/cleanup.go
```diff
@@ -45,7 +45,8 @@ func (stmt *CleanupStmt) CleanupOlderThan(
 					return database.CantPerformQuery(err, q)
 				}
 
-				rowsDeleted, err = rs.RowsAffected()
+				i, err := rs.RowsAffected()
+				rowsDeleted = uint64(i) // #nosec G115 -- negative amount of rows should not be possible
```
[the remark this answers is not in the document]
oxzi: Oops. Should be `if err == nil && i > 0` for the success case. I'm glad we have tests :)

https://github.com/Icinga/icingadb/compare/65cdadfa0d9fab8a6c556bb46e75bfdfc6c095f1..07ad635148db1d569d726f3d9254d8eaef0435d2
https://github.com/Icinga/icingadb/actions/runs/11251762048/job/31283488727

cmd/icingadb-migrate/convert.go
```diff
@@ -299,7 +299,7 @@ func convertDowntimeRows(
 			Author:             row.AuthorName,
 			Comment:            row.CommentData,
 			IsFlexible:         types.Bool{Bool: row.IsFixed == 0, Valid: true},
-			FlexibleDuration:   uint64(row.Duration) * 1000,
+			FlexibleDuration:   uint64(max(row.Duration, 0)) * 1000, // #nosec G115 -- IDO duration is a BIGINT; shouldn't be negative
```
julianbrost, line 302: Is that linter warning really due to the cast now or is it actually the `* 1000` which can also overflow?

In general, I'd prefer some extra checks with a "make linter happy" comment over a "disable linter" comment, as that makes more clear what is handled while the other might accidentally silence other problems in the same line.

pkg/icingadb/types/notification_type.go
```diff
@@ -15,17 +15,12 @@ type NotificationType uint16
 func (nt *NotificationType) UnmarshalText(text []byte) error {
 	s := string(text)
 
-	i, err := strconv.ParseUint(s, 10, 64)
+	i, err := strconv.ParseUint(s, 10, 16)
 	if err != nil {
 		return types.CantParseUint64(err, s)
 	}
```
julianbrost, line 21: > CantParse**Uint64**

That will create a misleading error message: https://github.com/Icinga/icinga-go-library/blob/bb5b3747011a3d20befc17dafa0757f2eb3eb1e6/types/utils.go#L32-L35

There are no corresponding functions for other sizes than 64 so far unfortunately. :(

pkg/icingadb/history/retention.go
```diff
@@ -172,6 +173,20 @@ func (r *Retention) Start(ctx context.Context) error {
...
+			// First, make sure that days do not exceed max int, as otherwise casts will overflow and later subtractions
+			// may overflow as well. By being capped to [0, max int), for each current timestamp t, t-days > min int.
+			// As the database schema uses an unsigned timestamp, the second check ensures that the relative offset,
+			// later called olderThan, does not overflow.
+			//
+			// For the periodic.Start function below, time marches forward while the value of days is static.
+			// Subtracting days from an increasing timestamp will also not result in a negative or overflowing value.
+			r.logger.Errorf("Skipping history retention for category %s as %d days will overflow", stmt.Category, days)
+			continue
+		}
```
julianbrost, line 189: > `days > math.MaxInt`

That still allows for `days` = $2^{63} - 1$ in 64 bit builds which is too much to handle for `AddDate()`.

> By being capped to [0, max int), for each current timestamp t, **t-days > min int**.

How is min int relevant here? Wouldn't be the min `time.Time` be the relevant thing here (or maybe the internals of `AddDate()` if that breaks even earlier)?

From https://github.com/Icinga/icingadb/pull/821#discussion_r1792140287:

> reducing days to uint16 sounds good to me too. Though I would do that in a separate PR.

Though doing it in a separate PR sounds like this would either be a prerequisite of this PR or this PR has to do a similar bounds check. Doing another PR first doesn't sound like a bad idea for me from a changelog perspective: a separate PR for "limit history retention to $2^16$ days (~180 years)" might be more helpful than hiding it in a linter PR.

pkg/icingadb/cleanup.go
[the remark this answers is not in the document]
julianbrost: Ah, yes, the muscle memory of typing `if err != nil` all the time :sweat_smile:

cmd/icingadb-migrate/convert.go
oxzi: I have extracted it into a separate block, allowing to get rid of the linter hint.

pkg/icingadb/history/retention.go
oxzi: The comment contained "outdated" fractions and implementation knowledge of `time.Time`. I have cleaned that up.

Furthermore, I have added a third check[^1] to check if the calculated time with `AddDate()` is after the current date, implying an overflow in the `AddDate()` internals.

However, I can also change the type to `uint16` and will prepare another PR for that. This will still require the current time comparison check against Unix epoch as subtraction `math.MaxUint16` days will currently end up in [1845](https://go.dev/play/p/7oXTiOTZ3Nn).

[^1]: yay, more checks

oxzi: The type change was proposed in #824.

julianbrost: > The type change was proposed in #824.

And it was just merged, which should actually make the `MaxInt` check and this new check already obsolete:

> Furthermore, I have added a third check[1](#user-content-fn-1-bb8cabe0a4b2ff53552d596385fb8e0b) to check if the calculated time with `AddDate()` is after the current date, implying an overflow in the `AddDate()` internals.

oxzi: Now this code is way easier to read :)

julianbrost: > Though doing it in a separate PR sounds like this would either be a prerequisite of this PR or [...]

Well, there was a reason why I said that earlier :sweat_smile:

pkg/icingadb/history/retention.go
```diff
@@ -173,6 +173,13 @@ func (r *Retention) Start(ctx context.Context) error {
 			continue
 		}
 
+		if time.Now().AddDate(0, 0, -int(days)).Before(time.Unix(0, 0)) {
+			r.logger.Errorf("Skipping history retention for category %s: "+
+				"%d days of retention will result in negative timestamps, overflowing unsigned timestamps",
+				stmt.Category, days)
+			continue
+		}
```
julianbrost, line 181: My first though: this could be moved closer to where the overflow would happen. This is the function that will actually use the value:

https://github.com/Icinga/icingadb/blob/a6d134c7f3e2d41e9ae2453f08e9672430f9c188/pkg/icingadb/cleanup.go#L21-L30

Adding the following right at the beginning would actually be a valid implementation of the function:

```go
// Timestamps in the database are unsigned, so there can't be any rows with a timestamp smaller than
// a negative value and the query can be skipped.
if olderThan.Before(time.Unix(0, 0)) {
	return 0, nil
}
```

My second though: oh, looks like there's actually no overflow at all. Like the timestamp is only passed around as `UnixMilli` in the code:

https://github.com/Icinga/icingadb/blob/a6d134c7f3e2d41e9ae2453f08e9672430f9c188/pkg/icingadb/cleanup.go#L40-L43

And it doesn't seem to ever be converted to `uint64`. And the `Value()` implementation of `UnixMilli` just returns the `int64` returned by [`time.Time.UnixMilli()`](https://pkg.go.dev/time#Time.UnixMilli):
https://github.com/Icinga/icinga-go-library/blob/bb5b3747011a3d20befc17dafa0757f2eb3eb1e6/types/unix_milli.go#L87-L95

So it will just generate a query with something like `WHERE event_time < -42` which won't match any rows but should be fine otherwise I guess? So I doubt we need this `if` at all. Like running in an overflow in scenario sounds plausible, but not with the current implementation.

oxzi: > My first though: this could be moved closer to where the overflow would happen. This is the function that will actually use the value:
> 
> https://github.com/Icinga/icingadb/blob/a6d134c7f3e2d41e9ae2453f08e9672430f9c188/pkg/icingadb/cleanup.go#L21-L30
> 
> Adding the following right at the beginning would actually be a valid implementation of the function:
> 
> ```go
> // Timestamps in the database are unsigned, so there can't be any rows with a timestamp smaller than
> // a negative value and the query can be skipped.
> if olderThan.Before(time.Unix(0, 0)) {
> 	return 0, nil
> }
> ```

A similar idea also came to me, but if the condition is true at the current position, it will also be true (at least for a while) within the `periodic.Start` callback function. Thus, the function will be a no-op within a loop. Doesn't result in load, but seems a bit unnecessary.

> My second though: oh, looks like there's actually no overflow at all. Like the timestamp is only passed around as `UnixMilli` in the code:
> 
> https://github.com/Icinga/icingadb/blob/a6d134c7f3e2d41e9ae2453f08e9672430f9c188/pkg/icingadb/cleanup.go#L40-L43
> 
> And it doesn't seem to ever be converted to `uint64`. And the `Value()` implementation of `UnixMilli` just returns the `int64` returned by [`time.Time.UnixMilli()`](https://pkg.go.dev/time#Time.UnixMilli): https://github.com/Icinga/icinga-go-library/blob/bb5b3747011a3d20befc17dafa0757f2eb3eb1e6/types/unix_milli.go#L87-L95
> 
> So it will just generate a query with something like `WHERE event_time < -42` which won't match any rows but should be fine otherwise I guess? So I doubt we need this `if` at all. Like running in an overflow in scenario sounds plausible, but not with the current implementation.

This seems right. However, I wanted to verify this with setting `retention.history-days` to 60,000 - ending up somewhere in the 1860s. After enabling debug logs, I found this query:
```
26 Execute  DELETE FROM notification_history WHERE environment_id = '\\Nha\\\/\\\\\\)W\\\' AND send_time < -3455367174796 ORDER BY send_time LIMIT 5000
```

The timestamp seems valid:
```
date -d @-3455367174.796
Tue Jul  3 08:40:33 AM LMT 1860
```

Manually running queries with negative timestamps also works, of couse. Thus, yes, you are right. I have removed the check now :)

lippserd (approved):

PR #824, opened 2024-10-10
RetentionConfig: restrict days to uint16

oxzi:
As discussed in #821, allowing huge retention values - as the type uint64 allows - may result in overflows. Especially Go's time.AddDate() method silently overflows for huge values, resulting in unexpected comparison times.

> $ ./icingadb --config <(echo 'retention: {history-days: -1}')
> can't parse YAML file /proc/self/fd/11: cannot unmarshal -1 into Go value of type uint16 ( overflow )
>
> $ ./icingadb --config <(echo 'retention: {history-days: 100000}')
> can't parse YAML file /proc/self/fd/11: cannot unmarshal 100000 into Go value of type uint16 ( overflow )

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -55,8 +55,8 @@ type Flags struct {
 
 // RetentionConfig defines configuration for history retention.
 type RetentionConfig struct {
-	HistoryDays uint64                   `yaml:"history-days"`
-	SlaDays     uint64                   `yaml:"sla-days"`
+	HistoryDays uint16                   `yaml:"history-days"`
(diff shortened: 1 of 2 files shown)


julianbrost (approved):
$2^{16}$ days is almost 180 years, so that range should be more than enough for all practical use (uint8 with less than a year would be too small). On the other hand, now - 180 years, i.e. the year 1844 is within `time.Time` bounds and no problem for `time.AddDate()`.

lippserd (approved):

PR #825, opened 2024-10-11
HA State: ensure not this and another instance can be responsible

oxzi:
In theory, this should not happen. This assumption is based on the trust in the database transaction performing the HA realization logic. However, one debugged log let one assume that this happened anyway.

This change mostly signals an error while also explicitly giving up the HA state. Doing so should at least alarm a person reading the logs.

---

ref/IP/55850

### DIFF
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -431,6 +431,11 @@ func (h *HA) realize(
 
 		h.signalTakeover(takeover)
 	} else if otherResponsible {
+		if state := h.state.Load(); state.responsible {
+			h.logger.Error("Other instance is responsible while this node itself is responsible, dropping responsibility")
+			h.signalHandover("other instance is responsible as well")
+			// h.signalHandover will update h.state
+		}
 		if state := h.state.Load(); !state.otherResponsible {
 			// Dereference pointer to create a copy of the value it points to.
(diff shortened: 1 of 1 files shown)


oxzi:
For the record, this behavior was seen in the linked ticket a short time after something in the database cluster went south.

Furthermore, I have rebased the branch to reflect the latest changes on the main branch.

lippserd (approved):
@oxzi If I'm not mistaken, the changes introduced were not triggered (anymore) after #800 was used in the referenced ticket, right? Anyway, a little more protection can't hurt.

lippserd:
> The build in the mentioned ticket contains both this branch as well as #800.

Yes, I know. I meant that with #800 there were no (more) logs that say "Other instance is responsible...". So I wanted to question whether this change is actually necessary, but at the same time say that more protection can't hurt.

oxzi:
> > The build in the mentioned ticket contains both this branch as well as #800.
> 
> Yes, I know. I meant that with #800 there were no (more) logs that say "Other instance is responsible...", right? So I wanted to question whether this change is actually necessary, but at the same time say that more protection can't hurt.

Unfortunately, there was no deployment of only #800, but always of these two PRs combined. Furthermore, I never got detailed enough logs to be able to answer this question. Thus, at least in this very scenario, both branches combined seemed to solve or at least mitigate the issue.

PR #830, opened 2024-10-25
HA: Reduce deadlocks via exclusive locking (`SELECT ... FOR UPDATE`)

yhabteab:
Within the HA transaction, we always retrieve rows from the icingadb_instance table to determine responsibility between instances. Previously, this selection was performed using a shared lock (`SELECT ... LOCK IN SHARE MODE`). Although selecting rows is generally not an issue, both Icinga DB instances perform an upsert on the same table afterward, resulting in deadlocks most of the time. In order to reduce the deadlocks on both sides, an exclusive lock on the selected rows is necessary, which can be achieved using the `SELECT ... FOR UPDATE` command. However, deadlocks can sill occur if the `icingadb_instance` table is empty and now rows are available to lock exclusively.

## Tests

```sql
MariaDB [test_test]> show create table icingadb_instance\G
       Table: icingadb_instance
Create Table: CREATE TABLE `icingadb_instance` (
  `id` int(11) NOT NULL,
  `environment_id` int(11) NOT NULL,
  `heartbeat` int(11) NOT NULL,
  `responsible` enum('n','y') NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC
```

<details><summary>Empty Table</summary>

```sql
MariaDB [test_test]> SELECT * FROM icingadb_instance;
Empty set (0.001 sec)
```

Note that none of the variants can prevent deadlocks when the table is empty.

| Tx A | Tx B |
|--------|--------|
| `START TRANSACTION;` | `START TRANSACTION;` |
| `SELECT id, heartbeat FROM icingadb_instance WHERE environment_id = 1 AND responsible = 'y' AND id <> 23 LOCK IN SHARE MODE;` <br> Empty set (0.001 sec) | `SELECT id, heartbeat FROM icingadb_instance WHERE environment_id = 1 AND responsible = 'y' AND id <> 42 LOCK IN SHARE MODE;` <br> Empty set (0.001 sec) |
| `INSERT INTO icingadb_instance VALUES (23, 1, 123, 'y') ON DUPLICATE KEY UPDATE id=VALUES(id), environment_id=VALUES(environment_id), heartbeat=VALUES(heartbeat), responsible=VALUES(responsible);` <br><br> **It was waiting for a lock until TxB was aborted due to the deadlock.** <br><br> Query OK, 1 row affected (1.668 sec) | `INSERT INTO icingadb_instance VALUES (42, 1, 123, 'y') ON DUPLICATE KEY UPDATE id=VALUES(id),environment_id=VALUES(environment_id),heartbeat=VALUES(heartbeat),responsible=VALUES(responsible);` <br><br> ERROR 1213 (40001): Deadlock found when trying to get lock; try restarting transaction |

```sql
MariaDB [test_test]> SELECT * FROM icingadb_instance;
+----+----------------+-----------+-------------+
| id | environment_id | heartbeat | responsible |
+----+----------------+-----------+-------------+
| 23 |              1 |       123 | y           |
+----+----------------+-----------+-------------+
```

</details> 

<details><summary>Non Empty Table</summary>

### Before

```sql
MariaDB [test_test]> SELECT * FROM icingadb_instance;
+----+----------------+-----------+-------------+
| id | environment_id | heartbeat | responsible |
+----+----------------+-----------+-------------+
| 23 |              1 |       123 | y           |
+----+----------------+-----------+-------------+
```

Result: Same as with an empty table, one Tx will deadlock while the other will succeed, typically the one that starts the upsert statement first.

### After

```sql
MariaDB [test_test]> SELECT * FROM icingadb_instance;
+----+----------------+-----------+-------------+
| id | environment_id | heartbeat | responsible |
+----+----------------+-----------+-------------+
| 23 |              1 |       123 | y           |
+----+----------------+-----------+-------------+
```

| Tx A | Tx B |
|--------|--------|
| `START TRANSACTION;` | `START TRANSACTION;` |
| `SELECT id, heartbeat FROM icingadb_instance WHERE environment_id = 1 AND responsible = 'y' AND id <> 23 FOR UPDATE;` <br> Empty set (0.003 sec) | `SELECT id, heartbeat FROM icingadb_instance WHERE environment_id = 1 AND responsible = 'y' AND id <> 42 LOCK FOR UPDATE;` <br> Blocked 🚫 (in waiting state) |
| `INSERT INTO icingadb_instance VALUES (23, 1, 123, 'y') ON DUPLICATE KEY UPDATE id=VALUES(id), environment_id=VALUES(environment_id), heartbeat=VALUES(heartbeat), responsible=VALUES(responsible);` <br><br> Query OK, 0 rows affected (0.001 sec) | The select statement from above is still blocked |
| `COMMIT;` <br> Query OK, 0 rows affected (0.001 sec) | SELECT Query from above succeeds and returned one row |

</details>

PS: this is basically the same as #788 but adds some before and after test cases as requested in https://github.com/Icinga/icingadb/pull/788#issuecomment-2388190057.

closes #788

### DIFF
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -294,9 +294,15 @@ func (h *HA) realize(
 
 			if h.db.DriverName() == database.MySQL {
 				// The RDBMS may actually be a Percona XtraDB Cluster which doesn't
-				// support serializable transactions, but only their following equivalent:
+				// support serializable transactions, but only their equivalent SELECT ... LOCK IN SHARE MODE.
+				// However, in order to reduce the deadlocks on both sides, it is necessary to obtain an exclusive lock
+				// on the selected rows. This can be achieved by utilising the SELECT ... FOR UPDATE command.
+				// Nevertheless, deadlocks may still occur, when the "icingadb_instance" table is empty, i.e. when
+				// there's no available row to be locked exclusively.
+				//
+				// See also: https://dev.mysql.com/doc/refman/5.7/en/innodb-locking-reads.html
 				isoLvl = sql.LevelRepeatableRead
-				selectLock = " LOCK IN SHARE MODE"
+				selectLock = " FOR UPDATE"
 			}
 
 			tx, errBegin := h.db.BeginTxx(ctx, &sql.TxOptions{Isolation: isoLvl})


lippserd (commented):
Should be tested with PostgreSQL as well. Anyway, I think we can just remove the driver switch and use FOR UPDATE for MySQL and PostgreSQL.

yhabteab:
> Should be tested with PostgreSQL as well.

See the updated PR description!

> Anyway, I think we can just remove the driver switch and use FOR UPDATE for MySQL and PostgreSQL.

Done!

oxzi (approved):

lippserd (approved):

PR #831, opened 2024-10-26
Support loading configuration from both YAML files and env vars

lippserd:
requires Icinga/icinga-go-library#83


internal/config/config.go
```diff
@@ -52,8 +52,25 @@ func (c *Config) Validate() error {
...
+	if f.Config == "" {
+		return DefaultConfigPath
+	}
+
+	return f.Config
+}
+
+// IsExplicitConfigPath indicates whether the configuration file path was explicitly set.
+func (f Flags) IsExplicitConfigPath() bool {
+	return f.Config != ""
```
oxzi, line 73: As commented above, this does not seem to work.

Thus, I am unable to parse a configuration based solely on environment variables as the code always bails out due the unavailability of the `/etc/icingadb/config.yml` file, <https://github.com/Icinga/icinga-go-library/blob/7ce60227c33d68d1a9d5a958a2c2f9409b05168c/config/config.go#L273>.

internal/config/config.go
```diff
@@ -49,8 +52,25 @@ func (c *Config) Validate() error {
...
+	Config string `short:"c" long:"config" description:"path to config file" default:"/etc/icingadb/config.yml"`
+	// default must be kept in sync with DefaultConfigPath.
+}
+
+// GetConfigPath retrieves the path to the configuration file.
+// It returns the path specified via the command line, or DefaultConfigPath if none is provided.
+func (f Flags) GetConfigPath() string {
+	if f.Config == "" {
+		return DefaultConfigPath
+	}
```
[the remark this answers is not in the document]
lippserd: 😆 After requesting your review, I thought it would be cool if the default value was still displayed, but I didn't consider the consequences and admittedly didn't test it:

https://github.com/Icinga/icingadb/compare/614d015e8d3e3c9846451e96edbe9064d7e508cc..8f5e40e2e193b4b5f7e55ed29ea5f0e278306a33

I will add the default value to the description instead. Sorry for the trouble!

internal/config/config.go
```diff
@@ -52,8 +52,25 @@ func (c *Config) Validate() error {
 type Flags struct {
```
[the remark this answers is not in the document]
lippserd: Will do.

internal/command/command.go
```diff
@@ -38,7 +38,10 @@ func New() *Command {
 	}
 
 	var cfg icingadbconfig.Config
-	if err := config.FromYAMLFile(flags.Config, &cfg); err != nil {
+	if err := config.Load(&cfg, config.LoadOptions{
+		Flags:      flags,
+		EnvOptions: config.EnvOptions{Prefix: "ICINGADB_"},
+	}); err != nil {
```
oxzi, line 44: This works as far as I have tested it.

However, the error output may be misleading for a non-existing `/etc/icingadb/config.yml` file when not explicitly setting the `--config` flag, relying on the default.

```
$ stat /etc/icingadb/config.yml
stat: cannot statx '/etc/icingadb/config.yml': No such file or directory

$ ./icingadb
invalid configuration: database host missing
$ ./icingadb -c /etc/icingadb/config.yml
can't open YAML file /etc/icingadb/config.yml: open /etc/icingadb/config.yml: no such file or directory
```

If there are no `ICINGADB_` environment variables and the default config is not existing, the reported error does not state that the expected file does not exist, but says that certain fields are absent. But honestly, I don't know how to address this. For a package installations the config file should be present.

internal/config/config.go
oxzi, line 1: There is also the `config_test.go` file, testing against the now unused `config.FromYAMLFile` function. Maybe you want to remove this test case and create another one for YAML-only, environment-only, and mixed-setups.

internal/command/command.go
lippserd: What about the following?

> invalid configuration: database host missing
> default config file /etc/icingadb/config.yml does not exist but can be ignored if the configuration is intended to be entirely provided via environment variables

oxzi: Would like that, as it not only tells what is wrong, but also gives suggestions to act upon.

oxzi (commented):
Having documentation (like human readable docs) for this PR would be nice. Even just stating that this is possible in the `03-Configuration` section.

internal/config/config.go
```diff
@@ -10,12 +10,15 @@ import (
...
 // Config defines Icinga DB config.
 type Config struct {
-	Database  database.Config `yaml:"database"`
-	Redis     redis.Config    `yaml:"redis"`
-	Logging   logging.Config  `yaml:"logging"`
-	Retention RetentionConfig `yaml:"retention"`
+	Database  database.Config `yaml:"database" envPrefix:"DATABASE_"`
+	Redis     redis.Config    `yaml:"redis" envPrefix:"REDIS_"`
+	Logging   logging.Config  `yaml:"logging" envPrefix:"LOGGING_"`
+	Retention RetentionConfig `yaml:"retention" envPrefix:"RETENTION_"`
```
oxzi, line 21: `RetentionConfig` still misses `env` struct tags.

oxzi (commented):
I have taken another look at why the CI fails and after some digging, it became quite obvious. The Icinga DB container bailed out as it cannot parse its config.

> can't parse YAML file /icingadb.yml: does not implemented Unmarshaler

However, the generated config looked fine.

```
database:
  type: pgsql
  host: 172.20.0.4
  port: 5432
  database: d1
  user: u1
  password: 7f96cc1ac9d3ad03
redis:
  host: 172.20.0.2
  port: 6379
logging:
  level: debug
  interval: 1s
retention:
  history-days: 7
  sla-days: 30
  options:
    acknowledgement: 0
    comment: 1
    downtime: 2
```

So I went down the YAML parsing and ended up at a faulty YAML parsing for the `RetentionOptions`.

Turns out, the custom `encoding.TextUnmarshaler` is not enough on its own, but a `yaml.InterfaceUnmarshaler`is also required for custom types.

I pushed a hopefully fix to the just created `config-from-environment-variables-fix-alvar` branch in 0572d6dfca23358876dfa938fa1597f28bd2b068. At least on my machine all tests went through.

It would be good to have unit test for this, similar to `RetentionOptions` to catch this earlier on.

lippserd:
@oxzi Thanks for the correction. We already had a similar issue in our Go library and I'll link it here for reference:

https://github.com/Icinga/icinga-go-library/pull/84

From my point of view, the changes from the PR can be merged now.

PR #878, opened 2024-12-17
HA: Use Transaction for each query

oxzi:
During #800, the commit dd0ca8fb072ae8713b7e3f32df5b6417aa0fb7b6 inlined the HA.insertEnvironment method into the retryable realization function. However, while doing so, I forgot to change the query execution context from h.db to tx. This resulted in an error when being used together with a single database connection, as introduced in #828.

---

Thanks for @yhabteab for the pair debug session :)


lippserd (approved):
Well, that was an unfortunate oversight. Thanks for troubleshooting!

PR #883, opened 2025-01-24
Retention: switch number of days type uint16 -> float64

Al2Klimov:
This allows specifying less than a day as threshold. That is especially useful if history doesn't mattter as much as disk space.

I want my large (ref/NC/820479) test setup to run over WE w/o running out of disk space. So I built this.

### DIFF
--- a/pkg/icingadb/history/retention.go
+++ b/pkg/icingadb/history/retention.go
@@ -11,6 +11,7 @@ import (
 	"github.com/icinga/icingadb/pkg/icingaredis/telemetry"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
+	"math"
 	"strconv"
 	"strings"
 	"time"
@@ -97,7 +98,7 @@ var RetentionStatements = []retentionStatement{{
 }}
 
 // RetentionOptions defines the non-default mapping of history categories with their retention period in days.
-type RetentionOptions map[string]uint16
+type RetentionOptions map[string]float64
 
 // UnmarshalText implements [encoding.TextUnmarshaler] to allow RetentionOptions to be parsed by env.
 //
@@ -105,7 +106,7 @@ type RetentionOptions map[string]uint16
 // After <https://github.com/caarlos0/env/pull/323> got merged and a new env release was drafted, this method can be
 // removed.
 func (o *RetentionOptions) UnmarshalText(text []byte) error {
-	optionsMap := make(map[string]uint16)
+	optionsMap := make(map[string]float64)
 
 	for _, pair := range strings.Split(string(text), ",") {
 		key, value, found := strings.Cut(pair, ":")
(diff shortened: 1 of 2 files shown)


pkg/icingadb/history/retention.go
```diff
@@ -177,12 +178,21 @@ func (r *Retention) Start(ctx context.Context) error {
...
+			zap.Float32("retention-days", days),
 		)
 
 		stmt := stmt
 		periodic.Start(ctx, r.interval, func(tick periodic.Tick) {
-			olderThan := tick.Time.AddDate(0, 0, -int(days))
+			olderThan := tick.Time
+			wholeDays, dayFraction := math.Modf(float64(days))
+
+			if wholeDays > 0 {
```
oxzi, line 189: Can this be negative? Above in line 171/172 it is already checked if `days < 1`.

Btw, you may want to change the abort constraint to `days < 0` as otherwise you cannot use half days, right?

julianbrost (commented):
If I remember correctly, that config attribute only exists in this way instead of being a standard `time.Duration` because the largest unit `time.ParseDuration` is hours and you'd have to configure large non-obvious values for realistic use-cases. An alternative could be to allow more advanced duration strings like for example `history: 1y`, then the change would also provide some real-world improvement.

And another thing to consider: could allowing very small values conflict with writing the history? Like could this trigger foreign key errors because the referenced row was already deleted? The minimum of 1 day might actually provide some safety margin there.

pkg/icingadb/history/retention.go
Al2Klimov: > Can this be negative?

I guess you answered your own question:

> Above in line 171/172 it is already checked if `days < 1`.

So no, it can't be.

(If I understand you correctly.)

pkg/icingadb/history/retention.go
```diff
@@ -177,12 +178,21 @@ func (r *Retention) Start(ctx context.Context) error {
...
+				olderThan = olderThan.AddDate(0, 0, -int(wholeDays))
+			}
+
+			if dayFraction > 0 {
+				olderThan = olderThan.Add(-time.Duration(dayFraction * float64(24*time.Hour)))
+			}
```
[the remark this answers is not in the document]
Al2Klimov: IIRC, only a limited number of days fit into time.Duration. I'm working around that limitation here.

Al2Klimov:
> The minimum of 1 day might actually provide some safety margin there.

E.g 1h safety margin would also do it.

Al2Klimov:
BUT then I'd tell users to increase the threshold.

oxzi:
Allowing fractions for days seems wrong. While I understand the use case, I would suggest not misusing `retention-days` but change this to a more generic `reteion-period` field, expecting a Go duration. Of course, this would need some migration, potentially making `retention-days` deprecated first before removing it.

However, before starting to implement such a change, some other opinions may be useful.

lippserd:
> Allowing fractions for days seems wrong. While I understand the use case, I would suggest not misusing `retention-days` but change this to a more generic `reteion-period` field, expecting a Go duration. Of course, this would need some migration, potentially making `retention-days` deprecated first before removing it.
> 
> However, before starting to implement such a change, some other opinions may be useful.

I wouldn't change anything and just close this PR until we have a specific feature request.

oxzi:
Thanks for your second opinion.

As having a retention period of less than a day is more a development than a production requirement and that the config either "feels" wrong or requires a breaking change, I would close this now.

Thanks for all your work, @Al2Klimov. If the needs arises, we can resurrect this PR.

PR #897, opened 2025-03-05
doc/03-Configuration.md: Environment Variables

oxzi:
New ways to configure Icinga DB were introduced in https://github.com/Icinga/icingadb/pull/831 and https://github.com/Icinga/icinga-go-library/pull/113. The first change allows configuring Icinga DB using environment variables instead of or next to the YAML configuration file. In addition, the second change allows setting certificates and keys as PEM-encoded strings next to referencing files. This was now documented.

As a structural change, the order of the Database and Redis sections were changed to reflect the order in the example configuration file.

Some words about the Logging Components were written, as the documentation previously that they exist.

---

Requires:
- [ ] #831 
- [x] Icinga/icinga-go-library#113
- [x] Release of the Icinga Go Library and dependency update here.

### DIFF
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -1,35 +1,27 @@
 # Configuration
 
-For package installations, the configuration is stored in `/etc/icingadb/config.yml`.
+Icinga DB can be configured using a YAML configuration file, environment variables, or both.
+Environment variables take precedence and override previously defined values from the configuration file.
+
+For package installations, the configuration file is stored in `/etc/icingadb/config.yml`.
 See [config.example.yml](../config.example.yml) for an example configuration.
 
-## Redis® Configuration
-
-Connection configuration for the Redis® server where Icinga 2 writes its configuration, state and history items.
-This is the same connection as configured in the
-[Icinga DB feature](https://icinga.com/docs/icinga-2/latest/doc/14-features/#icinga-db) of
-the corresponding Icinga 2 node. High availability setups require a dedicated Redis® server per Icinga 2 node and
-therefore a dedicated Icinga DB instance that connects to it.
-
-| Option   | Description                                                                                                             |
-|----------|-------------------------------------------------------------------------------------------------------------------------|
-| host     | **Required.** Host name or address, or absolute Unix socket path.                                                       |
-| port     | **Optional.** TCP port. Defaults to `6380` matching the Redis® open source server port in the `icingadb-redis` package. |
-| username | **Optional.** Authentication username, requires a `password` being set as well.                                         |
-| password | **Optional.** Authentication password. May be used alone or together with a `username`.                                 |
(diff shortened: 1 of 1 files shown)


doc/03-Configuration.md
```diff
@@ -101,6 +150,17 @@ There are separate options for the full history tables used to display history i
...
 
+For YAML configuration, the options are part of the `retention` dictionary.
+For environment variables, each option is prefixed with `ICINGADB_RETENTION_`.
+
+When using environment variables, the Retention `options` are formatted similar to the [logging components](#logging-components) from above.
+
+```
+# Environment Variables
+ICINGADB_RETENTION_HISTORY_DAYS=14
+ICINGADB_RETENTION_OPTIONS=comment:356,sla:90
```
lippserd, line 161: This will yield `invalid configuration: invalid key sla for history retention`. There's `sla-days` for sla.

lippserd (changes requested):
Looks solid and comprehensible, thank you very much. There is only one correction that needs to be made.

doc/03-Configuration.md
oxzi: Wait, shouldn't it be `sla_downtime` or `sla_state`? Then this is also wrong/outdated in the table below.

https://github.com/Icinga/icingadb/blob/0eb1ad8c4f0485a3bc06cb8ed70ed3a9bd672d82/pkg/icingadb/history/retention.go#L110-L114
https://github.com/Icinga/icingadb/blob/0eb1ad8c4f0485a3bc06cb8ed70ed3a9bd672d82/pkg/icingadb/history/retention.go#L79-L94

The `sla-days` should be `ICINGADB_RETENTION_SLA_DAYS`. I am going to alter the change.

oxzi: I have synchronized the retention categories from the `RetentionStatements` to both the docs and the example configuration. Good catch, btw! Please feel free to take another look.

oxzi: Oh wait. I messed up. The two `sla_` keys are of the `RetentionSla` type, not `RetentionHistory`..

oxzi: Now it should be fine. The `sla` entry in `options` seems to be a forgotten artifact and I confused myself after skimming the common code. Please feel free to take a look now :>

lippserd: > Oh wait. I messed up. The two `sla_` keys are of the `RetentionSla` type, not `RetentionHistory`..

Yes, `sla` cannot be further customized. There is only `sla-days`.

> Now it should be fine. The `sla` entry in `options` seems to be a forgotten artifact and I confused myself after skimming the common code. Please feel free to take a look now :>

I checked the example configuration and our code for a leftover, but couldn't find any. Is there anything else to do? Maybe I just misunderstood your comment :).

lippserd (approved):
This can be marked as ready for merging from my side.

PR #901, opened 2025-03-11
Auto Import Database Schema

### LINKED ISSUE #896: Command Line Flag For Schema Import
In order to get rid of `docker-icinagdb` and its entrypoint command, the initial schema import must be supported by Icinga DB itself.

The idea goes as follows: Add a new command line argument flag to Icinga DB to check and potentially import the schema. Within the Dockerfile, Icinga DB gets started with this flag, effectually mimicking [`initDb()`](https://github.com/Icinga/docker-icingadb/blame/a362254223b65f9c74e96395710b6d346c63c368/entrypoint/main.go#L108).

oxzi:
To get rid of docker-icingadb and its additional entry point, the schema import functionality has been implemented directly in Icinga DB. Using the new `--database-auto-import` command line argument will result in an automatic schema import if no schema is found.

The implementation is split between the already existing `CheckSchema` function and the introduced `ImportSchema` function.

The `CheckSchema` function is now able to distinguish between the absence of a schema and an incorrect schema version. Both situations return a separate error type.

As before, `CheckSchema` is called in the `main` function. If the error type now implies the absence of a schema (`ErrSchemaNotExists`) and the `--database-auto-import` flag is set, the auto-import is started.

The schema import itself is performed in the new `ImportSchema` function, which loads the schema from a given file and inserts it within a transaction, allowing to rollback in case of an error.

<details><summary>MySQL Run</summary>
<p>

```
$ docker run --name mariadb-icingadb -e MARIADB_ROOT_PASSWORD=supersicher -p 3306:3306 -d mariadb:latest

$ docker exec -it mariadb-icingadb mariadb --password=supersicher
MariaDB [(none)]> create database icingadb;
Query OK, 1 row affected (0.001 sec)
MariaDB [(none)]> ^D

$ # Start w/o a schema
$ ./icingadb --config config.yml
2025-03-27T09:40:43.379+0100    INFO    icingadb        Starting Icinga DB daemon (1.2.1-g3768602-dirty)
2025-03-27T09:40:43.379+0100    INFO    icingadb        Connecting to database at 'mysql://root@localhost:3306/icingadb'
2025-03-27T09:40:43.383+0100    FATAL   icingadb        The database schema is missing


$ # Start w/o a schema, but allow imports
$ ./icingadb --config config.yml --database-auto-import
2025-03-27T09:40:46.378+0100    INFO    icingadb        Starting Icinga DB daemon (1.2.1-g3768602-dirty)
2025-03-27T09:40:46.378+0100    INFO    icingadb        Connecting to database at 'mysql://root@localhost:3306/icingadb'
2025-03-27T09:40:46.383+0100    INFO    icingadb        Starting database schema auto import
2025-03-27T09:40:46.711+0100    INFO    icingadb        The database schema was successfully imported
[ . . . ]

$ # Simulate an outdated schema version
$ docker exec -it mariadb-icingadb mariadb --password=supersicher icingadb
MariaDB [icingadb]> update icingadb_schema set version = 1;
Query OK, 1 row affected (0.008 sec)
Rows matched: 1  Changed: 1  Warnings: 0
MariaDB [icingadb]> ^D
$ ./icingadb --config config.yml
2025-03-27T09:41:37.313+0100    INFO    icingadb        Starting Icinga DB daemon (1.2.1-g3768602-dirty)
2025-03-27T09:41:37.313+0100    INFO    icingadb        Connecting to database at 'mysql://root@localhost:3306/icingadb'
2025-03-27T09:41:37.317+0100    FATAL   icingadb        unexpected database schema version: v1 (expected v6), please make sure you have applied all database migrations after upgrading Icinga DB

$ # Reset schema version
$ docker exec -it mariadb-icingadb mariadb --password=supersicher icingadb
MariaDB [icingadb]> update icingadb_schema set version = 6;
Query OK, 1 row affected (0.008 sec)
Rows matched: 1  Changed: 1  Warnings: 0
MariaDB [icingadb]> ^D

$ # Start w/ correct schema
./icingadb --config config.yml
2025-03-27T09:42:00.220+0100    INFO    icingadb        Starting Icinga DB daemon (1.2.1-g3768602-dirty)
2025-03-27T09:42:00.220+0100    INFO    icingadb        Connecting to database at 'mysql://root@localhost:3306/icingadb'
[ . . . ]

$ # Retry without a database
$ docker stop mariadb-icingadb
$ docker container rm mariadb-icingadb
$ ./icingadb --config config.yml
2025-03-27T09:42:24.372+0100    INFO    icingadb        Starting Icinga DB daemon (1.2.1-g3768602-dirty)
2025-03-27T09:42:24.372+0100    INFO    icingadb        Connecting to database at 'mysql://root@localhost:3306/icingadb'
2025-03-27T09:42:24.373+0100    WARN    database        Can't connect to database. Retrying     {"error": "dial tcp [::1]:3306: connect: connection refused"}
[ . . . ]
```

</p>
</details>

<details><summary>PostgreSQL Run</summary>
<p>

```
$ docker run --name postgres-icingadb -e POSTGRES_PASSWORD=supersicher -p 5432:5432 -d postgres:latest

$ docker exec -it postgres-icingadb psql -U postgres
postgres=# create database icingadb;
CREATE DATABASE
postgres=# ^D

$ # Start w/o a schema
$ ./icingadb --config config.yml
2025-03-27T09:43:12.578+0100    INFO    icingadb        Starting Icinga DB daemon (1.2.1-g3768602-dirty)
2025-03-27T09:43:12.578+0100    INFO    icingadb        Connecting to database at 'pgsql://postgres@localhost:5432/icingadb'
2025-03-27T09:43:12.594+0100    FATAL   icingadb        The database schema is missing

$ # Start w/o a schema, but allow imports
$ ./icingadb --config config.yml --database-auto-import
2025-03-27T09:43:24.688+0100    INFO    icingadb        Starting Icinga DB daemon (1.2.1-g3768602-dirty)
2025-03-27T09:43:24.688+0100    INFO    icingadb        Connecting to database at 'pgsql://postgres@localhost:5432/icingadb'
2025-03-27T09:43:24.703+0100    INFO    icingadb        Starting database schema auto import
2025-03-27T09:43:24.765+0100    INFO    icingadb        The database schema was successfully imported
[ . . . ]

$ # Simulate an outdated schema version
$ docker exec -it postgres-icingadb psql -U postgres icingadb
icingadb=# update icingadb_schema set version = 1;
UPDATE 1
icingadb=# ^D
$ ./icingadb --config config.yml
2025-03-27T09:43:53.056+0100    INFO    icingadb        Starting Icinga DB daemon (1.2.1-g3768602-dirty)
2025-03-27T09:43:53.056+0100    INFO    icingadb        Connecting to database at 'pgsql://postgres@localhost:5432/icingadb'
2025-03-27T09:43:53.065+0100    FATAL   icingadb        unexpected database schema version: v1 (expected v4), please make sure you have applied all database migrations after upgrading Icinga DB

$ # Reset schema version
$ docker exec -it postgres-icingadb psql -U postgres icingadb
icingadb=# update icingadb_schema set version = 4;
UPDATE 1
icingadb=# ^D

$ # Start w/ correct schema
./icingadb --config config.yml
2025-03-27T09:44:09.194+0100    INFO    icingadb        Starting Icinga DB daemon (1.2.1-g3768602-dirty)
2025-03-27T09:44:09.194+0100    INFO    icingadb        Connecting to database at 'pgsql://postgres@localhost:5432/icingadb'
[ . . . ]

$ # Retry without a database
$ docker stop postgres-icingadb
$ docker container rm postgres-icingadb
$ ./icingadb --config config.yml
2025-03-27T09:44:34.330+0100    INFO    icingadb        Starting Icinga DB daemon (1.2.1-g3768602-dirty)
2025-03-27T09:44:34.330+0100    INFO    icingadb        Connecting to database at 'pgsql://postgres@localhost:5432/icingadb'
2025-03-27T09:44:34.331+0100    WARN    database        Can't connect to database. Retrying     {"error": "dial tcp [::1]:5432: connect: connection refused"}
[ . . . ]
```

</p>
</details>

Fixes #896.

### DIFF
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -67,7 +67,18 @@ func run() int {
 		}
 	}
 
-	if err := icingadb.CheckSchema(context.Background(), db); err != nil {
+	switch err := icingadb.CheckSchema(context.Background(), db); {
+	case errors.Is(err, icingadb.ErrSchemaNotExists):
+		if !cmd.Flags.DatabaseAutoImport {
+			logger.Fatal("The database schema is missing")
+		}
+
+		logger.Info("Starting database schema auto import")
+		if err := icingadb.ImportSchema(context.Background(), db, cmd.Flags.DatabaseSchemaDir); err != nil {
+			logger.Fatalf("%+v", errors.Wrap(err, "can't import database schema"))
+		}
+		logger.Info("The database schema was successfully imported")
+	case err != nil:
 		logger.Fatalf("%+v", err)
 	}
 
--- a/pkg/icingadb/schema.go
+++ b/pkg/icingadb/schema.go
@@ -2,11 +2,15 @@ package icingadb
 
 import (
 	"context"
+	stderrors "errors"
 	"fmt"
 	"github.com/icinga/icinga-go-library/backoff"
 	"github.com/icinga/icinga-go-library/database"
 	"github.com/icinga/icinga-go-library/retry"
+	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
+	"os"
+	"path"
 	"time"
 )
 
@@ -15,7 +19,19 @@ const (
 	expectedPostgresSchemaVersion = 4
 )
 
-// CheckSchema asserts the database schema of the expected version being present.
+// ErrSchemaNotExists implies that no Icinga DB schema has been imported.
+var ErrSchemaNotExists = stderrors.New("no database schema exists")
+
+// ErrSchemaMismatch implies an unexpected schema version, most likely after Icinga DB was updated but the database
+// missed the schema upgrade.
+var ErrSchemaMismatch = stderrors.New("unexpected database schema version")
+
+// CheckSchema verifies the correct database schema is present.
+//
+// This function returns the following error types, possibly wrapped:
+//   - If no schema exists, the error returned is ErrSchemaNotExists.
+//   - If the schema version does not match the expected version, the error returned is ErrSchemaMismatch.
+//   - Otherwise, the original error is returned, for example in case of general database problems.
 func CheckSchema(ctx context.Context, db *database.DB) error {
 	var expectedDbSchemaVersion uint16
 	switch db.DriverName() {
@@ -23,19 +39,25 @@ func CheckSchema(ctx context.Context, db *database.DB) error {
 		expectedDbSchemaVersion = expectedMysqlSchemaVersion
 	case database.PostgreSQL:
 		expectedDbSchemaVersion = expectedPostgresSchemaVersion
+	default:
+		return errors.Errorf("unsupported database driver %q", db.DriverName())
+	}
+
+	if hasSchemaTable, err := db.HasTable(ctx, "icingadb_schema"); err != nil {
+		return errors.Wrap(err, "can't verify existence of database schema table")
+	} else if !hasSchemaTable {
+		return ErrSchemaNotExists
 	}
 
 	var version uint16
-
 	err := retry.WithBackoff(
 		ctx,
-		func(ctx context.Context) (err error) {
+		func(ctx context.Context) error {
 			query := "SELECT version FROM icingadb_schema ORDER BY id DESC LIMIT 1"
-			err = db.QueryRowxContext(ctx, query).Scan(&version)
-			if err != nil {
-				err = database.CantPerformQuery(err, query)
+			if err := db.QueryRowxContext(ctx, query).Scan(&version); err != nil {
+				return database.CantPerformQuery(err, query)
 			}
-			return
+			return nil
 		},
 		retry.Retryable,
 		backoff.NewExponentialWithJitter(128*time.Millisecond, 1*time.Minute),
@@ -48,11 +70,50 @@ func CheckSchema(ctx context.Context, db *database.DB) error {
 		// Since these error messages are trivial and mostly caused by users, we don't need
 		// to print a stack trace here. However, since errors.Errorf() does this automatically,
 		// we need to use fmt instead.
-		return fmt.Errorf(
-			"unexpected database schema version: v%d (expected v%d), please make sure you have applied all database"+
-				" migrations after upgrading Icinga DB", version, expectedDbSchemaVersion,
+		return fmt.Errorf("%w: v%d (expected v%d), please make sure you have applied all database"+
+			" migrations after upgrading Icinga DB", ErrSchemaMismatch, version, expectedDbSchemaVersion,
 		)
 	}
 
 	return nil
+}
+
+// ImportSchema performs an initial schema import in the db.
+//
+// This function assumes that no schema exists. So it should only be called after a prior CheckSchema call.
+func ImportSchema(
+	ctx context.Context,
+	db *database.DB,
+	databaseSchemaDir string,
+) error {
+	var schemaFileDirPart string
+	switch db.DriverName() {
+	case database.MySQL:
+		schemaFileDirPart = "mysql"
+	case database.PostgreSQL:
+		schemaFileDirPart = "pgsql"
+	default:
+		return errors.Errorf("unsupported database driver %q", db.DriverName())
+	}
+
+	schemaFile := path.Join(databaseSchemaDir, schemaFileDirPart, "schema.sql")
+	schema, err := os.ReadFile(schemaFile) // #nosec G304 -- path is constructed from "trusted" command line user input
+	if err != nil {
+		return errors.Wrapf(err, "can't open schema file %q", schemaFile)
+	}
+
+	queries := []string{string(schema)}
+	if db.DriverName() == database.MySQL {
+		// MySQL/MariaDB requires the schema to be imported on a statement by statement basis.
+		queries = database.MysqlSplitStatements(string(schema))
+	}
+
+	return errors.Wrapf(db.ExecTx(ctx, func(ctx context.Context, tx *sqlx.Tx) error {
+		for _, query := range queries {
+			if _, err := tx.ExecContext(ctx, query); err != nil {
+				return errors.Wrap(database.CantPerformQuery(err, query), "can't perform schema import")
+			}
+		}
+		return nil
+	}), "can't import database schema from %q", schemaFile)
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -58,6 +58,14 @@ type Flags struct {
 	// Config is the path to the config file. If not provided, it defaults to DefaultConfigPath.
 	Config string `short:"c" long:"config" description:"path to config file (default: /etc/icingadb/config.yml)"`
 	// default must be kept in sync with DefaultConfigPath.
+
+	// DatabaseAutoImport results in an initial schema check and update; mostly for containerized setups.
+	DatabaseAutoImport bool `long:"database-auto-import" description:"import database schema on startup if database is empty"`
+
+	// DatabaseSchemaDir is the root directory for schema files to be used when DatabaseAutoImport is requested.
+	//
+	// The directory structure must mimic the git repo's schema dir, containing ./mysql/schema.sql and ./pgsql/schema.sql.
+	DatabaseSchemaDir string `long:"database-schema-dir" description:"directory for --database-auto-import, expects ./{my,pg}sql/schema.sql files" default:"./schema/"`
 }
 
 // GetConfigPath retrieves the path to the configuration file.


pkg/icingadb/schema.go
```diff
@@ -56,3 +58,109 @@ func CheckSchema(ctx context.Context, db *database.DB) error {
...
+		retry.Retryable,
+		backoff.NewExponentialWithJitter(128*time.Millisecond, 1*time.Minute),
+		db.GetDefaultRetrySettings())
+	if err != nil {
+		return false, errors.Wrap(err, "can't check if database has the schema applied")
+	}
+	if schemaTablesCount > 0 {
+		// icingadb_schema table exists; schema is at least partially applied
+		return false, nil
+	}
```
lippserd, line 115: Just reading the code, `schemaTablesCount` gave me the first impression that we would try to import missing tables. I would rather use `SELECT 1` and check for `sql.ErrNoRows`, which would also not `Scan()`:

```suggestion
	err = retry.WithBackoff(
		ctx,
		func(ctx context.Context) error {
			query := tx.Rebind("SELECT 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA=? AND TABLE_NAME='icingadb_schema'")
			if err := tx.QueryRowxContext(ctx, query, tableSchema).Err(); err != nil {
				// The `SELECT 1` query returns `sql.ErrNoRows` if the `icingadb_schema` table does not exist;
				// otherwise, it returns a database error.
				return database.CantPerformQuery(err, query)
			}

			return nil
		},
		retry.Retryable,
		backoff.NewExponentialWithJitter(128*time.Millisecond, 1*time.Minute),
		db.GetDefaultRetrySettings())
	if err == nil {
		// Table `icingadb_schema` exists.
		return false, nil
	}
	if !errors.Is(err, sql.ErrNoRows) {
		return false, errors.Wrap(err, "can't check if database has the schema applied")
	}
```

pkg/icingadb/schema.go
```diff
@@ -56,3 +58,109 @@ func CheckSchema(ctx context.Context, db *database.DB) error {
...
+		for _, query := range database.MysqlSplitStatements(string(schema)) {
+			err := retry.WithBackoff(
+				ctx,
+				func(ctx context.Context) (err error) {
+					_, err = tx.ExecContext(ctx, query)
+					if err != nil {
+						err = database.CantPerformQuery(err, query)
+					}
+					return
+				},
```
lippserd, line 134: I would always prefer the `if err := ...; err != nil { return err } ... return nil` pattern:

```suggestion
				func(ctx context.Context) error {
					if _, err := tx.ExecContext(ctx, query); err != nil {
						return database.CantPerformQuery(err, query) 
					}
					return nil
				},
```

pkg/icingadb/schema.go
```diff
@@ -56,3 +58,109 @@ func CheckSchema(ctx context.Context, db *database.DB) error {
...
+			}
+		}
+
+	case database.PostgreSQL:
+		err := retry.WithBackoff(
+			ctx,
+			func(ctx context.Context) (err error) {
+				_, err = tx.ExecContext(ctx, string(schema))
+				return
+			},
```
lippserd, line 149: Same here:

```suggestion
			func(ctx context.Context) error {
				if _, err := tx.ExecContext(ctx, string(schema)); err != nil {
					return err
				}
				return nil
			},
```

pkg/icingadb/schema.go
```diff
@@ -56,3 +58,109 @@ func CheckSchema(ctx context.Context, db *database.DB) error {
...
+
+	default:
+		return false, fmt.Errorf("unsupported database driver %q", db.DriverName())
+	}
+
+	err = tx.Commit()
+	if err != nil {
+		return false, errors.Wrap(err, "can't commit transaction")
+	}
+	return true, nil
```
lippserd, line 165: And here:

```suggestion
	if err := tx.Commit(); err != nil {
		return false, errors.Wrap(err, "can't commit transaction")
	}
	return true, nil
```

lippserd (changes requested):
In addition to the changes requested I want to propose/discuss two additional changes:

1. Embed schema files to reduce code related to flag and file handling. Also, this ensures functionality across all contexts without needing to specify the correct schema path.

2. Introduce something like `ErrNoSchema` and `ErrSchemaMismatch`, and modify `CheckSchema()` accordingly. Build error handling and schema import around these changes. This would enhance code flow and readability, as we now call `ImportSchema()` followed by `CheckSchema()`.

pkg/icingadb/schema.go
oxzi: I have tried to keep the new function consistent with the other `CheckSchema` function. So I have changed it there as well :)

julianbrost:
> * Embed schema files to reduce code related to flag and file handling. Also, this ensures functionality across all contexts without needing to specify the correct schema path.

I though about that as well while loosely following this PR. How would this work for package installs? Don't install as file? Both embed and install as a file? Introduce some build flags to control embedding?

Like when I was thinking about this, I wasn't sure if this would even simplify things in the big picture (maybe unless the schema is shipped twice, once for manual use in a file, once embedded for automated use).

lippserd:
> > * Embed schema files to reduce code related to flag and file handling. Also, this ensures functionality across all contexts without needing to specify the correct schema path.
> 
> I though about that as well while loosely following this PR. How would this work for package installs? Don't install as file? Both embed and install as a file? Introduce some build flags to control embedding?
> 
> Like when I was thinking about this, I wasn't sure if this would even simplify things in the big picture (maybe unless the schema is shipped twice, once for manual use in a file, once embedded for automated use).

I would ship the schema twice and not change anything in the documentation and package installation, as this currently only serves the purpose of moving our container file to this repo. As you said, the embedded file is for automated use and the file on disk is for package/manual use.

pkg/icingadb/schema.go
oxzi: > I would rather use `SELECT 1` and check for `sql.ErrNoRows`, which would also not require `Scan()`:

This does not work on PostgreSQL (haven't tested MySQL yet). While the query in question will return no rows, the `.Err()` method does return a `nil` error.

```
icingadb=# SELECT 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA='public' AND TABLE_NAME='icingadb_schema';
 ?column?
----------
(0 rows)
```


```go
			query := db.Rebind("SELECT 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA=? AND TABLE_NAME='icingadb_schema'")
			if err := db.QueryRowxContext(ctx, query, tableSchema).Err(); err != nil {
				// The "SELECT 1" query returns sql.ErrNoRows if the icingadb_schema table does not exist;
				// otherwise, it returns a database error.
				return database.CantPerformQuery(err, query)
			}
			return nil // <- We will end up here; verified in the debugger.
```

_(For the record, don't click "Unresolve conversation" before sending the comment, as GitHub will clean out the comment field and purge everything you have written so far. Thanks.)_

oxzi:
> In addition to the changes requested I want to propose/discuss two additional changes:
> 
>     1. Embed schema files to reduce code related to flag and file handling. Also, this ensures functionality across all contexts without needing to specify the correct schema path.

At the moment, I am against embedding the schema files because the container still needs the schema update files. So embedding the schema itself while still having to ship the schema updates is an inconsistency, also for the user.

If at some point we do have automatic schema updates, then I would argue in favor of embedding everything in the binary, but until we get there, I would leave it the way it is.

>     2. Introduce something like `ErrNoSchema` and `ErrSchemaMismatch`, and modify `CheckSchema()` accordingly. Build error handling and schema import around these changes. This would enhance code flow and readability, as we now call `ImportSchema()` followed by `CheckSchema()`.

I liked this idea and changed the PR in its favor. The initial check logic (schema exists, schema has expected version) went in `CheckSchema`, while `ImportSchema` now only does the import based on the `CheckSchema`'s error.

So I have updated the commit message and the initial post within this PR thread. I have also included two runs with hopefully every possible situations for MySQL/MariaDB and PostgreSQL in the OP.

Furthermore, `gofmt` was now unhappy about `./cmd/icingadb-migrate/main.go`, which was fixed as well.

lippserd:
> At the moment, I am against embedding the schema files because the container still needs the schema update files. So embedding the schema itself while still having to ship the schema updates is an inconsistency, also for the user.

Our container doesn't ship schema upgrade files, and I don't think it should unless we have support for automatic schema migrations (which would then be embedded as well anyway). I think you were thinking of the schema upgrade situation in container setups, but this is something where we leave our users alone at the moment as there is no documentation, which of course is not good. But that should rather be a place in the documentation where the schema upgrade files can be downloaded, otherwise you would have to document the path in the container and then users would have to copy the scripts out of the container.

oxzi:
> > At the moment, I am against embedding the schema files because the container still needs the schema update files. So embedding the schema itself while still having to ship the schema updates is an inconsistency, also for the user.
> 
> Our container doesn't ship schema upgrade files, and I don't think it should unless we have support for automatic schema migrations (which would then be embedded as well anyway).

You are right. I assumed that our container image would contain the schema upgrades as well, but in its current form it only [contains a zipped version of the schema](https://github.com/Icinga/docker-icingadb/blob/a362254223b65f9c74e96395710b6d346c63c368/Dockerfile#L42-L43).

However, for the moment I would still keep it as it is. When eventually adding automatic schema updates, I would argue for embedding the schema file and schema upgrades into the binary, adding Docker (or container) documentation and removing all SQL files from our package builds as well.

pkg/icingadb/schema.go
```diff
@@ -15,27 +17,74 @@ const (
...
+		db.GetDefaultRetrySettings())
+	if err != nil {
+		return errors.Wrap(err, "can't verify existence of database schema table")
+	}
+	if tableSchemaCount == 0 {
+		return ErrSchemaNotExists
+	}
+	if tableSchemaCount != 1 {
+		return fmt.Errorf("no or one 'icingadb_schema' tables are expected, found %d", tableSchemaCount)
+	}
```
lippserd, line 77: I am still not happy with `tableSchemaCount`. There can't be anything other than `0` or `1` here. We have the same functionality in our Kubernetes daemon:

https://github.com/Icinga/icinga-kubernetes/blob/main/cmd/icinga-kubernetes/main.go#L590

I suggest moving this to `DB#HasTable()`, with the adjustment to handle both PostgreSQL and MySQL.

pkg/icingadb/schema.go
```diff
@@ -15,27 +17,74 @@ const (
...
+	default:
+		return fmt.Errorf("unsupported database driver %q", db.DriverName())
+	}
 
+	var tableSchemaCount int
 	err := retry.WithBackoff(
 		ctx,
-		func(ctx context.Context) (err error) {
+		func(ctx context.Context) error {
+			// The following query should return either 0 or 1 depending on the existence of the icingadb_schema table.
```
lippserd, line 57: The `SELECT 1` thing was misleading on my part. I somehow thought that `Scan()` was not required, but of course it is, so you can remove the comment.

pkg/icingadb/schema.go
```diff
@@ -15,27 +17,74 @@ const (
 	expectedPostgresSchemaVersion = 4
 )
 
-// CheckSchema asserts the database schema of the expected version being present.
-func CheckSchema(ctx context.Context, db *database.DB) error {
-	var expectedDbSchemaVersion uint16
+var (
```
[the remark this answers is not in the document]
oxzi: > waste so many spaces

It is one `\t` and I like the logical grouping. But never mind.

pkg/icingadb/schema.go
```diff
@@ -48,11 +69,85 @@ func CheckSchema(ctx context.Context, db *database.DB) error {
...
+				retry.Retryable,
+				backoff.NewExponentialWithJitter(128*time.Millisecond, 1*time.Minute),
+				db.GetDefaultRetrySettings())
+			if err != nil {
+				return errors.Wrap(err, "can't perform schema import")
+			}
+		}
+
+	case database.PostgreSQL:
+		err := retry.WithBackoff(
```
lippserd, line 130: While the retry mechanism might work for MySQL/MariaDB since we split the statements, I doubt this will work for PostgreSQL if the connection breaks while importing the schema. Just wanted to note this in case it might be relevant in the future.

pkg/icingadb/schema.go
```diff
@@ -48,11 +69,85 @@ func CheckSchema(ctx context.Context, db *database.DB) error {
...
+				}
+				return nil
+			},
+			retry.Retryable,
+			backoff.NewExponentialWithJitter(128*time.Millisecond, 1*time.Minute),
+			db.GetDefaultRetrySettings())
+		if err != nil {
+			return errors.Wrap(err, "can't perform schema import")
+		}
+
```
lippserd, line 144: All other switch statements introduced have no line break between the cases. I opt to remove them here as well.

lippserd (changes requested):
Also, please use `errors.Errorf()` instead of `fmt.Errorf()`.

pkg/icingadb/schema.go
oxzi: I reworked the import schema statement execution and have removed the retry logic. Since the transaction is already started, the database must be available. If the schema fails, it will not magically recover.

cmd/icingadb/main.go
```diff
@@ -67,7 +67,19 @@ func run() int {
...
+		if !cmd.Flags.DatabaseAutoImport {
+			logger.Fatal("The database schema is missing")
+		}
+
+		logger.Info("Starting database schema auto import")
+		if err := icingadb.ImportSchema(context.Background(), db, cmd.Flags.DatabaseSchemaDir); err != nil {
+			logger.Fatalf("%+v", errors.Wrap(err, "can't import database schema"))
+		}
+		logger.Info("The database schema was successfully imported")
+
```
lippserd, line 81: Please remove the newline :).

pkg/icingadb/schema.go
```diff
@@ -15,27 +18,45 @@ const (
...
+// ErrSchemaMismatch implies an unexpected schema version, most likely after Icinga DB was updated but the database
+// missed the schema upgrade.
+var ErrSchemaMismatch = stderrors.New("unexpected database schema version")
+
+// CheckSchema verifies the correct database schema is present.
+//
+// This function returns the following error types, possibly wrapped:
+//   - If no schema exists, the error returned is ErrSchemaNotExists.
+//   - If the schema version does not match the expected version, the error returned is ErrSchemaMismatch.
+//   - Otherwise, the original error is returned, e.g., for general database problems.
```
lippserd, line 33: I don't think the comma after e.g. is necessary.

pkg/icingadb/schema.go
```diff
@@ -48,11 +69,57 @@ func CheckSchema(ctx context.Context, db *database.DB) error {
...
+	switch db.DriverName() {
+	case database.MySQL:
+		schemaFileDirPart = "mysql"
+	case database.PostgreSQL:
+		schemaFileDirPart = "pgsql"
+	default:
+		return errors.Errorf("unsupported database driver %q", db.DriverName())
+	}
+
+	tx, err := db.BeginTxx(ctx, nil)
```
lippserd, line 98: I would start the transaction after reading the schema file.

lippserd (changes requested):
There are just a few more nitpicky details to address, but otherwise it looks excellent.

pkg/icingadb/schema.go
yhabteab: There’s also already a helper function `DB#ExecTx()` for such cases 🤷‍♂️!

pkg/icingadb/schema.go
oxzi: Usually a comma follows after "e.g.", but not "i.e.".

https://english.stackexchange.com/questions/16172/should-i-always-use-a-comma-after-e-g-or-i-e

lippserd (approved):

PR #913, opened 2025-03-28
Downtimes: mark their histories as cancelled when removed from conf file

### LINKED ISSUE #910: Downtime on a removed object are never closed.
## Describe the bug

If a object with a Downtime gets disabled (even just temporary) the end of the associated Downtime is never written out to the IDO / IcingaDB. 

## To Reproduce

1. Create a host in the director and deploy it.
2. Create a downtime on the host
3. Use the director to roll back to an older version
4.Redeploy the new version

## Expected behavior

I would expect the Downtime to be terminated once the object is deactivated (The actual_end_time set to the current time). But since the downtime is dropped without ever setting this field, the object looks in the reports as if it where in a constant downtime. which does not correspond to the internal state of icinga2. 

## Screenshots

image

yhabteab:
This PR tries to work around a problem we currently have, as described in #910. That is when removing a `Checkable` object that has some downtimes set, or even just the downtime configuration, for whatever reason, manually from the conf files or through the [`/v1/config/packages`](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#create-a-config-package) API endpoint as the Icinga Director does, Icinga 2 will not be able to close the downtimes properly before removing their corresponding configuration. In such situations, you will end up with a broken histories and even corrupt the SLA results for a given checkable.

Since Icinga DB knows when the actual downtime configuration gets deleted with each initial config dump, this PR now makes use of the fact and tries to fake the corresponding removed event for these downtimes. As of now, only the `downtime_hostory` table is updated and the `cancel_time`, `has_been_cancelled` and `cancelled_by` fields of this table are updated, but we also want to create some kind of event for the regular `history` and `sla_downtime_history` tables.

TBD:
- Should we just insert a fake `downtime_end` event into the aforementioned two history tables or something else to indicate that the downtime no longer exists. @julianbrost suggested generating a completely new event type like `checkable_deleted` instead of faking the end event ourselves, but since this problem also occurs without even deleting the checkable itself, I'm not sure if this event would reflect what's happening. However, in cases where a checkable has been deleted along with it's downtimes, it could be an alternative and would somehow serve as a _reset to factory settings_.

fixes #910

### DIFF
--- a/pkg/icingadb/v1/history/downtime.go
+++ b/pkg/icingadb/v1/history/downtime.go
@@ -1,10 +1,13 @@
 package history
 
 import (
+	"context"
 	"database/sql/driver"
 	"github.com/icinga/icinga-go-library/database"
 	"github.com/icinga/icinga-go-library/types"
 	"github.com/icinga/icingadb/pkg/contracts"
+	"golang.org/x/sync/errgroup"
+	"time"
 )
 
 type DowntimeHistoryEntity struct {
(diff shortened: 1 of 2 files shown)


pkg/icingadb/v1/history/downtime.go
```diff
@@ -147,6 +150,47 @@ func (et SlaDowntimeEndTime) Value() (driver.Value, error) {
...
+	g, ctx := errgroup.WithContext(ctx)
+	g.Go(func() error {
+		defer close(downtimes)
+
+		// Transform the downtime IDs into DowntimeHistory entities with the necessary fields
+		// and stream them into the downtimes channel.
+		for _, id := range downtimeIds {
+			// The downtimes channel is buffered, so this will never block, and we can just keep sending.
+			downtimes <- &DowntimeHistory{
+				DowntimeHistoryEntity: DowntimeHistoryEntity{DowntimeId: id.(types.Binary)},
```
oxzi:
```
pkg/icingadb/v1/history/downtime.go:185:62: type assertion must be checked (forcetypeassert)
                                DowntimeHistoryEntity: DowntimeHistoryEntity{DowntimeId: id.(types.Binary)},
```

When rebased on `main`, this will make golangci-lint unhappy.

oxzi (commented):
In general and when excluding the open _TODOs_: LGTM, thanks!

I have spent a bit of a time testing this PR in different situations and it worked as expected. Btw, I have pushed a rebased version to the [`fake-downtime-end-event-review-oxzi` branch](https://github.com/Icinga/icingadb/tree/fake-downtime-end-event-review-oxzi), also including an initial draft for an integration test, 0414c2abe28d6ade22208811d2e3f65d7401cb8b.

I would say: Let's continue with this PR by also populating the other tables.

@yhabteab: If you currently do not have the time and want to hand this PR over, I am free to continue here.

Thanks again, looks good so far!

yhabteab:
> @yhabteab: If you currently do not have the time and want to hand this PR over, I am free to continue here.

I'm working on a completely different thing right now, so you can of course take over this :).

oxzi:
@lippserd: If you want to, feel free to start reviewing this PR now.

PR #914, opened 2025-03-31
Containerfile: order layers by how likely they change over time

Al2Klimov:
This caches more layers users don't have to download.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -21,13 +21,6 @@ RUN --mount=target=. \
 
 FROM scratch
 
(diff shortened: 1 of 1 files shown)


oxzi (approved):
I would put this change under micro-optimization and would have liked to see it in the other PR. But nothing against it from my side.

lippserd:
> I would put this change under micro-optimization and would have liked to see it in the other PR. But nothing against it from my side.

I have the same feeling.

lippserd (approved):

PR #920, opened 2025-04-02
Always log certain HA and Startup Messages

### LINKED ISSUE #918: HA: Different Log Level for "Taking Over" and "Handing Over"
## Describe the bug

While inspecting some logs and thinking I found an HA error, I stumbled about the fact that "Taking Over" is logged with "info" severity while "Handing Over" is logged as "warn". Since the log level was set to "warn", the logs somehow implicated a defunct HA state.

https://github.com/Icinga/icingadb/blob/5d43cb20d937f1d88d838090b10831ca47b131ea/cmd/icingadb/main.go#L175

https://github.com/Icinga/icingadb/blob/5d43cb20d937f1d88d838090b10831ca47b131ea/cmd/icingadb/main.go#L337

## To Reproduce

N/A

## Expected behavior

That related events are logged together, even if one is "bad" and the other is "okay". The status quo hides the recovery, healing, or however one wants to name it.

## Your Environment

N/A

## Additional context

This made me cross-read logs for at least 15 minutes while starting to question essential Icinga DB constraints and the setup in question.

oxzi:
Based on the configured log level, some useful log messages will be discarded. While this works as intended, this will result in unbalanced logs - HA hand over is logged, taking over is not - or missing version strings useful for log analysis.

Fixes #918

### DIFF
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -52,7 +52,7 @@ func run() int {
 	logger := logs.GetLogger()
 	defer func() { _ = logger.Sync() }()
 
-	logger.Infof("Starting Icinga DB daemon (%s)", internal.Version.Version)
+	logger.WithOptions(logs.ForceLog()).Infof("Starting Icinga DB daemon (%s)", internal.Version.Version)
 
 	db, err := cmd.Database(logs.GetChildLogger("database"))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/icingadb/main.go
```diff
@@ -172,7 +172,7 @@ func run() int {
 		for hactx.Err() == nil {
 			select {
 			case takeoverReason := <-ha.Takeover():
-				logger.Infow("Taking over", zap.String("reason", takeoverReason))
+				logger.WithOptions(logs.ForceLog()).Infow("Taking over", zap.String("reason", takeoverReason))
```
oxzi, line 175: On second thought, I would rather create both messages with the warning severity. If someone wants to run Icinga DB with error as the default log level, they may really not be interested in anything but the worst.

So I would use the force log features only for the startup message.

What do you think, @lippserd?

oxzi: Okay :)

lippserd (approved):

PR #915, opened 2025-04-01
`container-image` workflow: Tag `latest` only on release

lippserd:
`latest` is currently tagged for all pushes to the `main` branch:

https://hub.docker.com/r/icinga/icingadb/tags

PR #923, opened 2025-04-03
`container-image` workflow: Tag latest/major version only when applicable

lippserd:
To ensure the workflow is reusable across our repositories, it must support releasing from different release trains, such as when issuing bugfix or security releases for older versions alongside newer ones. This change introduces custom logic to 
 determine when to apply the latest tag and major version tags. The docker/metadata-action unconditionally applies these tags, which could override existing greater tags. The new logic compares the current release tag against all existing tags using semantic versioning rules, a feature not inherently provided by docker/metadata-action.

You can check the changes in this repo:

https://github.com/lippserd/container-image-workflow-test/pkgs/container/container-image-workflow-test

Supersedes: #915

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -27,6 +27,14 @@ env:
   # Since our repository names are already lowercase, no additional modifications are necessary.
   IMAGE_NAME: icinga/${{ github.event.repository.name }}
 
+  # The LATEST variable determines if the current release tag is the greatest tag overall.
+  # If true, the container image will be tagged as 'latest' when pushed to the container registries.
+  LATEST: false
+
+  # The LATEST_MAJOR variable determines if the current release tag is the greatest within its major version.
+  # If true, the container image will be tagged with the major version (e.g., '1') when pushed to the registries.
+  LATEST_MAJOR: false
+
 jobs:
   build-and-publish-container-image:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/container-image.yml
```diff
@@ -38,6 +46,49 @@ jobs:
...
+      - name: Prepare metadata (release tags)
+        if: github.event_name == 'release' && github.event.action == 'published'
+        run: |
+          # Retrieve the greatest existing tag in the repository by sorting tags in descending order.
+          # Options used:
+          # * --sort=-v:refname sorts tags as versions, placing the highest version at the top.
+          # * -c 'versionsort.suffix=-' ensures that pre-release tags (e.g., 1.0.0-rc1) are sorted correctly,
+          #   so they are not considered greater than their corresponding final release (e.g., 1.0.0).
+          # Intentionally not using head -1 to prevent potential broken pipe errors.
+          greatest_tag=$(git -c 'versionsort.suffix=-' tag --list --sort=-v:refname | awk 'NR<=1')
```
oxzi, line 69: Even though I have written some awk, I failed to parse this at first glance. If you want to use awk for the mentioned reasons, why not say `NR==1`? Btw, there are different awk implementations and I am not sure if some will not optimize this query and terminate early.

.github/workflows/container-image.yml
```diff
@@ -38,6 +46,49 @@ jobs:
...
+
+          if [ "${{ github.ref_name }}" = "$greatest_tag" ]; then
+            echo "The current tag ${{ github.ref_name }} is the greatest overall. Tagging as 'latest'."
+
+            # Update environment variable to enable tagging as 'latest'.
+            echo "LATEST=true" >> $GITHUB_ENV
```
oxzi, line 75:
```suggestion
            echo "LATEST=true" >> "$GITHUB_ENV"
```

Prevents globbing or word splitting. Also recommended this way in the GitHub docs:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-environment-variable

.github/workflows/container-image.yml
```diff
@@ -38,6 +46,49 @@ jobs:
...
+          fi
+
+
+          major_version=$(echo ${{ github.ref_name }} | cut -d. -f1)
+          greatest_major=$(git -c 'versionsort.suffix=-' tag --list "${major_version}.*" --sort=-v:refname | awk 'NR<=1')
+          if [ "${{ github.ref_name }}" = "$greatest_major" ]; then
+            echo "The current tag ${{ github.ref_name }} is the greatest within its major version. Tagging with major version ${major_version#v}."
+
+            # Update environment variable to enable tagging with major version.
+            echo "LATEST_MAJOR=true" >> $GITHUB_ENV
```
oxzi, line 87:
```suggestion
            echo "LATEST_MAJOR=true" >> "$GITHUB_ENV"
```

.github/workflows/container-image.yml
lippserd: I just copied it from our GitLab actions. But yes, `NR==1` is better.

oxzi: Technically, everything is fine. I just wanted to comment on this line since it took me longer to grasp than the average shell script line should take. If we are using the same pattern somewhere else, this is okay for me.

lippserd: It's a valid point, and I've just adapted it.

.github/workflows/container-image.yml
lippserd: Done, thanks.

.github/workflows/container-image.yml
lippserd: Also done.

oxzi (approved):
LGTM, Thanks!

PR #942, opened 2025-04-14
docs/04-Upgrading.md: Database Schema Upgrades

### LINKED ISSUE #926: Provide detailed instructions for schema upgrades
According to the latest docs feedback survey, some users have difficulties to understand how schema updates have to be applied.

Providing the required CLI commands in the Upgrading docs should help.

Related docs page: https://icinga.com/docs/icinga-db/latest/doc/04-Upgrading/

oxzi:
A new section on how to actually apply the database schema upgrade was introduced. This new section should take a user's fear to mess up a schema upgrade.

Fixes #926.

### DIFF
--- a/doc/04-Upgrading.md
+++ b/doc/04-Upgrading.md
@@ -1,7 +1,44 @@
 # Upgrading Icinga DB
 
-Specific version upgrades are described below. Please note that version upgrades are incremental.
-If you are upgrading across multiple versions, make sure to follow the steps for each of them.
+Some Icinga DB upgrades require manual intervention, others do not. If you need to intervene, the release notes will
+point you to the specific upgrade section on this page.
+
+Please note that version upgrades are incremental. If you are upgrading across multiple versions, make sure to follow
+the steps for each of them.
+
+## Database Schema Upgrades
+
+Certain Icinga DB version upgrades require a database schema upgrade. If the upgrade section of the specific Icinga DB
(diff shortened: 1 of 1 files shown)


doc/04-Upgrading.md
```diff
@@ -1,7 +1,44 @@
...
+modify them for your setup and the schema upgrade you want to apply.
+
+* MySQL/MariaDB:
+  ```
+  mysql -u icingadb -p icingadb < /usr/share/icingadb/schema/mysql/upgrades/1.2.1.sql
+  ```
+* PostgreSQL:
+  ```
+  psql -U icingadb icingadb < /usr/share/icingadb/schema/pgsql/upgrades/1.2.1.sql
+  ```
```
julianbrost, line 35: I'd suggest avoiding migration files that actually exist in the examples. Otherwise, copy & pasting could lead to accidents one might want to avoid. Imagine you accidentally also copy the trailing `\n` and you're using a shell that doesn't detect this as a paste but simply executes, then it'll be better if you get a "no such file" error instead of reapplying a random old schema upgrade.

Given that 1.3.0 was released, 1.2.3 would be a good placeholder version that also won't exist in the future (or at least wouldn't get a schema upgrade should it ever come into existence).

oxzi: That's a valid concern, thanks. I have changed it to the non-existent version 1.2.3 and slightly rephrased the sentence above. Please take another look.

julianbrost (approved):

Issue #893, opened 2025-02-25
Lock wait timeout exceeded for state_history when using retention


oxzi:
While working on another issue with a larger test instance, I ran into a database locking issue after enabling the Icinga DB retention.

After adding the following `retention` block to my config, the problems started.

```yaml
retention:
  history-days: 1
  sla-days: 1
```

Shortly after restarting Icinga DB, the following error appeared, always for the `state_history` table.

```
2025-02-25T10:58:39.859Z        FATAL   icingadb        Error 1205 (HY000): Lock wait timeout exceeded; try restarting transaction
can't perform "INSERT INTO \"state_history\" (\"previous_soft_state\", \"check_attempt\", \"check_source\", \"scheduling_source\", \"environment_id\", \"state_type\", \"id\", \"previous_hard_state\", \"output\", \"max_check_attempts\", \"object_type\", \"event_time\", \"endpoint_id\", \"soft_state\", \"long_output\", \"host_id\", \"service_id\", \"hard_state\") VALUES (:previous_soft_state,:check_attempt,:check_source,:scheduling_source,:environment_id,:state_type,:id,:previous_hard_state,:output,:max_check_attempts,:object_type,:event_time,:endpoint_id,:soft_state,:long_output,:host_id,:service_id,:hard_state) ON DUPLICATE KEY UPDATE \"id\" = VALUES(\"id\")"
github.com/icinga/icinga-go-library/database.CantPerformQuery
        github.com/icinga/icinga-go-library@v0.4.0/database/utils.go:16
github.com/icinga/icinga-go-library/database.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2.1
        github.com/icinga/icinga-go-library@v0.4.0/database/db.go:535
github.com/icinga/icinga-go-library/retry.WithBackoff
        github.com/icinga/icinga-go-library@v0.4.0/retry/retry.go:65
github.com/icinga/icinga-go-library/database.(*DB).NamedBulkExec.func1.(*DB).NamedBulkExec.func1.1.2
        github.com/icinga/icinga-go-library@v0.4.0/database/db.go:530
golang.org/x/sync/errgroup.(*Group).Go.func1
        golang.org/x/sync@v0.10.0/errgroup/errgroup.go:78
runtime.goexit
        runtime/asm_amd64.s:1700
retry deadline exceeded
github.com/icinga/icinga-go-library/retry.WithBackoff
[...]

PR #961, opened 2025-05-15
Retention: Unordered deletion in CleanupStmt

oxzi:
Speed up history retention by deleting old rows in an unordered fashion.

History retention deletes all rows in the history table if their timestamp is older than the configured retention period. This deletion is performed in bulk, typically limiting each statement to 5,000 entries. After the final iteration, all older entries are deleted.

So far, the deletion queries required an order for the deletion process to start with the oldest rows. If the tables to be cleaned were huge, ordering the data would be expensive. However, since all old data will be deleted without further delay, deleting the data in order is unnecessary.

For reference, compare the two following execution plans generated by PostgreSQL for a state_history table of round about 32k entries.

```
icingadb=> EXPLAIN SELECT id FROM state_history WHERE environment_id = E'\\xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' AND event_time < 1747300387000 ORDER BY event_time LIMIT 5000;
                                                                         QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.41..662.54 rows=5000 width=30)
   ->  Index Scan using idx_state_history_env_event_time on state_history  (cost=0.41..4266.61 rows=32216 width=30)
         Index Cond: (((environment_id)::bytea = '\xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'::bytea) AND ((event_time)::bigint < '1747300387000'::bigint))
(3 rows)

icingadb=> EXPLAIN SELECT id FROM state_history WHERE environment_id = E'\\xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' AND event_time < 1747300387000 LIMIT 5000;
                                                                       QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.00..264.82 rows=5000 width=22)
   ->  Seq Scan on state_history  (cost=0.00..1706.28 rows=32216 width=22)
         Filter: (((event_time)::bigint < '1747300387000'::bigint) AND ((environment_id)::bytea = '\xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'::bytea))
(3 rows)
```

The calculated cost decreased significantly from "0.41..662.54" to "0.00..264.82", since unordered deletion do no longer requires an expensive index scan.

By the way, this is not just nitpicking, but addresses an actual problem of expensive retention queries locking tables, as reported in #893.

---

As a related second change:

While the starting of each cleanup run was logged at debug level, only runs actually deleting entries were logged. Thus, from reading the logs it was not clear when the history retention job has finished.

The code was adjusted to continue logging deleting history queries at info level, while non-effective runs are now logged at debug.

Furthermore, the loop variable was removed since it became obsolete with Go 1.22, <https://go.dev/wiki/LoopvarExperiment>.

### DIFF
--- a/pkg/icingadb/cleanup.go
+++ b/pkg/icingadb/cleanup.go
@@ -80,11 +80,11 @@ func (stmt *CleanupStmt) CleanupOlderThan(
 func (stmt *CleanupStmt) build(driverName string, limit uint64) string {
 	switch driverName {
 	case database.MySQL:
-		return fmt.Sprintf(`DELETE FROM %[1]s WHERE environment_id = :environment_id AND %[2]s < :time
-ORDER BY %[2]s LIMIT %[3]d`, stmt.Table, stmt.Column, limit)
+		return fmt.Sprintf(`DELETE FROM %[1]s WHERE environment_id = :environment_id AND %[2]s < :time LIMIT %[3]d`,
+			stmt.Table, stmt.Column, limit)
 	case database.PostgreSQL:
 		return fmt.Sprintf(`WITH rows AS (
-SELECT %[1]s FROM %[2]s WHERE environment_id = :environment_id AND %[3]s < :time ORDER BY %[3]s LIMIT %[4]d
+SELECT %[1]s FROM %[2]s WHERE environment_id = :environment_id AND %[3]s < :time LIMIT %[4]d
 )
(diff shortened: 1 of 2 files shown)


oxzi:
After further investigations, I would reclassify this PR as a micro optimization only for PostgreSQL if deleting lots of history entries, e.g., after configuring a new retention policy and restarting Icinga DB.

If someone has any objections, I am open to revert this.

PR #964, opened 2025-05-21
Drop some superfluous Redis -> DB value mappings

yhabteab:
This PR is basically the same as https://github.com/Icinga/icinga2/pull/10452 and removes the now superfluous Redis -> DB value mappings, and also makes the `{host,service}_state.is_acknowledged` column a simple `boolenum` and introduces a new `is_sticky_acknowledgement` column to indicate sticky acknowledgements.

## Requires

- https://github.com/Icinga/icinga2/pull/10452

### DIFF
--- a/tests/object_sync_test.go
+++ b/tests/object_sync_test.go
@@ -500,10 +500,10 @@ func TestObjectSync(t *testing.T) {
 							StateType               string     `db:"state_type"`
 						}
 						var row Row
-						err = db.Get(&row, `SELECT is_acknowledged, is_sticky_acknowledgement, state_type
+						err = db.Get(&row, db.Rebind(`SELECT is_acknowledged, is_sticky_acknowledgement, state_type
 							FROM host_state
 							  INNER JOIN host ON host_state.host_id = host.id
-							WHERE host.name = ?`, host.Name,
+							WHERE host.name = ?`), host.Name,
 						)
 						require.NoError(t, err, "querying host state should not fail")
 						require.Equal(t, types.Bool{Bool: isAcknowledged, Valid: true}, row.IsAcknowledged, "host should be acknowledged")


schema/pgsql/upgrades/1.4.0.sql
oxzi, line 1: Please note that applying this schema upgrade results in a different schema as importing the `schema.sql` file would.

In particular, the position of the columns differs. This is a bit of a downer in PostgreSQL since you cannot add columns `AFTER` as it is possible in MySQL.

<details><summary>Click here for a frightening large diff</summary>
<p>

```diff
--- pr964-schema-import.sql     2025-05-21 16:08:36.096827348 +0200
+++ pr964-schema-upgrade.sql    2025-05-21 16:07:08.413505435 +0200
@@ -1817,7 +1817,6 @@
     check_timeout public.uint DEFAULT NULL::bigint,
     check_interval public.uint NOT NULL,
     check_retry_interval public.uint NOT NULL,
-    total_children public.uint DEFAULT NULL::bigint,
     active_checks_enabled public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
     passive_checks_enabled public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
     event_handler_enabled public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
@@ -1837,7 +1836,8 @@
     zone_name public.citext NOT NULL,
     zone_id public.bytea20 DEFAULT NULL::bytea,
     command_endpoint_name public.citext NOT NULL,
-    command_endpoint_id public.bytea20 DEFAULT NULL::bytea
+    command_endpoint_id public.bytea20 DEFAULT NULL::bytea,
+    total_children public.uint DEFAULT NULL::bigint
 );
 ALTER TABLE ONLY public.host ALTER COLUMN id SET STORAGE PLAIN;
 ALTER TABLE ONLY public.host ALTER COLUMN environment_id SET STORAGE PLAIN;
@@ -2049,11 +2049,9 @@
     is_flapping public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
     is_overdue public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
     is_acknowledged public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
-    is_sticky_acknowledgement public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
     acknowledgement_comment_id public.bytea20 DEFAULT NULL::bytea,
     last_comment_id public.bytea20 DEFAULT NULL::bytea,
     in_downtime public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
-    affects_children public.boolenum NOT NULL,
     execution_time public.uint DEFAULT NULL::bigint,
     latency public.uint DEFAULT NULL::bigint,
     check_timeout public.uint DEFAULT NULL::bigint,
@@ -2062,7 +2060,9 @@
     last_update public.biguint DEFAULT NULL::bigint,
     last_state_change public.biguint NOT NULL,
     next_check public.biguint NOT NULL,
-    next_update public.biguint NOT NULL
+    next_update public.biguint NOT NULL,
+    affects_children public.boolenum NOT NULL,
+    is_sticky_acknowledgement public.boolenum DEFAULT 'n'::public.boolenum NOT NULL
 );
 ALTER TABLE ONLY public.host_state ALTER COLUMN id SET STORAGE PLAIN;
 ALTER TABLE ONLY public.host_state ALTER COLUMN host_id SET STORAGE PLAIN;
@@ -3081,7 +3081,6 @@
     check_timeout public.uint DEFAULT NULL::bigint,
     check_interval public.uint NOT NULL,
     check_retry_interval public.uint NOT NULL,
-    total_children public.uint DEFAULT NULL::bigint,
     active_checks_enabled public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
     passive_checks_enabled public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
     event_handler_enabled public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
@@ -3101,7 +3100,8 @@
     zone_name public.citext NOT NULL,
     zone_id public.bytea20 DEFAULT NULL::bytea,
     command_endpoint_name public.citext NOT NULL,
-    command_endpoint_id public.bytea20 DEFAULT NULL::bytea
+    command_endpoint_id public.bytea20 DEFAULT NULL::bytea,
+    total_children public.uint DEFAULT NULL::bigint
 );
 ALTER TABLE ONLY public.service ALTER COLUMN id SET STORAGE PLAIN;
 ALTER TABLE ONLY public.service ALTER COLUMN environment_id SET STORAGE PLAIN;
@@ -3320,11 +3320,9 @@
     is_flapping public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
     is_overdue public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
     is_acknowledged public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
-    is_sticky_acknowledgement public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
     acknowledgement_comment_id public.bytea20 DEFAULT NULL::bytea,
     last_comment_id public.bytea20 DEFAULT NULL::bytea,
     in_downtime public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
-    affects_children public.boolenum NOT NULL,
     execution_time public.uint DEFAULT NULL::bigint,
     latency public.uint DEFAULT NULL::bigint,
     check_timeout public.uint DEFAULT NULL::bigint,
@@ -3333,7 +3331,9 @@
     last_update public.biguint DEFAULT NULL::bigint,
     last_state_change public.biguint NOT NULL,
     next_check public.biguint NOT NULL,
-    next_update public.biguint NOT NULL
+    next_update public.biguint NOT NULL,
+    is_sticky_acknowledgement public.boolenum DEFAULT 'n'::public.boolenum NOT NULL,
+    affects_children public.boolenum NOT NULL
 );
 ALTER TABLE ONLY public.service_state ALTER COLUMN id SET STORAGE PLAIN;
 ALTER TABLE ONLY public.service_state ALTER COLUMN host_id SET STORAGE PLAIN;
@@ -4569,7 +4569,7 @@
 -- Data for Name: host; Type: TABLE DATA; Schema: public; Owner: postgres
 --

-COPY public.host (id, environment_id, name_checksum, properties_checksum, name, name_ci, display_name, address, address6, address_bin, address6_bin, checkcommand
_name, checkcommand_id, max_check_attempts, check_timeperiod_name, check_timeperiod_id, check_timeout, check_interval, check_retry_interval, total_children, activ
e_checks_enabled, passive_checks_enabled, event_handler_enabled, notifications_enabled, flapping_enabled, flapping_threshold_low, flapping_threshold_high, perfdat
a_enabled, eventcommand_name, eventcommand_id, is_volatile, action_url_id, notes_url_id, notes, icon_image_id, icon_image_alt, zone_name, zone_id, command_endpoin
t_name, command_endpoint_id) FROM stdin;
+COPY public.host (id, environment_id, name_checksum, properties_checksum, name, name_ci, display_name, address, address6, address_bin, address6_bin, checkcommand
_name, checkcommand_id, max_check_attempts, check_timeperiod_name, check_timeperiod_id, check_timeout, check_interval, check_retry_interval, active_checks_enabled
, passive_checks_enabled, event_handler_enabled, notifications_enabled, flapping_enabled, flapping_threshold_low, flapping_threshold_high, perfdata_enabled, event
command_name, eventcommand_id, is_volatile, action_url_id, notes_url_id, notes, icon_image_id, icon_image_alt, zone_name, zone_id, command_endpoint_name, command_
endpoint_id, total_children) FROM stdin;
 \.

@@ -4585,7 +4585,7 @@
 -- Data for Name: host_state; Type: TABLE DATA; Schema: public; Owner: postgres
 --

-COPY public.host_state (id, host_id, environment_id, properties_checksum, state_type, soft_state, hard_state, previous_soft_state, previous_hard_state, check_attempt, severity, output, long_output, performance_data, normalized_performance_data, check_commandline, is_problem, is_handled, is_reachable, is_flapping, is_overdue, is_acknowledged, is_sticky_acknowledgement, acknowledgement_comment_id, last_comment_id, in_downtime, affects_children, execution_time, latency, check_timeout, check_source, scheduling_source, last_update, last_state_change, next_check, next_update) FROM stdin;
+COPY public.host_state (id, host_id, environment_id, properties_checksum, state_type, soft_state, hard_state, previous_soft_state, previous_hard_state, check_attempt, severity, output, long_output, performance_data, normalized_performance_data, check_commandline, is_problem, is_handled, is_reachable, is_flapping, is_overdue, is_acknowledged, acknowledgement_comment_id, last_comment_id, in_downtime, execution_time, latency, check_timeout, check_source, scheduling_source, last_update, last_state_change, next_check, next_update, affects_children, is_sticky_acknowledgement) FROM stdin;
 \.


@@ -4626,7 +4626,8 @@
 --

 COPY public.icingadb_schema (id, version, "timestamp") FROM stdin;
-1      5       1747836507286
+1      4       1747836352502
+2      5       1747836400057
 \.


@@ -4746,7 +4747,7 @@
 -- Data for Name: service; Type: TABLE DATA; Schema: public; Owner: postgres
 --

-COPY public.service (id, environment_id, name_checksum, properties_checksum, host_id, name, name_ci, display_name, checkcommand_name, checkcommand_id, max_check_attempts, check_timeperiod_name, check_timeperiod_id, check_timeout, check_interval, check_retry_interval, total_children, active_checks_enabled, passive_checks_enabled, event_handler_enabled, notifications_enabled, flapping_enabled, flapping_threshold_low, flapping_threshold_high, perfdata_enabled, eventcommand_name, eventcommand_id, is_volatile, action_url_id, notes_url_id, notes, icon_image_id, icon_image_alt, zone_name, zone_id, command_endpoint_name, command_endpoint_id) FROM
stdin;
+COPY public.service (id, environment_id, name_checksum, properties_checksum, host_id, name, name_ci, display_name, checkcommand_name, checkcommand_id, max_check_attempts, check_timeperiod_name, check_timeperiod_id, check_timeout, check_interval, check_retry_interval, active_checks_enabled, passive_checks_enabled, event_handler_enabled, notifications_enabled, flapping_enabled, flapping_threshold_low, flapping_threshold_high, perfdata_enabled, eventcommand_name, eventcommand_id, is_volatile, action_url_id, notes_url_id, notes, icon_image_id, icon_image_alt, zone_name, zone_id, command_endpoint_name, command_endpoint_id, total_children) FROM
stdin;
 \.


@@ -4762,7 +4763,7 @@
 -- Data for Name: service_state; Type: TABLE DATA; Schema: public; Owner: postgres
 --

-COPY public.service_state (id, host_id, service_id, environment_id, properties_checksum, state_type, soft_state, hard_state, previous_soft_state, previous_hard_state, check_attempt, severity, output, long_output, performance_data, normalized_performance_data, check_commandline, is_problem, is_handled, is_reachable, is_flapping, is_overdue, is_acknowledged, is_sticky_acknowledgement, acknowledgement_comment_id, last_comment_id, in_downtime, affects_children, execution_time, latency, check_timeout, check_source, scheduling_source, last_update, last_state_change, next_check, next_update) FROM stdin;
+COPY public.service_state (id, host_id, service_id, environment_id, properties_checksum, state_type, soft_state, hard_state, previous_soft_state, previous_hard_state, check_attempt, severity, output, long_output, performance_data, normalized_performance_data, check_commandline, is_problem, is_handled, is_reachable, is_flapping, is_overdue, is_acknowledged, acknowledgement_comment_id, last_comment_id, in_downtime, execution_time, latency, check_timeout, check_source, scheduling_source, last_update, last_state_change, next_check, next_update, is_sticky_acknowledgement, affects_children) FROM stdin;
 \.


@@ -4914,7 +4915,7 @@
 -- Name: icingadb_schema_id_seq; Type: SEQUENCE SET; Schema: public; Owner: postgres
 --

-SELECT pg_catalog.setval('public.icingadb_schema_id_seq', 1, true);
+SELECT pg_catalog.setval('public.icingadb_schema_id_seq', 2, true);


 --
```

</p>
</details> 

Since our queries are (hopefully) all auto-generated with an explicit column list, this should be no problem. However, having a structural change between a newly imported and an updated schema feels dangerous.

cmd/icingadb-migrate/convert.go
```diff
@@ -701,30 +699,56 @@ func convertNotificationRows(
 //
 // [1]: https://github.com/Icinga/icinga2/blob/32c7f7730db154ba0dff5856a8985d125791c/lib/db_ido/dbevents.cpp#L1507-L1524
 // [2]: https://github.com/Icinga/icingadb/blob/8f31ac143875498797725adb9bfacf3d4/pkg/types/notification_type.go#L53-L61
-func convertNotificationType(notificationReason, state uint8) icingadbTypes.NotificationType {
+func convertNotificationType(notificationReason, state uint8) (string, error) {
```
oxzi, line 702: The function documentation is no longer accurate, especially the second part about the removed numeric representation in Icinga DB.

cmd/icingadb-migrate/convert.go
```diff
@@ -701,30 +699,56 @@ func convertNotificationRows(
...
+	case 8:
+		return "custom", nil
+	default:
+		return "", fmt.Errorf("bad notification type: %#v", notificationReason)
+	}
+}
+
+// convertCommentType maps IDO comment_type values to either "comment" or "ack" accordingly.
+// The rule of thumb is: 1 = comment, 4 = ack, if anything else is provided, this function panics.
+func convertCommentType(commentType uint8) string {
```
oxzi, line 731: This function is currently only called once, where `state` is always 1.

So it can be removed and "comment" can be directly set above.

https://github.com/Icinga/icingadb/blob/85b55ab65bcae3248f28badee98a1689c6505450/cmd/icingadb-migrate/convert.go#L91

cmd/icingadb-migrate/convert.go
```diff
@@ -701,30 +699,56 @@ func convertNotificationRows(
...
+	case 5:
+		return "downtime_start", nil
+	case 6:
+		return "downtime_end", nil
+	case 7:
+		return "downtime_removed", nil
+	case 8:
+		return "custom", nil
+	default:
+		return "", fmt.Errorf("bad notification type: %#v", notificationReason)
```
oxzi, line 725: I have just checked what happens with a non-empty error return value...

https://github.com/Icinga/icingadb/blob/85b55ab65bcae3248f28badee98a1689c6505450/cmd/icingadb-migrate/convert.go#L622-L625

Nothing. The error is silently discarded.

While this was already the case before opening this PR, could you please either log the error or cause a `panic`?

pkg/icingadb/history/sla.go
```diff
@@ -21,7 +20,7 @@ func stateHistoryToSlaEntity(entry redis.XMessage) ([]history.UpserterEntity, er
 	}
 	slaState := slaStateInterface.(*history.SlaHistoryState)
 
-	if slaState.StateType != types.StateHard {
+	if slaState.StateType != "hard" {
```
oxzi, line 23: I am a bit unhappy with the appearance of the magic strings. Before, there were `const` values to compare against, but now there are "raw" strings, being prone for typos and to be forgotten in a future refactoring.

This is also prominent in the `cmd/icingadb-migrate/convert.go` file.

cmd/icingadb-migrate/convert.go
oxzi, line 725: Nitpick: The `%#v` format verb results in a lowercase hex representation for `uint8`. In the following two functions you have decided to use `%d` for the same type. Please keep them alike.

oxzi (commented):
First, thanks for this PR. In general, it looks good.

I went over the code - and its counterpart in Icinga 2 - multiple times and added some small comments.

Furthermore, I did some testing in a playground setup and clicking a non-sticky and a sticky in Icinga DB Web resulted in the expected data within the relation database.

```
MariaDB [icingadb]> select comment.text, is_acknowledged, is_sticky_acknowledgement  from service_state join comment on service_state.acknowledgement_comment_id = comment.id where is_acknowledged = 'y'\G
                     text: F
          is_acknowledged: y
is_sticky_acknowledgement: n
*************************** 2. row ***************************
                     text: sticky F
          is_acknowledged: y
is_sticky_acknowledgement: y
2 rows in set (0.004 sec)
```

cmd/icingadb-migrate/convert.go
yhabteab: Just copy pasted from the now removed `notification_type` code :)!

tests/history_test.go
```diff
@@ -702,14 +703,30 @@ func assertStreamConsistency(t testing.TB, clients []*redis.Client, stream strin
 	}
 }
 
-func processCheckResult(t *testing.T, client *utils.Icinga2Client, hostname string, status int) time.Time {
+// processCheckResult sends a passive check result to the Icinga2 API and returns the time of the check result.
+// The provided hostOrServiceName must be either a host name or a service name in the format "host!service".
+func processCheckResult(t *testing.T, client *utils.Icinga2Client, hostOrServiceName string, status int, isHost bool) time.Time {
```
julianbrost, line 708: The `isHost` parameter is redundant, you can just check whether the name contains `!` instead.

tests/history_test.go
```diff
@@ -702,14 +703,30 @@ func assertStreamConsistency(t testing.TB, clients []*redis.Client, stream strin
...
+	)
+
+	if isHost {
+		objType = "Host"
+		filter = fmt.Sprintf(`host.name==%q`, hostOrServiceName)
+	} else {
+		objType = "Service"
+		parts := strings.Split(hostOrServiceName, "!")
+		filter = fmt.Sprintf(`service.name==%q&&host.name==%q`, parts[1], parts[0])
+	}
```
julianbrost, line 724: If touching that part anyways: instead of using a filter expression, it's much simpler to just add `{"host":"example.com"}` or `{"service":"example.com!ping"}` to the JSON.

pkg/icingadb/v1/history/comment.go
```diff
@@ -42,13 +41,13 @@ type CommentHistory struct {
...
-	Author                 string                    `json:"author"`
-	Comment                string                    `json:"comment"`
-	EntryType              icingadbTypes.CommentType `json:"entry_type"`
-	IsPersistent           types.Bool                `json:"is_persistent"`
-	IsSticky               types.Bool                `json:"is_sticky"`
-	ExpireTime             types.UnixMilli           `json:"expire_time"`
+	EntryTime              types.UnixMilli `json:"entry_time"`
+	Author                 string          `json:"author"`
+	Comment                string          `json:"comment"`
+	EntryType              string          `json:"entry_type"`
```
julianbrost, line 47: When changing the history types, have you considered that there may still be some pending history entries present in the Redis stream written by a previous version of Icinga 2? So for the history streams, Icinga DB should ideally still be able to process the older entries.

tests/history_test.go
yhabteab: Yes, that would be enough, then you've to always make sure that you don't set the `host` and `service` JSON keys at the same time or trigger the request with one of them set to empty string. And that alone is never _simpler_ than just using a well supported filter string.

yhabteab: Done!

pkg/icingadb/v1/history/comment.go
yhabteab: `CommentEntryType` supports now both formats for `comment#entry_type`.

yhabteab: Why 😭?
```go
2025-05-30T11:33:15.678Z	FATAL	icingadb	invalid character 'c' looking for beginning of value
can't unmarshal JSON into *interface {}
github.com/icinga/icinga-go-library/types.UnmarshalJSON
	/go/pkg/mod/github.com/icinga/icinga-go-library@v0.7.0/types/utils.go:51
github.com/icinga/icingadb/pkg/icingadb/v1/history.(*CommentEntryType).UnmarshalJSON
	/icingadb/pkg/icingadb/v1/history/comment.go:125
github.com/icinga/icingadb/pkg/icingadb/v1/history.(*CommentEntryType).UnmarshalText
	/icingadb/pkg/icingadb/v1/history/comment.go:149
github.com/icinga/icinga-go-library/structify.parseString
	/go/pkg/mod/github.com/icinga/icinga-go-library@v0.7.0/structify/structify.go:111
github.com/icinga/icinga-go-library/structify.structifyMapByTree
	/go/pkg/mod/github.com/icinga/icinga-go-library@v0.7.0/structify/structify.go:83
github.com/icinga/icinga-go-library/structify.MakeMapStructifier.func1
	/go/pkg/mod/github.com/icinga/icinga-go-library@v0.7.0/structify/structify.go:39
github.com/icinga/icingadb/pkg/icingadb/history.writeOneEntityStage.func1
	/icingadb/pkg/icingadb/history/sync.go:187
github.com/icinga/icingadb/pkg/icingadb/history.writeOneEntityStage.writeMultiEntityStage.func2.1
	/icingadb/pkg/icingadb/history/sync.go:224
golang.org/x/sync/errgroup.(*Group).add.func1
	/go/pkg/mod/golang.org/x/sync@v0.14.0/errgroup/errgroup.go:130
runtime.goexit
	/usr/local/go/src/runtime/asm_arm64.s:1223
can't parse entry_type into the CommentEntryType CommentHistory#EntryType: comment
github.com/icinga/icinga-go-library/structify.structifyMapByTree
	/go/pkg/mod/github.com/icinga/icinga-go-library@v0.7.0/structify/structify.go:94
github.com/icinga/icinga-go-library/structify.MakeMapStructifier.func1
	/go/pkg/mod/github.com/icinga/icinga-go-library@v0.7.0/structify/structify.go:39
github.com/icinga/icingadb/pkg/icingadb/history.writeOneEntityStage.func1
	/icingadb/pkg/icingadb/history/sync.go:187
github.com/icinga/icingadb/pkg/icingadb/history.writeOneEntityStage.writeMultiEntityStage.func2.1
	/icingadb/pkg/icingadb/history/sync.go:224
golang.org/x/sync/errgroup.(*Group).add.func1
	/go/pkg/mod/golang.org/x/sync@v0.14.0/errgroup/errgroup.go:130
runtime.goexit
	/usr/local/go/src/runtime/asm_arm64.s:1223
```

pkg/icingadb/v1/history/comment.go
```diff
@@ -109,14 +111,58 @@ func (et CommentEventTime) Value() (driver.Value, error) {
 	}
 }
 
+// CommentEntryType represents the type of a comment in the history.
+//
+// Starting with Icinga 2 2.15, the comment entry type will always be written to Redis as a string.
+// For compatibility reasons with older history entries within Redis, it supports also uint8 values.
+type CommentEntryType struct {
+	Entry any `db:"-"`
+}
```
julianbrost, line 120: Couldn't this simply be `type CommentEntryType string`? It's set to a string in all cases below already, I'm just not sure if there's still another reason for the struct around it.

tests/history_test.go
```diff
@@ -702,17 +703,24 @@ func assertStreamConsistency(t testing.TB, clients []*redis.Client, stream strin
...
 	output := utils.RandomString(8)
-	reqBody, err := json.Marshal(ActionsProcessCheckResultRequest{
-		Type:         "Host",
-		Filter:       fmt.Sprintf(`host.name==%q`, hostname),
-		ExitStatus:   status,
-		PluginOutput: output,
-	})
+	aprr := ActionsProcessCheckResultRequest{ExitStatus: status, PluginOutput: output}
+
+	if strings.IndexRune(hostOrServiceName, '!') == -1 {
```
julianbrost, line 715: I think `strings.ContainsRune()` would be the better function to use here as that one already says in the name what is checked here.

tests/history_test.go
```diff
@@ -702,17 +703,24 @@ func assertStreamConsistency(t testing.TB, clients []*redis.Client, stream strin
...
 	time.Sleep(10 * time.Millisecond)
 
 	output := utils.RandomString(8)
-	reqBody, err := json.Marshal(ActionsProcessCheckResultRequest{
-		Type:         "Host",
-		Filter:       fmt.Sprintf(`host.name==%q`, hostname),
-		ExitStatus:   status,
-		PluginOutput: output,
-	})
+	aprr := ActionsProcessCheckResultRequest{ExitStatus: status, PluginOutput: output}
```
julianbrost, line 713: I presume `aprr` is supposed to be an abbreviation of the type name, but then it misses a `c`. However, unless it was a common abbreviation, I don't see how this provides any benefit over a name like let's say `r`. A more meaningful name could be `req` or maybe `processCR` (as we use `cr` frequently to abbreviate check result, at least in the C++ code).

tests/history_test.go
```diff
@@ -779,6 +788,8 @@ type ActionsRemoveCommentRequest struct {
 
 type ActionsProcessCheckResultRequest struct {
 	Type         string `json:"type"`
+	Host         string `json:"host,omitempty"`    // Set either Host or Service, not both.
+	Service      string `json:"service,omitempty"` // Set either Host or Service, not both.
 	Filter       string `json:"filter"`
```
julianbrost, line 793: That comment pretty much applies to `Filter` as well, you want to set exactly one of the three. So I'd visually separate them from the others by an empty line and then add a single comment above that block.

pkg/icingadb/v1/history/comment.go
julianbrost: Without having verified this myself, but this fits a suspicion I had while looking at `CommentEntryType`: you shouldn't need the `UnmarshalText` for that type. In the Redis streams, the individual attributes are already key-value pairs, i.e. there's no wrapping JSON object and string values are just there as strings, without additionally being JSON-encoded.

yhabteab: > In the Redis streams, the individual attributes are already key-value pairs, i.e. there's no wrapping JSON object and string values are just there as strings, without additionally being JSON-encoded.

Yes, that's correct but I do need an implementation for `UnmarshalText()` that doesn't call `UnmarshalJSON()`, otherwise, how is the text `comment` going to be decoded into `CommentEntryType`.

yhabteab:
That was my final push :). So please have another look again!

pkg/icingadb/v1/history/notification.go
```diff
@@ -42,10 +44,49 @@ func (*HistoryNotification) TableName() string {
...
+	t := string(text)
+	for bitset, name := range notificationTypes {
+		if t == name || t == bitset {
+			*nt = NotificationType(name)
+			return nil
+		}
+	}
+
+	return errors.Errorf("bad notification type: %#v", t)
+}
```
julianbrost, line 65: How much do we care for actually detecting invalid values here? I mean we would have been fine with this being just a plain `string` that would do no validation, this type exists only to provide compatibility.

That would then allow simplifying this a bit (and could also be used for the other two types where it would avoid spelling out each string literal twiche):

```suggestion
func (nt *NotificationType) UnmarshalText(text []byte) error {
	switch t := string(text); t {
	case "1":
		*nt = "downtime_start"
	case "2":
	// [...] (only cases for numbers)
	default:
		*nt = t
	}
	return nil
}
```

julianbrost (dismissed):

oxzi (dismissed):
LGTM.

Went over the code another time, tested it together with Icinga 2 based on your other PR, tested updated `icingadb-migrate` command. Nothing fell apart. Thanks!

PR #998, opened 2025-07-29
Initial Icinga Notifications Source

oxzi:
This is the first version to use Icinga DB as an event source for Icinga Notifications. If configured accordingly, Icinga DB forwards events crafted from the Redis pipeline to the Icinga Notifications API.

This required a small refactoring of the history synchronization to allow hooking into the Redis stream. Afterwards, the newly introduced notifications package handles the rest.

Note: As part of this architectural change, Icinga Notifications offers filters to be evaluated by Icinga DB. At the moment, these are SQL queries being executed on the Icinga DB relational database. Either consider both Icinga DB and Icinga Notifications to be part of the same trust domain or consider the security implications.

Furthermore, this change requires a change on Icinga Notifications as well. This will not work with the current version 0.1.1.

---

Other PRs are
- Icinga/icinga-notifications#324 and
- Icinga/icinga-go-library#145.

### DIFF
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -171,14 +171,14 @@ func run() int {
 
 	{
 		var extraStages map[string]history.StageFunc
-		if cfg := cmd.Config.NotificationsSource; cfg.ApiBaseUrl != "" {
+		if cfg := cmd.Config.Notifications; cfg.Url != "" {
 			logger.Info("Starting Icinga Notifications source")
 
 			notificationsSource, err := notifications.NewNotificationsClient(
 				ctx,
 				db,
 				rc,
-				logs.GetChildLogger("notifications-source"),
+				logs.GetChildLogger("notifications"),
 				cfg)
 			if err != nil {
 				logger.Fatalw("Can't create Icinga Notifications client from config", zap.Error(err))
--- a/config.example.yml
+++ b/config.example.yml
@@ -142,9 +142,9 @@ redis:
 
 # Icinga DB can act as an event source for Icinga Notifications. If the following block is not empty, Icinga DB will
 # submit events to the Icinga Notifications API.
-#notifications-source:
+#notifications:
   # URL to the API root.
-#  api-base-url: http://localhost:5680
+#  url: http://localhost:5680
 
   # Username to authenticate against the Icinga Notifications API.
 #  username: icingadb
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -174,7 +174,7 @@ ICINGADB_RETENTION_OPTIONS=comment:356
 | count        | **Optional.** Number of old historical data a single query can delete in a `"DELETE FROM ... LIMIT count"` manner. Defaults to `5000`.                                                                        |
 | options      | **Optional.** Map of history category to number of days to retain its data. Available categories are `acknowledgement`, `comment`, `downtime`, `flapping`, `notification` and `state`.                        |
 
-## Notifications Source Configuration
+## Notifications Configuration
 
 !!! warning
 
@@ -185,14 +185,14 @@ ICINGADB_RETENTION_OPTIONS=comment:356
 Icinga DB can act as an event source for [Icinga Notifications](https://icinga.com/docs/icinga-notifications/).
 If configured, Icinga DB will submit events to the Icinga Notifications API.
 
-For YAML configuration, the options are part of the `notifications-source` dictionary.
-For environment variables, each option is prefixed with `ICINGADB_NOTIFICATIONS_SOURCE_`.
+For YAML configuration, the options are part of the `notifications` dictionary.
+For environment variables, each option is prefixed with `ICINGADB_NOTIFICATIONS_`.
 
-| Option       | Description                                                                       |
-|--------------|-----------------------------------------------------------------------------------|
-| api-base-url | **Optional.** Icinga Notifications API base URL, such as `http://localhost:5680`. |
-| username     | **Optional.** Icinga Notifications API user for this source.                      |
-| password     | **Optional.** Icinga Notifications API user password.                             |
+| Option   | Description                                                                       |
+|----------|-----------------------------------------------------------------------------------|
+| url      | **Optional.** Icinga Notifications API base URL, such as `http://localhost:5680`. |
+| username | **Optional.** Icinga Notifications API user for this source.                      |
+| password | **Optional.** Icinga Notifications API user password.                             |
 
 ## Appendix
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -16,11 +16,11 @@ const DefaultConfigPath = "/etc/icingadb/config.yml"
 
 // Config defines Icinga DB config.
 type Config struct {
-	Database            database.Config `yaml:"database" envPrefix:"DATABASE_"`
-	Redis               redis.Config    `yaml:"redis" envPrefix:"REDIS_"`
-	Logging             logging.Config  `yaml:"logging" envPrefix:"LOGGING_"`
-	Retention           RetentionConfig `yaml:"retention" envPrefix:"RETENTION_"`
-	NotificationsSource source.Config   `yaml:"notifications-source" envPrefix:"NOTIFICATIONS_SOURCE_"`
+	Database      database.Config `yaml:"database" envPrefix:"DATABASE_"`
+	Redis         redis.Config    `yaml:"redis" envPrefix:"REDIS_"`
+	Logging       logging.Config  `yaml:"logging" envPrefix:"LOGGING_"`
+	Retention     RetentionConfig `yaml:"retention" envPrefix:"RETENTION_"`
+	Notifications source.Config   `yaml:"notifications" envPrefix:"NOTIFICATIONS_"`
 }
 
 func (c *Config) SetDefaults() {


yhabteab:
I've changed quite a few things since your last push, so I'll summarize the changes here for you:

- You previously used the `any` type to represent the history entries streamed from the history pipelines, but since we know they are going to be one of the `v1/history` types, I've replaced `any` with our `database.Entity` type, which is the common type for all our v1 database models. I've pushed this change directly into your commit, so you won't find extra commit for this change.
- Instead of having to maintain duplicate types of the Icinga Notifications `Event`, `Severity` etc. types, I've taken the liberty to outsource them into our Icinga Go Library [^1], so that they can be used by all Icinga Notifications sources (ours or external ones) built in Go. Additionally, I've introduced there a common Icinga Notifications client that can be used to interact with the Icinga Notifications API in a common way. Thus, some of the methods you previously implemented on the `Source` type have been moved to the new client with some minor adjustments to make them more generic and reusable. So, 92b4e4a is the commit that makes use of that new client and also removes the now obsolete types and methods from the `pkg/notifications` package.
- I know it's only for the time being, but you previously limited the query parameters to be used for the SQL queries of the event rules config `ObjectFilter` to only `{host,service,environment}_id` but since the v1/history types provide numerous other fields that can be used for filtering, I've removed this limitation and now the whole entity is passed to the sqlx query builder, so that any field can be used for filtering as long as it is supported by that concrete entity. You'll find this change in the commit 4568b93.
- Since the v1 entities don't have access to the actual Host or Service names, you've previously performed SQL queries to the `hosts` and `services` tables to get the names of the hosts and services, but since these database queries were triggered with incoming history entry, I found it a bit inefficient, so I've changed that to retrieve the names from Redis via the `HGET` command instead, which doesn't require SQL parsing and round trip to the database. I believe this will bring a significant performance improvement, especially in a big Icinga environment with numerous objects. If you don't see it that way, you can also revert it, and we'll discuss it further when I'm back. Performance-wise, there's still a room for improvement though, because when starting Icinga 2 with a lot of objects, Icinga Notifications still takes a while to actually receive a state change event triggered 1-3m after Icinga DB Web shows it in the history view (e.g. when you submit a process-check-result with OK state, the respective incident in Icinga Notifications won't be closed until 1-3m later). The commit that implements this change is b01e61c.
- Previously, when Icinga Notifications rejected the process event request with _outdated rules version_, you have posted the current event in a background in a `go s.Submit()` manner, so that it can be re-processed by the worker and re-evaluate the newly updated rules. However, since the workers input channel is buffered, this can mess up the order of the events when there are already other events waiting in the queue. So, I've changed this to directly re-evaluate the event using  a magic but simple `goto ...` statement. The commit that implements this change is 55753b5.
- Finally, I've dropped the `rulesMutex` mutex that was used to protect the `rules` map, since it is no longer needed. May be you've planned to parallelize the event submission at some point, but at the moment, there's no concurrent access to it, so I've dropped it entirely with 3549849 for the time being.

Also, all the individual commits are self-contained, and includes the relevant commit message with some details about
the changes, so you can always refer to them to see the justifications for the changes.

[^1]: https://github.com/Icinga/icinga-go-library/pull/145

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,447 @@
...
+	}
+
+	if h.HasBeenCancelled.Valid && h.HasBeenCancelled.Bool {
+		ev.Type = event.TypeDowntimeRemoved
+		ev.Message = "Downtime was cancelled"
+
+		if h.CancelledBy.Valid {
+			ev.Username = h.CancelledBy.String
+		}
+	} else if h.EndTime.Time().Compare(time.Now()) <= 0 {
```
julianbrost, line 246: Huh, is comparing with now really the best option here? Shouldn't there be a more clear indication if this is a start or end event in the history entry?

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,447 @@
...
+			}
+
+			var ev *event.Event
+			var eventErr error
+
+			// Keep the type switch in sync with syncPipelines from pkg/icingadb/history/sync.go
+			switch h := entity.(type) {
+			case *v1history.NotificationHistory:
+				// Ignore for the moment.
+				continue
```
julianbrost, line 351: For the moment? I doubt we want to handle this later as this project is supposed to replace that. (And then it also doesn't make sense to attach anything to the notification history stream.)

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,447 @@
...
+				}
+
+				ev, eventErr = s.buildStateHistoryEvent(s.ctx, h)
+
+			case *v1history.DowntimeHistory:
+				ev, eventErr = s.buildDowntimeHistoryEvent(s.ctx, h)
+
+			case *v1history.CommentHistory:
+				// Ignore for the moment.
+				continue
```
julianbrost, line 365: The comment (as in code comment here) would be more helpful if it said when this should be implemented / what is currently missing to implement it. And it should probably include the TODO keyword so that someone might actually find it later.

julianbrost (commented):
Apart from the inline comments, there's also a more general question that still needs consideration: error handling. In the sense that what happens if submitting the event to Icinga Notifications doesn't work. With the current implementation, it just discarded and thereby lost. That would be solved if the submission was synchronous in the history pipeline, just as writing the history to the SQL database currently is. In that case, the event would only be deleted from Redis after it was both written to the database and sent to Icinga Notifications (on the other hand, if Icinga Notifications is unavailable, this would probably prevent history from being written).

This would also solve a related issue currently present in the implementation: once it's written to the Go channel, it's considered done. If Icinga DB is stopped, there may be still events in flight and there seems to be nothing that would wait for them. So unless I'm missing something, a clean restart of the Icinga DB process might result in lost events.

pkg/icingadb/history/sync.go
```diff
@@ -66,6 +67,10 @@ func (s Sync) Sync(ctx context.Context) error {
 		// forward the entry after it has completed its own sync so that later stages can rely on previous stages being
 		// executed successfully.
 
+		if callback != nil {
+			pipeline = append(pipeline, makeCallbackStageFunc(callback))
```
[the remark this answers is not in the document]
oxzi: Valid point. I have changed the pipeline logic, addressing your other comment <https://github.com/Icinga/icingadb/pull/998#discussion_r2314243700>.

Now, the variable is being shadowed and only this one gets altered.

pkg/icingadb/history/sync.go
```diff
@@ -360,6 +362,64 @@ func userNotificationStage(ctx context.Context, s Sync, key string, in <-chan re
...
+			structPtr = (*v1.DowntimeHistory)(nil)
+		case "comment":
+			structPtr = (*v1.CommentHistory)(nil)
+		case "flapping":
+			structPtr = (*v1.FlappingHistory)(nil)
+		case "acknowledgement":
+			structPtr = (*v1.AcknowledgementHistory)(nil)
+		default:
+			return fmt.Errorf("unsupported key %q", key)
+		}
```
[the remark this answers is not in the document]
oxzi: I have reworked this and removed the duplicate type mapping switch case. Now there is a "consumer"-supplied map of requested pipeline keys and destination types.

So there are only callback calls for keys one is actually interested in and one is able to say what kind of type should be used. This became necessary for the DowntimeHistory/HistoryDowntime thingy.

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,447 @@
...
+// buildCommonEvent creates an event.Event based on Host and (optional) Service names.
+//
+// This function is used by all event builders to create a common event structure that includes the host and service
+// names, the absolute URL to the Icinga Web 2 Icinga DB page for the host or service, and the tags for the event.
+// Any event type-specific information (like severity, message, etc.) is added by the specific event builders.
+func (s *Source) buildCommonEvent(rlr *redisLookupResult) (*event.Event, error) {
+	var (
+		eventName string
+		eventUrl  *url.URL
+		eventTags map[string]string
```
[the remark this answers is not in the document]
oxzi: Done.

pkg/notifications/redis_fetch.go
```diff
@@ -0,0 +1,89 @@
...
+			retry.Settings{},
+		)
+		if err != nil {
+			if errors.Is(err, redis.Nil) {
+				return fmt.Errorf("%s with ID %s not found in Redis", typ, hostId)
+			}
+			return fmt.Errorf("failed to fetch %s with ID %s from Redis: %w", typ, field, err)
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
julianbrost: Here's roughly what I had in mind (compiles but untested apart from that):

```go
func (client *Client) fetchHostServiceName(ctx context.Context, hostId, serviceId types.Binary) (*redisLookupResult, error) {
	getNameFromRedis := func(typ, id string) (string, error) {
		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
		defer cancel()

		var data string
		err := retry.WithBackoff(
			ctx,
				data, err = client.redisClient.HGet(ctx, "icinga:"+typ, id).Result()
				return
			},
			retry.Retryable,
			backoff.DefaultBackoff,
			retry.Settings{},
		)
		if err != nil {
			if errors.Is(err, redis.Nil) {
				return "", fmt.Errorf("%s with ID %s not found in Redis", typ, hostId)
			}
			return "", fmt.Errorf("failed to fetch %s with ID %s from Redis: %w", typ, id, err)
		}

		var result struct {
			Name string `json:"name"`
		}

		if err := json.Unmarshal([]byte(data), &result); err != nil {
			return "", fmt.Errorf("failed to unmarshal redis result: %w", err)
		}

		return result.Name, nil
	}

	var result redisLookupResult
	var err error

	result.HostName, err = getNameFromRedis("host", hostId.String())
	if err != nil {
		return nil, err
	}

	if serviceId != nil {
		result.ServiceName, err = getNameFromRedis("service", serviceId.String())
		if err != nil {
			return nil, err
		}
	}

	return &result, nil
}

// redisLookupResult defines the structure of the Redis message we're interested in.
type redisLookupResult struct {
	HostName    string `json:"-"` // Name of the host (never empty).
	ServiceName string `json:"-"` // Name of the service (only set in service context).
}
```

<details>
<summary>Less readable diff with the same changes that can just be applied to the file</summary>

```diff
diff --git a/pkg/notifications/redis_fetch.go b/pkg/notifications/redis_fetch.go
index 3cacde1..c221400 100644
--- a/pkg/notifications/redis_fetch.go
+++ b/pkg/notifications/redis_fetch.go
@@ -24,40 +24,52 @@ import (
 // request and return an error indicating that the operation timed out. In case of the serviceId being set, the
 // maximum execution time of the Redis HGet commands is 10s (5s for each HGet call).
 func (client *Client) fetchHostServiceName(ctx context.Context, hostId, serviceId types.Binary) (*redisLookupResult, error) {
-	redisHGet := func(typ, field string, out *redisLookupResult) error {
+	getNameFromRedis := func(typ, id string) (string, error) {
 		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 		defer cancel()
 
+		var data string
 		err := retry.WithBackoff(
 			ctx,
-			func(ctx context.Context) error { return client.redisClient.HGet(ctx, "icinga:"+typ, field).Scan(out) },
+			func(ctx context.Context) (err error) {
+				data, err = client.redisClient.HGet(ctx, "icinga:"+typ, id).Result()
+				return
+			},
 			retry.Retryable,
 			backoff.DefaultBackoff,
 			retry.Settings{},
 		)
 		if err != nil {
 			if errors.Is(err, redis.Nil) {
-				return fmt.Errorf("%s with ID %s not found in Redis", typ, hostId)
+				return "", fmt.Errorf("%s with ID %s not found in Redis", typ, hostId)
 			}
-			return fmt.Errorf("failed to fetch %s with ID %s from Redis: %w", typ, field, err)
+			return "", fmt.Errorf("failed to fetch %s with ID %s from Redis: %w", typ, id, err)
 		}
-		return nil
+
+		var result struct {
+			Name string `json:"name"`
+		}
+
+		if err := json.Unmarshal([]byte(data), &result); err != nil {
+			return "", fmt.Errorf("failed to unmarshal redis result: %w", err)
+		}
+
+		return result.Name, nil
 	}
 
 	var result redisLookupResult
-	if err := redisHGet("host", hostId.String(), &result); err != nil {
+	var err error
+
+	result.HostName, err = getNameFromRedis("host", hostId.String())
+	if err != nil {
 		return nil, err
 	}
 
-	result.HostName = result.Name
-	result.Name = "" // Clear the name field for the host, as we will fetch the service name next.
-
 	if serviceId != nil {
-		if err := redisHGet("service", serviceId.String(), &result); err != nil {
+		result.ServiceName, err = getNameFromRedis("service", serviceId.String())
+		if err != nil {
 			return nil, err
 		}
-		result.ServiceName = result.Name
-		result.Name = "" // It's not needed anymore, clear it!
 	}
 
 	return &result, nil
@@ -67,23 +79,4 @@ func (client *Client) fetchHostServiceName(ctx context.Context, hostId, serviceI
 type redisLookupResult struct {
 	HostName    string `json:"-"` // Name of the host (never empty).
 	ServiceName string `json:"-"` // Name of the service (only set in service context).
-
-	// Name is used to retrieve the host or service name from Redis.
-	// It should not be used for any other purpose apart from within the [Client.fetchHostServiceName] function.
-	Name string `json:"name"`
-}
-
-// UnmarshalBinary implements the [encoding.BinaryUnmarshaler] interface for redisLookupResult.
-//
-// It unmarshals the binary data of the Redis HGet result into the redisLookupResult struct.
-// This is required for the HGet().Scan() usage in the [Client.fetchHostServiceName] function to work correctly.
-func (rlr *redisLookupResult) UnmarshalBinary(data []byte) error {
-	if len(data) == 0 {
-		return errors.New("empty data received for redisLookupResult")
-	}
-
-	if err := json.Unmarshal(data, rlr); err != nil {
-		return fmt.Errorf("failed to unmarshal redis result: %w", err)
-	}
-	return nil
 }
```
</summary>

cmd/icingadb/main.go
```diff
@@ -169,9 +171,25 @@ func run() int {
...
+
+			notificationsSource := notifications.NewNotificationsClient(
+				ctx,
+				db,
+				rc,
+				logs.GetChildLogger("notifications-source"),
+				cfg)
+			callback = notificationsSource.Submit
+			callbackKeyStructPtr = notifications.SyncKeyStructPtrs
+		}
```
julianbrost: This shouldn't be done in this background routine given that `notifications.NewNotificationsClient()` can terminate the process:

https://github.com/Icinga/icingadb/blob/ed7b0d1e0cf739fde84f5f9aedb60f654061c85c/pkg/notifications/notifications.go#L65-L68

Also, look at the number of calls to `logger.Fatal*()` in this function. So for consistency, that function shouldn't even terminate the process by itself but leave the job to `main()`.

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+
+	if client.ctx.Err() != nil {
+		client.logger.Errorw("Cannot process submitted entity as client context is done", zap.Error(client.ctx.Err()))
+		return true
+	}
+
+	var ev *event.Event
+	var eventErr error
+
+	// Keep the type switch in sync with syncPipelines from pkg/icingadb/history/sync.go
```
julianbrost: That comment seems to be outdated.

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+// Based on the entity's type, a different kind of event will be constructed. The event will be sent to the API in a
+// blocking fashion.
+//
+// Returns true if this entity was processed or cannot be processed any further. Returns false if this entity should be
+// retried later.
+//
+// This method usees the Client's logger.
+func (client *Client) Submit(entity database.Entity) bool {
+	client.submissionMutex.Lock()
+	defer client.submissionMutex.Unlock()
```
julianbrost: What exactly is this mutex protecting? Like does it really need to span the entire function?

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+			"host": rlr.HostName,
+		}
+	}
+
+	return &event.Event{
+		Name: objectName,
+		URL:  objectUrl.String(),
+		Tags: objectTags,
+	}, nil
+}
```
julianbrost: I don't think `url.URL` is still doing anything useful anymore, it just inserts a `?` now. That could also be string concatenation.

Also, the variables provide little benefit here, both branches could just `return &event.Event{...}` directly.

Finally, there's no more case returning an error, so that could be removed from the function signature.

pkg/icingadb/history/sync.go
```diff
@@ -354,32 +380,210 @@ func userNotificationStage(ctx context.Context, s Sync, key string, in <-chan re
...
+					backlogMsgCounter++
+					telemetry.Stats.Callback.Add(1)
+
+					if len(msgs) == 1 {
+						backlogLastId = ""
+						logger.Infow("Finished rolling back backlog of callback elements", zap.Int("delay", backlogMsgCounter))
+					} else {
+						backlogLastId = msgs[1].ID
+						backlogTimerInterval = backlogTimerMinInterval
+						_ = backlogTimer.Reset(backlogTimerInterval)
```
julianbrost: Wont this effectively do a `sleep(10ms)`? So if everything works fine, won't this artificially limit throughput?

julianbrost (commented):
General question on something I didn't think about before: the history is passed over multiple streams that are processed independently. What consideration did you put into this? At first glance, this sounds like a nasty source of race conditions to me if things happen out of order.

pkg/notifications/redis_fetch.go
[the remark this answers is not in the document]
julianbrost: Any response at all would have been nice. If you need to figure out who is supposed to address the review comments, that would be a first step.

pkg/icingaredis/telemetry/stats.go
```diff
@@ -23,6 +28,7 @@ func WriteStats(ctx context.Context, client *redis.Client, logger *logging.Logge
 		"config_sync":     &Stats.Config,
 		"state_sync":      &Stats.State,
 		"history_sync":    &Stats.History,
+		"callback_sync":   &Stats.Callback,
```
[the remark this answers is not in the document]
oxzi: Refactored to allow callback-specific names.

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+		redisClient: rc,
+
+		ctx: ctx,
+	}
+
+	notificationsClient, err := source.NewClient(client.Config, fmt.Sprintf("Icinga DB %s", internal.Version.Version))
+	if err != nil {
+		logger.Fatalw("Cannot create Icinga Notifications client", zap.Error(err))
+	}
+	client.notificationsClient = notificationsClient
```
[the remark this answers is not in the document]
oxzi: Done.

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+		eventLogger.Errorw("Cannot evaluate rules for event, assuming no rule matched", zap.Error(err))
+		eventRuleIds = []string{}
+	}
+
+	ev.RulesVersion = client.rulesInfo.Version
+	ev.RuleIds = eventRuleIds
+
+	newEventRules, err := client.notificationsClient.ProcessEvent(client.ctx, ev)
+	if errors.Is(err, source.ErrRulesOutdated) {
+		eventLogger.Infow("Cannot submit event to Icinga Notifications due to rule changes, will be retried",
```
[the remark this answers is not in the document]
oxzi: Reworded.

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,428 @@
...
+		outRuleIds = append(outRuleIds, id)
+	}
+
+	return outRuleIds, nil
+}
+
+// buildCommonEvent creates an event.Event based on Host and (optional) Service names.
+//
+// This function is used by all event builders to create a common event structure that includes the host and service
+// names, the absolute URL to the Icinga Web 2 Icinga DB page for the host or service, and the tags for the event.
```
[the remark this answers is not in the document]
oxzi: Changed.

pkg/notifications/notifications.go
oxzi: Changed to string and reworked method a bit.

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,494 @@
...
+			rows, err := client.db.QueryContext(ctx, client.db.Rebind(webQuery.Query), queryArgs...)
+			if err != nil {
+				return false, err
+			}
+			defer func() { _ = rows.Close() }()
+
+			if !rows.Next() {
+				return false, nil
+			}
+			return true, nil
```
julianbrost: Why not just `return rows.Next(), nil`? Feels easier to understand for me as well: the rule matches (maybe a better name than "evaluates") if there a next (i.e. a first) row in the result.

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,494 @@
...
+	for id, filterExpr := range client.rulesInfo.Rules {
+		if filterExpr == "" {
+			outRuleIds = append(outRuleIds, id)
+			continue
+		}
+
+		var webRule IcingaDbWebRule
+		if err := json.Unmarshal([]byte(filterExpr), &webRule); err != nil {
+			return nil, errors.Wrap(err, "cannot decode rule filter expression as JSON into struct")
+		}
```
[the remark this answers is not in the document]
oxzi: Dunno, with the changes from <https://github.com/Icinga/icingadb/pull/998#discussion_r2469576501>, this goes down to parsing a simple JSON struct for each evaluation. Since the next action is firing (potentially expensive) SQL queries next, I don't think caching this makes things faster.

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,494 @@
...
+			switch param {
+			case ":host_id":
+				queryArgs = append(queryArgs, hostId.String())
+			case ":service_id":
+				if !serviceId.Valid() {
+					return nil, errors.New("host rule filter expression contains :service_id for replacement")
+				}
+				queryArgs = append(queryArgs, serviceId.String())
+			case ":environment_id":
+				queryArgs = append(queryArgs, environmentId.String())
```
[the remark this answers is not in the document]
oxzi: Honestly, I have no idea why I did so. Thus, changed it.

pkg/notifications/notifications.go
oxzi: Changed.

pkg/icingaredis/telemetry/stats.go
```diff
@@ -2,34 +2,74 @@ package telemetry
...
+var Stats = &StatsKeeper{}
+
+// Keys for different well known Stats entries.
+const (
+	StatConfig         = "config_sync"
+	StatState          = "state_sync"
+	StatHistory        = "history_sync"
+	StatOverdue        = "overdue_sync"
+	StatHistoryCleanup = "history_cleanup"
+)
```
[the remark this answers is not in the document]
oxzi: Effectively reverted this change.

pkg/icingadb/history/sorter.go
```diff
@@ -0,0 +1,320 @@
...
+		return errors.Wrap(err, "cannot parse Redis Stream ID")
+	}
+
+	submission := streamSorterSubmission{
+		msg:          msg,
+		args:         args,
+		out:          out,
+		streamIdMs:   ms,
+		streamIdSeq:  seq,
+		submitTimeNs: time.Now().UnixNano(),
```
julianbrost: Why `UnixNano()`? Go's `time.Time` automatically contains a [monotonic value](https://pkg.go.dev/time#hdr-Monotonic_Clocks) that you lose with this call. `submitTimeNs` seems to be only used in a comparison, so just storing the `time.Time` should be the better option for this.

pkg/notifications/fetch.go
```diff
@@ -0,0 +1,159 @@
...
+		if err := retry.WithBackoff(
+			ctx,
+			func(ctx context.Context) (err error) {
+				data, err = client.redisClient.HGet(ctx, key, id).Result()
+				return
+			},
+			retry.Retryable,
+			backoff.DefaultBackoff,
+			retry.Settings{},
+		); err != nil {
```
julianbrost: IMHO that's an example where it definitely doesn't make sense to initialize the variable within the `if` statement.
```suggestion
		err := retry.WithBackoff(
			ctx,
			func(ctx context.Context) (err error) {
				data, err = client.redisClient.HGet(ctx, key, id).Result()
				return
			},
			retry.Retryable,
			backoff.DefaultBackoff,
			retry.Settings{})
		if err != nil {
```

pkg/notifications/fetch.go
```diff
@@ -0,0 +1,159 @@
...
+	type customVar struct {
+		Name  string `db:"name"`
+		Value string `db:"value"`
+	}
+
+	getCustomVarsFromSql := func(ctx context.Context, typ string, id types.Binary) ([]customVar, error) {
+		stmt, err := client.db.Preparex(client.db.Rebind(
+			`SELECT customvar.name AS name, customvar.value AS value
+			FROM ` + typ + `_customvar
+			LEFT JOIN customvar
```
julianbrost: I'm not sure about the **`LEFT`** `JOIN` here. Does it serve a purpose? I think it will have an effect but I'm unsure if that's actually intentional: In case there's no row to join, this will have `NULL` values in the result. Given that the members of `customVar` are of type `string`, that should then result in an error.

pkg/icingadb/history/sorter.go
```diff
@@ -0,0 +1,402 @@
...
+
+				if time.Since(sub.submitTime) < submissionMinAge {
+					if sorter.verbose {
+						sorter.logger.Debugw("Stopped popping heap as submission is not old enough",
+							zap.Object("submission", sub),
+							zap.Int("submissions", submissionCounter),
+							zap.Duration("duration", time.Since(start)))
+					}
+
+					heap.Push(submissionHeap, sub)
```
julianbrost: It shouldn't be necessary to pop the element to take a peek at it:

> The minimum element in the tree is the root, at index 0.

-- https://pkg.go.dev/container/heap

pkg/icingadb/history/sync.go
```diff
@@ -361,32 +402,164 @@ func userNotificationStage(ctx context.Context, s Sync, key string, in <-chan re
...
+		if success {
+			statPtr.Add(1)
+		}
+		return success
+	}
+
+	sorter := NewStreamSorter(ctx, logger, sorterCallbackFn)
+
+	return func(ctx context.Context, s Sync, key string, in <-chan redis.XMessage, out chan<- redis.XMessage) error {
+		defer close(out)
```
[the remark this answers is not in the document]
oxzi: Thanks, I have added this commit to this branch.

julianbrost:
I have an idea on how this code could be restructured a bit and need to know what you think about it.

The high-level goal would be to move as much of the new code into `pkg/notifications/` as that package name pretty much describes what the code is added for. And I think I have an idea how this can be achieved with few changes (conceptually, given that code is moved around, measured by lines of code the change will be big):

The signature of `Sync.Sync()` would be changed as follows: Instead of adding a parameter of type `SyncCallbackConf` (as currently done in this PR), it would just take a `map[string /* redis key */]stageFunc` instead. That's mostly the same type as already used in the pipeline stage definitions for the history sync itself (only difference is that it's only a single function instead of a slice because that's all what's needed): https://github.com/Icinga/icingadb/blob/4250dae2254f1f8071331f78d85ec38e5c728f8e/pkg/icingadb/history/sync.go#L364

For each pipeline started by `Sync()`, it would just insert that additional stage (like it already does basically). As a result, that should then allow `makeSortedCallbackStageFunc` and `sorter.go` be moved to `pkg/notifications/`. As a result, the changes to the existing code would be smaller and better show how this feature hooks into everything.

oxzi:
@julianbrost: Reworked as suggested.

pkg/notifications/sorter_test.go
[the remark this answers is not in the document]
oxzi: I have just pushed cd7a5979d57a57ce0178109009fd53141f4e18c2 to speed up the tests, using just a fraction of the timeouts for the tests.

julianbrost (approved):

PR #1059, opened 2026-01-08
schema: Index user_notification_history.notification_history_id

### LINKED ISSUE #1003: IcingadB 1.4.0 cleanup causes high database load
## Describe the bug

I've set up IcingaDB with a few thousand hosts and a postgres cluster as backend.

since the upgrade from 1.1.1 to 1.4.0, including applying all outstanding database migrations, icingadb runs regular cleanups of history tables, notification_history cleanup takes a very long time due to missing indexes on `user_notification_history`.

If I create the index manually (`CREATE INDEX idx_user_notification_history_notification_history_id ON user_notification_history (notification_history_id);`), the performance is back to normal and cleanups take just a few milliseconds. Is this index missing in migrations?

* Icinga DB version: 1.4.0
* Icinga 2 version: 2.15.0
* Operating System and version: CentOS Stream 9

oxzi:
Add a new INDEX to user_notification_history.notification_history_id to speed up the notification history retention.

This column is a FOREIGN KEY to notification_history with a "ON DELETE CASCADE" clause. The notification_history table can be cleaned up by the notification retention, resulting in lots of DELETE queries. However, without the INDEX, the DELETE CASCADE might result in a full table scan for each retention operation.

I have further checked every other retention table, but no other had the same issue. Most retention tables have their primary key referenced by a foreign key in the history table, where indexes were already present.

Many thanks to @rezemble for both reporting this issue and coming up with the exact solution.

Fixes #1003.

### DIFF
--- a/schema/mysql/schema.sql
+++ b/schema/mysql/schema.sql
@@ -1142,7 +1142,9 @@ CREATE TABLE user_notification_history (
 
   PRIMARY KEY (id),
 
-  CONSTRAINT fk_user_notification_history_notification_history FOREIGN KEY (notification_history_id) REFERENCES notification_history (id) ON DELETE CASCADE
+  CONSTRAINT fk_user_notification_history_notification_history FOREIGN KEY (notification_history_id) REFERENCES notification_history (id) ON DELETE CASCADE,
+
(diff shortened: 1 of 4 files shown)


oxzi:
Renamed the schema upgrade files to suffix them with the PR number to ease merging multiple PRs with different schema upgrades.

PR #1060, opened 2026-01-08
Retry more SQL Queries

### LINKED ISSUE #997: overdue: Sync.initSync misses retryable query
## Describe the bug

After updating and restarting almost every package, a race between the relational database and Icinga DB resulted in a failing query which is currently not being retried.

```
FATAL   icingadb        read tcp [::1]:6464->[::1]:5432: read: connection reset by peer
can't perform "SELECT id FROM service_state WHERE is_overdue='y'"
github.com/icinga/icinga-go-library/database.CantPerformQuery
        github.com/icinga/icinga-go-library@v0.7.2/database/utils.go:19
github.com/icinga/icingadb/pkg/icingadb/overdue.Sync.initSync
        github.com/icinga/icingadb/pkg/icingadb/overdue/sync.go:89
github.com/icinga/icingadb/pkg/icingadb/overdue.Sync.Sync.func2
        github.com/icinga/icingadb/pkg/icingadb/overdue/sync.go:53
golang.org/x/sync/errgroup.(*Group).add.func1
        golang.org/x/sync@v0.15.0/errgroup/errgroup.go:128
runtime.goexit
        runtime/asm_amd64.s:1699
can't sync overdue indicators
github.com/icinga/icingadb/pkg/icingadb/overdue.Sync.Sync
        github.com/icinga/icingadb/pkg/icingadb/overdue/sync.go:57
main.run.func6.3
        github.com/icinga/icingadb/cmd/icingadb/main.go:215
golang.org/x/sync/errgroup.(*Group).add.func1
        golang.org/x/sync@v0.15.0/errgroup/errgroup.go:128
runtime.goexit
        runtime/asm_amd64.s:1699
```

https://github.com/Icinga/icingadb/blob/a9794098db5ef595e48d2ff3e5aed6aebd873d0b/pkg/icingadb/overdue/sync.go#L85-L90

## Expected behavior

Same retry logic as everywhere else.

## Your Environment

* Icinga DB version: 1.4.0

## Additional context

In general, we should scan the code for SQL queries not wrapped in our retry logic. I would guess there are some more which can be wrapped safely.

oxzi:
overdue.initSync: Retry SQL query

The overdue sync starts with an initializing SQL query which was not
retryed so far. Thus, if there is a race between the relational database
and Icinga DB, this query might fail and let Icinga DB crash.

This might happen if restarting Icinga DB together with the relational
database server, for example after an upgrade.

---

notifications.fetchCustomVarFromSql: Retry SQL query

The fetchCustomVarFromSql method is used deep in the call trace to fetch
custom variables for notification events to be submitted to the Icinga
Notifications API. If, for whatever reason, the SQL query within this
method fails, the event will be discarded.

By retrying the SQL query using our usual retry logic, temporary
database hiccups do not result in loosing an event.

---

notifications: Retry Timeout for fetchHostServiceFromRedis

Set the default timeout for the retryable HGET Redis query for fetching
host or service names. Without any timeout so far, this query could fail
and be retryed until eternity.

---

Fixes #997.

### DIFF
--- a/pkg/icingadb/overdue/sync.go
+++ b/pkg/icingadb/overdue/sync.go
@@ -5,11 +5,13 @@ import (
 	_ "embed"
 	"fmt"
 	"github.com/google/uuid"
+	"github.com/icinga/icinga-go-library/backoff"
 	"github.com/icinga/icinga-go-library/com"
 	"github.com/icinga/icinga-go-library/database"
 	"github.com/icinga/icinga-go-library/logging"
 	"github.com/icinga/icinga-go-library/periodic"
 	"github.com/icinga/icinga-go-library/redis"
+	"github.com/icinga/icinga-go-library/retry"
 	"github.com/icinga/icingadb/pkg/icingadb/v1"
 	"github.com/icinga/icingadb/pkg/icingadb/v1/overdue"
 	"github.com/icinga/icingadb/pkg/icingaredis/telemetry"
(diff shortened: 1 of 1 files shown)


pkg/notifications/fetch.go
```diff
@@ -30,7 +31,7 @@ func (client *Client) fetchHostServiceFromRedis(
 			},
 			retry.Retryable,
 			backoff.DefaultBackoff,
-			retry.Settings{},
+			retry.Settings{Timeout: retry.DefaultTimeout},
```
lippserd: This seems like a change that should be explained in a separate commit.

oxzi: Thanks, I have split the changes into three independent commits, each explaining containing reasoning.

PR #1120, opened 2026-05-22
notifications: Attribute Negotiation

### LINKED ISSUE #1098: Support for Icinga Notifications v1.0 – Attribute Negotiation
With v1.0, Icinga Notifications will require a different event structure. This is about changes to how rules are evaluated.

https://github.com/Icinga/icinga-notifications/issues/406 introduces _attribute negotiation_. Each event is expected to describe the _relations_ an object has. This moves actual rule evaluation back to Icinga Notifications.

Icinga DB is not expected to always include all _relations_, let alone all attributes of one. The default should be a _sane_ default. (e.g. The object itself and it's own groups. Attributes for each: `name` and `display_name`)

During evaluation of a rule filter, Icinga Notifications may detect that a relation or one of its attributes is missing. In this case, it will respond with a list of missing components in the form of a JSONPath. This may look like this:

Request:
```json
{
  "type": "attribute negotiation",
  "attributes": [
     "servicegroups",
     "host.address"
  ]
}
```

Response: (Showing a full update)
```json
{
  "relations": {
    "host": {
      "name": "web-1-a",
      "display_name": "Web 1 A",
      "address": "10.1.8.25"
    },
    "hostgroups": [
      {
        "name": "web-servers",
        "display_name": "Web Servers"
      }
    ],
    "servicegroups": [
      {
        "name": "http-services",
        "display_name": "HTTP Services"
      }
    ]
  }
}
```

If Icinga DB is configured to always include particular relations, it can signal Icinga Notifications that these are complete and cannot be updated by using a request header: `X-Icinga-Complete-Relations: host.vars, services[*].vars`

Though, the actual protocol isn't set in stone yet. This should be streamlined during implementation and reconciled with whoever implements https://github.com/Icinga/icinga-notifications/issues/406.

The previous implementation in v1.5.0 also allowed to differentiate between object types in filter rules, which is why there should be another relation describing the nature of the object itself:

```json
{
  "relations": {
    "object": {
      "type": "host"
    },
    "host": {
      "name": "web-1-a",
      "display_name": "Web 1 A",
      "address": "10.1.8.25"
    }
  }
}
```

oxzi:
Initial support for attribute negotiation, the big change with Icinga Notifications v1.0.

Now, the notifications.Client does not evaluate the rules by itself anymore, but passes a subset of available information to Icinga Notifications. From there on, Icinga Notifications might return multiple JSONPath objects, which are being used to enrich the submitted event.

This change required an adjustment to the icinga-go-library, which was bumped here to its PR branch.

---

The IGL change was suggested in https://github.com/Icinga/icinga-go-library/pull/201.

Fixes #1098.


oxzi:
Updated the PR after talking about it with @yhabteab offline. Next to some smaller bugfixes and changes, the following more prominent changes were made:
- Bulk fetch Redis objects of the same type via HMGET.
- Bulk fetch custom vars from SQL via IN queries.
- Track complete relations and submit with the event.
- Drop fetching custom vars for host and service groups.

oxzi:
Sorry, pushed a [small change](https://github.com/Icinga/icingadb/compare/3f347ee2b5a0fea623852810a2401c322f17c8d3..208bf74b06eff214f79e9a1866d0f3b54b9d72ac) to ensure that the complete relations are always populated and remove duplicate code.

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,488 @@ package notifications
...
+	key := "icinga:" + typ
+	strIds := make([]string, 0, len(ids))
+	for _, id := range ids {
+		strIds = append(strIds, id.String())
+	}
+
+	var out map[string]map[string]any
+	err := retry.WithBackoff(
+		ctx,
+		func(ctx context.Context) (err error) {
```
yhabteab: This named return value isn't required here as you're returning the error explicitly.

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,488 @@ package notifications
...
+
+			for {
+				select {
+				case pair, ok := <-pairCh:
+					if !ok {
+						return nil
+					}
+
+					obj := make(map[string]any)
+					if err := json.Unmarshal([]byte(pair.Value), &obj); err != nil {
```
yhabteab: You're unmarshalling the entire JSON representation of the object here, but we only need `name` and `display_name` for the moment because all the other ones aren't allowed to be used as filter columns in Notifications Web. See https://github.com/Icinga/icingadb-web/blob/main/library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php#L39-L44 (display_name isn't listed there but will be soon, I guess).

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,488 @@ package notifications
...
+	if len(multi) == 0 {
+		// Objects without any custom vars are also valid.
+		return nil, nil
 	}
 
-	return
+	obj, ok := multi[id.String()]
+	if !ok {
+		return nil, errors.Errorf("no custom vars for %q of type %q", id, typ)
+	}
```
yhabteab: Is this really necessary? `fetchMultiCustomVarsFromSql` should return an empty map if there are no custom vars for the given ID (which is already handled above), so there's no way that the requested ID is not in the map at this point. You should be able to just return `multi[id.String()]` without adding such existence extra check.

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,488 @@ package notifications
...
-				}
-				defer func() { _ = stmt.Close() }()
+// relationsProvider implements a relations provider backed by the Client, to be replaced for testing.
+func (client *Client) relationsProvider(ctx context.Context, rel *relations, query []string) error {
+	if len(query) == 0 {
+		return errors.New("cannot provide relations for empty query")
+	}
+
+	// NOTE: At the moment, the root selector in the query is enough to fetch all child attributes. This should be open
+	// for consideration and potentially be changed later.
```
yhabteab: I think that's still an overkill as you're still sending (if I'm not missing something) all the customvars of a given type with the initial request but not their corresponding groups. Quote from #1098:

> Icinga DB is not expected to always include all _relations_, let alone all attributes of one. The default should be a _sane_ default. (e.g. The object itself and it's own groups. Attributes for each: `name` and `display_name`)

So, I suggest to do this the other way around, i.e., include hostgroups for hosts and servicegroups for services respectively in the initial request and retrieve the custom vars only when Icinga Notifications rejects the request with at least one attribute that requires customvars.

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,488 @@ package notifications
...
-	}
+	case `["servicegroups"]`:
+		if slices.Contains(rel.completeRelations, "servicegroups") {
+			return nil
+		}
+
+		// For the moment, consider that hosts cannot have servicegroup memberships.
+		if rel.serviceId == nil {
+			return nil
+		}
```
yhabteab: What does _for the moment_ mean? In this case, the memberships have to be constructed via host -> services -> servicesgroups since there's no such constraint in Notifications Web.

pkg/notifications/notifications.go
```diff
@@ -175,54 +101,89 @@ func (client *Client) evaluateRulesForObject(ctx context.Context, hostId, servic
...
 	}
 
-	return &event.Event{
-		Name:      objectName,
-		URL:       objectUrl,
-		Tags:      objectTags,
-		ExtraTags: info.customVars,
-	}, info, nil
+	ev := &fetchableEvent{
+		Event: &event.Event{Name: objectName,
```
yhabteab: The formatting seems to be odd here ;)

yhabteab (commented):
Quote from #1098:

> If Icinga DB is configured to always include particular relations, it can signal Icinga Notifications that these are complete and cannot be updated by using a request header: `X-Icinga-Complete-Relations: host.vars, services[*].vars`

This is still missing!

pkg/notifications/fetch.go
oxzi: It felt like a more defensive approach to me.

If, under whatever circumstances, the `fetchMultiCustomCarsFromSql` method returns a non empty map which does not contain this `id` as a field, the code would panic. At the moment, this should not be possible, unless the SQL server misbehaves or a bug is present in the method. However, the method can change in the future, introducing a slightly changed behavior.

So, I would argue in favor for this check as it costs next to nothing and ensures we don't panic during runtime.

yhabteab: > If, under whatever circumstances, the `fetchMultiCustomCarsFromSql` method returns a non empty map which does not contain this `id` as a field, the code would panic.

Where would it panic? I mean, what's even the point of that check? In case the SQL query didn't yield nothing you just return nil but you error out in this case? If `multi` is a non-nil map (which is guaranteed to be non-nil here), then `multi[id.String()]` should never panic but would return nil instead if the key doesn't exist.

pkg/notifications/fetch.go
oxzi: While the web code supports your argument, this PR's issue #1098 also lists other fields, such as the `address` for the legacy IP address.

yhabteab: The restriction was added while implementing the Web counter part for this, so that issue is a bit outdated (though, one can also take everything from that issue as an example and not something that should be strictly followed). So, we should follow the web implementation! I've just created a new issue for the missing columns https://github.com/Icinga/icingadb-web/issues/1379.

pkg/notifications/fetch.go
oxzi: Sorry, you are of course right. Just forget what I wrote earlier.

pkg/notifications/fetch.go
oxzi: I wasn't aware that the issue was outdated as I only briefly followed the web part and expected the issue to be accurate. Thanks for the clarification and the issue.

oxzi:
[Updated the PR](https://github.com/Icinga/icingadb/compare/208bf74b06eff214f79e9a1866d0f3b54b9d72ac..d97709ce0f76d63184496a85da97ab351aa0a738) and tried to address all of @yhabteab's review comments. Most prominent, the following changes were made:

- Limit object fields to `name` and `display_name`.
- Only fetch host and service object information by default, no custom vars or groups anymore.
- Inspect the provided JSONPath in more detail and re-fetch required attributes.
- Add a minimal test suite.

However, I still have to address the main review comment, which I am going to do tomorrow. Afterwards, maybe a bit of refactoring would be pleasant for the reviewer's eyes.

oxzi:
After a bit of internal refactoring and adding configuration options, I would kindly ask you to take another look, @yhabteab. All prior comments should have been addressed in the diff.

internal/config/config.go
```diff
@@ -50,6 +51,12 @@ func (c *Config) Validate() error {
 		return errors.Wrap(err, "invalid notifications configuration")
 	}
 
+	for _, relation := range c.Notifications.DefaultRelations {
+		if _, err := jsonpath.Parse(relation); err != nil {
+			return errors.Wrapf(err, "invalid default relation %q is not a valid JSONPath", relation)
```
yhabteab: s/invalid //.

Also, is it expected that users also include the JSONPath root selector? I'd make this optional but I'm also fine to require it from users.

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,678 @@ package notifications
...
+	err := retry.WithBackoff(
+		ctx,
+		func(ctx context.Context) error {
+			out = make(map[string]*icingaObject)
+			pairCh, errCh := client.redisClient.HMYield(ctx, key, strIds...)
+
+			g, ctx := errgroup.WithContext(ctx)
+
+			com.ErrgroupReceive(g, errCh)
+			g.Go(func() error {
```
yhabteab, line 57: Is this extra worker group actually necessary?

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,678 @@ package notifications
...
+	// Vars might be unpopulated for certain types, such as host or service groups.
+	Vars map[string]any `json:"-"`
 }
 
-// getValue returns this customvar's value as a string, transforming SQL NULLs to empty strings.
-func (cv customVar) getValue() string {
-	if cv.Value.Valid {
-		return cv.Value.String
+// asMap creates a Go map populated by this icingaObject.
+func (obj *icingaObject) asMap() map[string]any {
```
yhabteab: I wonder if this would be necessary at all when using `vars,omitempty` as a json tag for the `Vars` object field instead? Effectively, making the that struct json serializable without manual conversion into a map. You ould then just use...

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,678 @@ package notifications
...
+}
+
+// asMap creates a Go map populated by this relations and all its fields to be used in event.Event.
+func (rel *relations) asMap() map[string]any {
+	out := make(map[string]any)
+
+	out["object"] = map[string]any{"type": rel.Object.Type}
+
+	if rel.Host != nil {
+		out["host"] = rel.Host.asMap()
```
yhabteab: ...these objects as they are and use the `rel.Host` object and others too directly.

yhabteab (commented):
I've left few comments from briefly looking at the code. But, I can tell you that it looks way better than the previous version. I need to spend some more time to understand the `relationsProvider` implementation though, since it seems to be doing a lot of things. I'll continue to review the code tomorrow!

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,678 @@ package notifications
...
-		if err != nil {
-			return nil, err
+	for _, selector := range segments[0].Selectors() {
+		rootSelector := selector.String()
+		childSegments := segments[1:]
+
+		var fetchObject, fetchCustomVars bool
+		if len(childSegments) == 0 {
+			fetchObject = true
+			fetchCustomVars = true
```
yhabteab, line 342: This and all the other cases below where you set `fetchCustomVars` to true causes to always print a warning about customvars being not supported for groups even though I didn't use `hostgroups[*].vars` as a filter.

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,678 @@ package notifications
...
+		} else {
+			// Host is the only directly used object, while every other object is within an array, one layer deeper.
+			var selectors []spec.Selector
+			if rootSelector == `"host"` {
+				selectors = childSegments[0].Selectors()
+			} else {
+				if len(childSegments) == 1 {
```
yhabteab: This can be an `else if` :)

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,678 @@ package notifications
...
+						return
+					})
+				}
+				if err := g.Wait(); err != nil {
+					return errors.Wrap(err, "cannot fetch service information")
+				}
+
+				if fetchCustomVars {
+					serviceObj.Vars = serviceCustomVars
+				}
```
yhabteab: I think, you don't need the if check here :)

```suggestion
				serviceObj.Vars = serviceCustomVars
```

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,678 @@ package notifications
...
+
+			serviceGroupsSlice := make([]*icingaObject, 0, len(serviceGroups))
+			for _, serviceGroup := range serviceGroups {
+				serviceGroupsSlice = append(serviceGroupsSlice, serviceGroup)
+			}
+
+			updateCompleteRelations("servicegroups[*]")
+			rel.Servicegroups = serviceGroupsSlice
+		default:
+			return errors.Errorf("unsupported JSONPath root segment selector %q", rootSelector)
```
yhabteab, line 565: Using `object.type` as a filter lands here now.

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,678 @@ package notifications
...
+
+		childSegmentsStr := make([]string, 0, len(childSegments))
+		for _, childSegment := range childSegments {
+			childSegmentsStr = append(childSegmentsStr, childSegment.String())
+		}
+
+		client.logger.Debugw("Client evaluated relation selector",
+			zap.String("root_selector", rootSelector),
+			zap.Any("child_segments", childSegmentsStr),
+			zap.Any("complete_relations", rel.completeRelations))
```
yhabteab:
```suggestion

		client.logger.Debugw("Client evaluated relation selector",
			zap.String("root_selector", rootSelector),
			zap.Stringers("child_segments", childSegments),
			zap.Strings("complete_relations", rel.completeRelations))
```

pkg/notifications/notifications.go
```diff
@@ -414,60 +358,39 @@ func (client *Client) Submit(entity database.Entity) bool {
...
-				zap.String("new_rules_version", newEventRules.Version))
-
-			client.rulesInfo = newEventRules
+	maxAttempts := 3
+	for attempt := range maxAttempts {
+		attributes, err := client.notificationsClient.ProcessEvent(client.ctx, ev.Event, true)
+		if errors.Is(err, source.ErrAttrsNegotiation) {
+			client.logger.Info("Icinga Notifications requests more attributes",
+				zap.String("event", ev.Name),
+				zap.Any("attributes", attributes))
```
yhabteab:
```suggestion
				zap.Strings("attributes", attributes))
```

pkg/notifications/notifications.go
```diff
@@ -414,60 +358,39 @@ func (client *Client) Submit(entity database.Entity) bool {
...
+				// Another completeAndUpdate call would be useless, as not evaluated anymore.
+				break
+			}
 
-			continue
+			err := ev.completeAndUpdate(client.ctx, attributes)
+			if err != nil {
+				client.logger.Errorw("Cannot fetch required attribute for event",
+					zap.String("event", ev.Name),
+					zap.Any("attributes", attributes),
```
yhabteab:
```suggestion
					zap.Strings("attributes", attributes),
```

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,678 @@ package notifications
...
+
+				services := make([]*icingaObject, 0, len(serviceObjs))
+				for id, service := range serviceObjs {
+					if fetchCustomVars {
+						customVars, ok := serviceCustomVars[id]
+						if ok {
+							// There might be services without any custom vars. So don't error out.
+							service.Vars = customVars
+						}
+					}
```
yhabteab:
```suggestion
					if len(serviceCustomVars) > 0 {
						service.Vars = serviceCustomVars[id]
					}
```

internal/config/config.go
oxzi: > Also, is it expected that users also include the JSONPath root selector?

Locally, I had one version where I verified that it contains the root selector followed by either `host`, `services`, `hostgroups`, or `servicegroups`, but then it mimicked the implementation too much. So I dropped it.

However, the root selector seems to be a good idea; added.

pkg/notifications/fetch.go
oxzi: It eases checking the error channel via `com.ErrgroupReceive` and is a pattern used at other places as well. Otherwise, I would need to check the `errCh` even after `pairCh` was closed.

pkg/notifications/fetch.go
oxzi: I was already asking myself something similar, but came back to the manual `map` as `event.Event.Relations` requires a `map[string]any`. Otherwise, I would have thought about something like a `json.RawMessage` or the like.

However, when JSON marshalling `icingaObject`, I had to unmarshall it to place it into the `´event.Event.Relations` and this sounds unnecessary expensive.

oxzi: Nevermind, I have misunderstood you. Thanks for the clarification offline. I am going to change this :)

pkg/notifications/fetch.go
yhabteab: Ok, fine!

pkg/notifications/fetch.go
oxzi: Seems fair. Same also applied for the `host` case.

pkg/notifications/fetch.go
oxzi: TIL about `zap.Stringers`, neat.

pkg/notifications/fetch.go
yhabteab: Also using the `q` verb to quote the error string causes the selector to be double-quoted (since it's already quoted) which kind looks ugly IMHO in the logs.

oxzi: What would you suggest otherwise? The `%q` verb ensures that one cannot mess with the output. Furthermore, not every first segment must be a single object name, it can also be a JSON array or a wildcard, e.g., for `$.*`.

yhabteab: Ok, NVM!

oxzi:
Tried to address all comments next to some changes, such as a combined SQL query to fetch host and service groups instead of an SQL query for IDs and a Redis query afterwards.

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,601 @@ package notifications
...
 
-	return
+	return obj, nil
 }
 
-// customVar is used as an internal representation in Client.fetchCustomVarFromSql.
-type customVar struct {
-	Name  string       `db:"name"`
-	Value types.String `db:"value"`
+// fetchObjectFromSql executes a query for one ID parameter and returns the output in the output argument pointer.
```
yhabteab: This desc is outdated now.

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,601 @@ package notifications
...
+	for _, customVar := range customVars {
+		var (
+			customVarMap   map[string]any
+			customVarValue any
+			ok             bool
+		)
+
+		customVarMap, ok = vars[customVar.TypId.String()]
+		if !ok {
+			customVarMap = make(map[string]any)
```
yhabteab, line 219: I would also move the map re-assignment from the end of the loop to within this if block like this as it's required only in this case.

```suggestion
		if !ok {
			customVarMap = make(map[string]any)
			vars[customVar.TypId.String()] = customVarMap
```

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,601 @@ package notifications
...
+					fetchObject = true
+				}
+			}
+		}
+
+		updateCompleteRelations := func(prefix string, isGroup bool) {
+			var newRelations []string
+
+			if fetchObject {
+				// Both name and display_name are fetched together via Redis.
```
yhabteab: s/via/from/.

pkg/notifications/fetch.go
```diff
@@ -3,187 +3,601 @@ package notifications
...
+				newRelations = append(newRelations, "vars")
+			}
+
+			for _, newRelation := range newRelations {
+				k := prefix + "." + newRelation
+				if !slices.Contains(rel.completeRelations, k) {
+					rel.completeRelations = append(rel.completeRelations, k)
+				}
+			}
+		}
```
yhabteab, line 392: Would it be possible to make this a method of the `relations` type? This would make the loop a bit smaller.

pkg/notifications/fetch.go
oxzi: Personally, I prefer my version as I find it easier to read in terms of data flow and I don't have to keep in mind that a map is a reference. But as it makes no difference in the behavior, I am going to change it..

pkg/notifications/fetch.go
oxzi: I have explicitly inlined this function as this allows referencing `fetchObject` and `fetchCustomVars`. Otherwise, this method would take lots of parameters and would be quite hollow as it cannot be used outside the `Client.relationsProvider` anyway. Furthermore, I am not afraid of long loops/functions/methods as long as they are well formatted :)

yhabteab: > Furthermore, I am not afraid of long loops/functions/methods as long as they are well formatted :)

The function is ~250 LOC long, that's really long but I don't see how one can easily extract some of its body into functions, so 🤷! I was just thinking maybe outsourcing this function might make it a couple of LOC shorter.

yhabteab (dismissed):
I've also tested it with PosgreSQL, so LFTM!

oxzi:
Thanks, but please wait a second. I have a small diff I still want to push :upside_down_face:

oxzi:
Pushed a small change.
- Document the constraints of the config check.
- Went over the code comments again and reworded a few things.
- Created a method for `relations` to set the object type and complete `object.type` in the complete relations, which was missing before.
- Always use `services[*]` instead of a mix of `services.*` and `services[*]`. While these are the same JSONPath, I preferred some unity in the code.

yhabteab (approved):

PR #1128, opened 2026-06-11
Process customvar runtime updates only as part of the config udpates

yhabteab:
The `RuntimeUpdates::Sync` method is invoked twice, both for the normal runtime config and state updates. However, the customvars runtime updates processing logic was running in parallel in both invokations. There were already curly brackets around that code, probably because someone wanted to add some kind of condition for that, so I've add just added the missing condition for that.

### DIFF
--- a/pkg/icingadb/runtime_updates.go
+++ b/pkg/icingadb/runtime_updates.go
@@ -144,8 +144,8 @@ func (r *RuntimeUpdates) Sync(
 		})
 	}
 
-	// customvar and customvar_flat sync.
-	{
+	// customvar and customvar_flat sync don't need to be processed with state updates too.
+	if _, exists := streams["icinga:runtime"]; exists {
 		updateMessages := make(chan redis.XMessage, r.redis.Options.XReadCount)
 		upsertEntities := make(chan database.Entity, r.redis.Options.XReadCount)
 		deleteIds := make(chan any, r.redis.Options.XReadCount)


pkg/icingadb/runtime_updates.go
```diff
@@ -144,8 +144,8 @@ func (r *RuntimeUpdates) Sync(
 		})
 	}
 
-	// customvar and customvar_flat sync.
-	{
+	// customvar and customvar_flat sync don't need to be processed with state updates too.
+	if len(streams) == 1 && streams["icinga:runtime"] != "" {
```
oxzi: The `len(streams) == 1` part of the condition confuses me. Why did you add this one as well?

The two calls to `RuntimeUpdates.Sync` are either with the `config` or the `state` stream, created in `RuntimeUpdates.ClearStreams` and called in the endless main `run` function. Unless I am missing something, these streams were not altered later and are both just a map with a single key, either `icinga:runtime` or `icinga:runtime:state`. Thus, this condition is always true and will only result to bugs in case of future changes?

https://github.com/Icinga/icingadb/blob/af392f65ab6718c1d6cd3a3dc23e21a4e6a44411/pkg/icingadb/runtime_updates.go#L45-L47

https://github.com/Icinga/icingadb/blob/af392f65ab6718c1d6cd3a3dc23e21a4e6a44411/cmd/icingadb/main.go#L218

So, unless I am overseeing something, wouldn't the following be more straight forward?

```suggestion
	if _, ok := streams["icinga:runtime"]; ok {
```

Sorry for the long reply over a one line diff, but the length check for exactly one did confuse me :(

oxzi (commented):
Tested the change and it seems to work. However, as I was missing a bit of background (How did you stumble about this? Why did you change this?) it was a bit hard to follow this PR's motivation.

yhabteab:
> (How did you stumble about this?

Is it relevant how I came accross that thing? I'm currently making changes the `RuntimeUpdates.Sync` method for the Icinga Notifications component, and just noticed this little one and wanted to fix it separately.

> Why did you change this?) it was a bit hard to follow this PR's motivation.

I changed it, because, it obviously is a bug (even if it's a harmless one). Customvars are not state but config updates and as such they should only be processed as part of the respective `icinga:runtime` pipeline.

pkg/icingadb/runtime_updates.go
yhabteab: > The `len(streams) == 1` part of the condition confuses me. Why did you add this one as well?

I just wanted to make sure to process the customvars, if and only if the `icinga:runtime` key is exlusively provided, but if that's to confusing then I don't mind changing it.

oxzi:
> > (How did you stumble about this?
> 
> Is it relevant how I came accross that thing?

Knowing the motivation behind a change helps me a lot reviewing it, especially as it is in parts of the code which were quite untouched for a while. When looking at this PR, I don't know what's in your head :)

pkg/icingadb/runtime_updates.go
oxzi: This is fine. Maybe add a comment stating this. For me, this read like a round about way to check that the map is not empty, which should not happen as the map should (at the moment) always contain one element, as outlined above.

yhabteab: Just applied your suggestion! That's simpler than having to explain something in a comment :)

oxzi (approved):

PR #1130, opened 2026-06-15
notifications: consume state change events from runtime updates

### LINKED ISSUE #1127: Update event transmission to Icinga Notificiations v1.0
With Icinga Notifications v1.0 the event structure changes in a fundamental way. Events are now not fully compatible anymore with Icinga Events. The `type` key is gone and instead four new flags are introduced to represent similar functionality:

- `incident: true` To open an incident or to trigger escalations
- `close: true` To close an incident irrespective of the severity
- `muted: bool` To mute/unmute an incident (Always in combination with `muted_reason: string`)
- `notify: true` To bypass checks that may prevent notifications

## Events

In order to avoid the need for any in-memory cache and to ensure to prevent false-positives and -negatives, history events are not relevant anymore with regards to incidents. Runtime updates are the sole base for the notifications component and hence it will need to always transmit e.g. the `muted` flag with every transmission.

The severity always reflects the current hard state and the message the latest check result. The other flags are determined as follows:

| Problem | Downtime / Flapping / Ack | Volatile Check | Result |
|-|-|-|-|
| y | n | n | `{"incident": true, "muted": false}` |
| y | n | y | `{"incident": true, "muted": false, "notify": true}` |
| n | y | 🚫 | `{"incident": true, "close": true, "muted": true}` |
| n | n | 🚫 | `{"incident": true, "close": true, "muted": false}` |
| y | y | 🚫 | `{"incident": true, "muted": true}` |

yhabteab:
Currently, all the relevant events are consumed from the respective history pipelines, and were not as is the case for histories, HA aware and thus were running on both Icinga DB HA instances. This meant that both instances were consuming the same events, but since Icinga Notifications had a logic to detect such superfluous events, it was not a problem. Also, the history streams are lacking some information, in order to be able to determine whether a checkable is in downtime, flapping state, or is acknowledged, which is relevant for each state change event. So, the first attempt in overcoming this was to add the missing flags to the state change history stream (basically, patching Icinga 2 and requiring a specific version of it just for this), but fortunately, @nilmerg came up with a much better solution, which I hadn't thought of, which is to consume the events directly from the runtime updates stream, which already contains all the relevant flags. Apart from that, the runtime updates are also already HA aware, so just like the other runtime updates consumers, the Icinga Notifications component will also pause consuming events on the passive instance.

Now, this change alone wouldn't have been that big of a deal, because my initial plan was to just hook into the existing runtime state updates pipeline, but then after some discussion with Johannes, he suggested to completely decouple the Notifications component from the existing runtime updates pipeline. This means that instead of making the Notifications component dependent on the success rate and performance of the database queries of the runtime updates pipeline, it will have its own dedicated Redis message reader. This way, both the Notifications component and the existing runtime updates pipeline can operate independently (can move their own stream offsets without affecting each other) while still consuming the same events from the same Redis stream. Obviously, this also means that they can't send XDel commands for each message they consume independently, because that would cause the other component to miss messages, so instead, they will both just acknowledge the messages they consume, and a separate collector goroutine will be responsible for tracking the acknowledged messages and sending the XDel commands to Redis in batches. Note that is not a Redis consumer group, but rather a Go channel based implementation, that mimics some of the consumer group concepts.

~Last but not least, in order to always have the latest check output of a checkable, there is also a new goroutine that periodically (every 5 minutes) syncs the check output of all checkables that are currently in a non-OK state from the `icinga:{host,service}:state` Redis hashes to Icinga Notifications. This is necessary because a checkable's check can, for instance, be in `CRITICAL` state for a long time, while its output can change with each check result. And since that won't trigger any state change events, thus, not queued to the runtime updates stream, there's no way for Icinga Notifications to know about the updated output.~ **EDIT** that part of the code is no longer part of this PR.

All in all, with this PR, all the code related to consuming history events from the history pipelines has become obsolete, and isn't used anymore. However, since we are planning to implement a way to send non-state notifications (e.g. for downtime start/end, acknowledgement ste, etc.) in the future, which will eventually require consuming the respective history events, I didn't remove the code yet, instead, I just temporarily disabled it with a reference to respective Icinga Notifications issue.

## Requires

- https://github.com/Icinga/icinga-go-library/pull/212
- https://github.com/Icinga/icinga-notifications/pull/442

resolves #1127


pkg/notifications/notifications.go
```diff
@@ -143,36 +199,49 @@ func (client *Client) buildCommonEvent(
...
+		return nil, errors.Wrapf(err, "cannot build event for %q,%q", hostId, serviceId)
 	}
 
-	ev.Type = event.TypeState
+	if s.Output.Valid {
+		ev.Message = s.Output.String
+	}
+	if s.LongOutput.Valid {
+		ev.Message += "\n" + s.LongOutput.String
+	}
```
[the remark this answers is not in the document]
yhabteab: `Output` can never be empty if `LogOutput` is set, so no, it's fine as-is right now.

pkg/notifications/notifications.go
```diff
@@ -87,13 +84,72 @@ func NewNotificationsClient(
...
+			if err != nil {
+				client.logger.Errorw("Failed to sync service state check outputs", zap.Error(err))
+				if errors.Is(err, context.Canceled) {
+					return ctx.Err()
+				}
+			}
+			lastSync = now
+			// In case the sync took a long time, the next tick might come immediately,
+			// so we reset the ticker to ensure a full interval between syncs.
+			ticker.Reset(interval)
```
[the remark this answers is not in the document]
yhabteab: LOL! You need to keep up my thing :)! Since Go 1.23 `Reset` drains the channel automatically.

pkg/notifications/notifications.go
```diff
@@ -182,24 +251,48 @@ func (client *Client) buildStateHistoryEvent(ctx context.Context, h *v1history.S
...
-			return nil, fmt.Errorf("unexpected service state %d", h.HardState)
+			return nil, fmt.Errorf("unexpected service state %d", s.HardState)
 		}
 	} else {
-		switch h.HardState {
+		switch s.HardState {
 		case 0:
 			ev.Severity = event.SeverityOK
 		case 1:
 			ev.Severity = event.SeverityCrit
```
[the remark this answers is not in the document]
yhabteab: NEVER EVER!

pkg/notifications/notifications.go
```diff
@@ -143,36 +199,49 @@ func (client *Client) buildCommonEvent(
...
+	} else {
+		isVolatile = ev.Host.isVolatile
+	}
+
+	// For non-hard state changes, we want to just update the event message without affecting any other incident state.
+	// However, if the checkable is volatile, it's always treated as a hard state change, it's still set to `SOFT` due
+	// to an Icinga 2 bug (see https://github.com/Icinga/icinga2/issues/10879).
+	if s.StateType == volatileState || (s.StateType != common.HardState && !isVolatile) {
+		return ev, nil
+	}
```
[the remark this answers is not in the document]
yhabteab: That part of the code is gone now, so resolving this!

pkg/icingadb/runtime_updates.go
```diff
@@ -351,34 +408,56 @@ func structifyStream(
...
 			}
 		}
 	}
 }
+
+// RUOption defines an option for the [RuntimeUpdates.Sync] method.
+type RUOption func(*RUOptions)
+
+// RUupsertFunc defines the type of the callback that can be provided to the [RuntimeUpdates.Sync] method via the [WithRMOnUpsert] option.
+type RUupsertFunc func(context.Context, database.Entity) error
```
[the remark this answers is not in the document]
yhabteab: Fixed.

yhabteab:
Rebased and dropped all the timer aka check output syncer related code from this PR as discussed offline. I'll open another PR for that once this PR is done.

yhabteab:
Fixed some _bugs_ caused by the newest changes in the referenced IGL PR.

pkg/notifications/notifications.go
```diff
@@ -214,23 +251,15 @@ func (client *Client) buildDowntimeHistoryMetaEvent(ctx context.Context, h *v1hi
...
-		ev.Type = event.TypeDowntimeStart
-		ev.Username = h.Author
 		ev.Message = h.Comment
-		ev.Mute = types.MakeBool(true)
-		ev.MuteReason = "Checkable is in downtime"
 
 	case "downtime_end":
-		ev.Mute = types.MakeBool(false)
 		if h.HasBeenCancelled.Valid && h.HasBeenCancelled.Bool {
-			ev.Type = event.TypeDowntimeRemoved
```
oxzi, line 226: While the old `Event.Mute` and `Event.Type` fields were removed, the new `Event.Muted`/`Event.Incident` fields are still missing. This should result in an invalid `Event`.

https://github.com/Icinga/icinga-go-library/blob/5ed29e82a1a7e6bfbc2468d3792e6f942db9bf5e/notifications/event/event.go#L189-L191

---

Unless I am mistaken, similar issues exist in `buildFlappingHistoryEvent` and `buildAcknowledgementHistoryEvent` as well. These fields seem to be only set in the new `buildStateEvent` method.

pkg/notifications/notifications.go
```diff
@@ -307,103 +320,136 @@ func (client *Client) buildAcknowledgementHistoryEvent(ctx context.Context, h *v
...
 
-			err := ev.completeAndUpdate(client.ctx, attributes)
-			if err != nil {
-				client.logger.Errorw("Cannot fetch required attribute for event",
+	attributes := client.DefaultRelations
+	return retry.WithBackoff(
+		ctx,
+		func(ctx context.Context) (err error) {
+			for {
+				if err := ev.completeAndUpdate(ctx, attributes); err != nil {
```
oxzi, line 406: With this PR, the timeout-bound context was removed from the `completeAndUpdate` method. As it is now called within an `retry.WithBackoff` wrapper, this makes totally sense. However, there is no context termination anymore, allowing the `complete` callback function to block. Ftr, the `retry.Settings.Timeout` has no effect on the context or the inner function.

pkg/notifications/notifications.go
```diff
@@ -307,103 +320,136 @@ func (client *Client) buildAcknowledgementHistoryEvent(ctx context.Context, h *v
...
-	if client.ctx.Err() != nil {
-		client.logger.Errorw("Cannot process submitted entity as client context is done", zap.Error(client.ctx.Err()))
-		return true
-	}
-
+// blocking fashion and will be retried with an exponential backoff in case of retryable errors until a non-retryable
+// error occurs (like ctx cancellation) or the deadline is exceeded. In other words, when this method only returns
+// an error, then it usually means that there's nothing it can do anymore to successfully submit the event, thus it
+// should be treated as a fatal error.
+func (client *Client) Submit(ctx context.Context, entity database.Entity) error {
```
oxzi, line 336: First, I was confused why you changed the return type, as the only direct call handles it as an error. Only when used as an `RUUpsertFunc`, it is being checked. Maybe just document that this method satisfies this `type` for future code inspections.

pkg/icingadb/runtime_updates.go
```diff
@@ -57,50 +56,171 @@ func (r *RuntimeUpdates) ClearStreams(ctx context.Context) (config, state redis.
...
-	ctx context.Context, factoryFuncs []database.EntityFactoryFunc, streams redis.Streams, allowParallel bool,
+	ctx context.Context,
+	factoryFuncs []database.EntityFactoryFunc,
+	streams redis.Streams,
+	options ...RUOption,
 ) error {
-	g, ctx := errgroup.WithContext(ctx)
+	if len(streams) > 1 {
+		return errors.New("too many Redis streams provided, expected at most 1")
+	}
```
oxzi: Better check for `== 1` as this is the expected behavior. Later, the code expects one element to be present, resulting in a `panic` otherwise.

https://github.com/Icinga/icingadb/blob/84df82f23f901c63dfbce60f3b8a1aa3d7d74d20/pkg/icingadb/runtime_updates.go#L300

pkg/icingadb/runtime_updates.go
```diff
@@ -146,97 +276,43 @@ func (r *RuntimeUpdates) Sync(
...
-func (r *RuntimeUpdates) xRead(ctx context.Context, updateMessagesByKey map[string]chan<- redis.XMessage, streams redis.Streams) func() error {
+func (r *RuntimeUpdates) xRead(
+	ctx context.Context,
+	updateMessagesByKey map[string]chan<- redis.XMessage,
+	acknowledgementOutCh chan<- string,
+	streams redis.Streams,
+) func() error {
 	return func() error {
 		defer func() {
+			defer close(acknowledgementOutCh)
```
oxzi: Might be pedantic, but a `defer` in a `defer` in a returned `func` in a method is kinda deep. Couldn't you just move the `close()` call after the for` loop in the `defer` function?

pkg/icingadb/runtime_updates.go
```diff
@@ -146,97 +276,43 @@ func (r *RuntimeUpdates) Sync(
...
-			return r.db.NamedBulkExec(
-				ctx, cvFlatStmt, cvFlatCount, sem, flatCustomvars,
-				database.SplitOnDupId[database.Entity], database.OnSuccessIncrement[database.Entity](&counter),
-				database.OnSuccessIncrement[database.Entity](&telemetry.Stats.Config),
-			)
-		})
+	// Since all xRead goroutines are going to consume messages from the same stream independently, we are only
+	// allowed to send XDel commands after all of them have processed the messages and sent the IDs to be deleted
+	// to the xRedisMessageAcks channels. This way, they can move forward independently without waiting for the
+	// others to finish processing the same batch of messages.
```
oxzi: Unsure where exactly to place this comment, thus don't consider these lines too much.

With this design, `xRead` results in entities to appear in the `upsertEntities` channel for processing by the `opts.upsertFn`, where they might fail. However, during - or even before - `upsertFn` callback failure, `r.redis.XelOnAllConsumersAck` is already called, deleting the entry.

In the comment above, you are talking about "XDEL after processed", but it is more like dispatched.

Thus, my concern is, that a failing `upsertFn` - being most likely `notifications.Client.Submit` - might result in a lost notification.

pkg/notifications/notifications.go
```diff
@@ -143,36 +138,52 @@ func (client *Client) buildCommonEvent(
...
-	}
-
-	return ev, nil
+	}, nil
 }
 
+// errNonVolatileNonHardState is returned when a non-hard state change is attempted to be submitted for a non-volatile checkable.
+var errNonVolatileNonHardState = errors.New("non-hard state change for non-volatile checkable")
+
 // buildStateHistoryEvent builds a fully initialized event.Event from a state history entry.
```
oxzi: API doc requires an update, old method name.

pkg/notifications/notifications.go
```diff
@@ -143,36 +138,52 @@ func (client *Client) buildCommonEvent(
...
 	if err != nil {
-		return nil, errors.Wrapf(err, "cannot build event for %q,%q", h.HostId, h.ServiceId)
+		return nil, errors.Wrapf(err, "cannot build event for %q,%q", hostId, serviceId)
+	}
+
+	if s.Output.Valid {
+		ev.Message = s.Output.String
+	}
+	if s.LongOutput.Valid {
+		ev.Message += "\n" + s.LongOutput.String
```
oxzi, line 168: Unlikely, but if `Output` is invalid and `LongOutput` is valid, there would be a leading newline. (Is this even possible?)

pkg/notifications/notifications.go
```diff
@@ -307,103 +320,136 @@ func (client *Client) buildAcknowledgementHistoryEvent(ctx context.Context, h *v
...
 		eventErr error
 	)
 
+	canIgnoreStateUpdate := func(s *v1.State) bool {
+		// Ignore PENDING -> OK, otherwise we'll have a bunch of incidents that are be closed immediately.
+		// Also ignore any Pending states (99), as these are not relevant for notifications.
+		return s.HardState == 99 || (s.HardState == 0 && s.PreviousHardState == 99)
+	}
+
 	// Keep the type switch in sync with the values of SyncKeyStructPtrs below.
```
oxzi: Independent of this PR: This comment seems to be outdated. There is no longer `SyncKeyStructPtrs`, but only `syncKeyStructPtrs` as a local variable within `Client.SyncExtraStages`.

With the changes of this PR, they are not even in sync any more, as there are now also state changes.

pkg/notifications/notifications.go
```diff
@@ -182,24 +193,50 @@ func (client *Client) buildStateHistoryEvent(ctx context.Context, h *v1history.S
...
+	isAcked := s.IsAcknowledged.Valid && s.IsAcknowledged.Bool
+	isFlapping := s.IsFlapping.Valid && s.IsFlapping.Bool
+	ev.Muted = types.MakeBool(inDowntime || isAcked || isFlapping)
+	if ev.IsMuted() {
+		ev.MutedReason = "Checkable is muted due to"
+		if inDowntime {
+			ev.MutedReason += " currently active downtime"
+		}
+		if isAcked && inDowntime {
+			ev.MutedReason += ", and acknowledgement"
```
oxzi:
```suggestion
			ev.MutedReason += ", and an acknowledgement"
```

pkg/notifications/notifications.go
```diff
@@ -182,24 +193,50 @@ func (client *Client) buildStateHistoryEvent(ctx context.Context, h *v1history.S
...
+		} else if isAcked {
+			ev.MutedReason += " an acknowledgement"
+		}
+		if isFlapping && (inDowntime || isAcked) {
+			ev.MutedReason += ", and flapping as well."
+		} else if isFlapping {
+			ev.MutedReason += " flapping state"
+		}
```
oxzi: Consider removing the trailing period from the `if` case and add one at the end regardless of any branch, as currently only this very branch creates a sentence with ending with a period.

oxzi (commented):
First thing first: This is a big change and it looks solid at first glance. I went over the code multiple times and tried only to highlight relevant things. However, due to the complexity, please take some of my comments with a grain of salt.

pkg/notifications/notifications.go
yhabteab: As already states in the PR descpription these events aren't used anymore, so I don't currently care whether the resulting event is valid or not because they are disabled. I've removed these lines simply because to be able to build Icinga DB without errors.
https://github.com/Icinga/icingadb/blob/84df82f23f901c63dfbce60f3b8a1aa3d7d74d20/cmd/icingadb/main.go#L189-L194

pkg/notifications/notifications.go
yhabteab: > However, there is no context termination anymore, allowing the `complete` callback function to block. Ftr, the `retry.Settings.Timeout` has no effect on the context or the inner function.

I don't understand what you're trying to say with this! What kind of context termination would be relevant here? If the context gets cancelled somewhere else, then all its users will just fail as usual. Just as it states in the comment within that if statement, `completeAndUpdate` and thus, the `complete` function already retries all retryable errors internally with `retry.WithBackoff`, which also uses the same `retry.DefaultTimeout` settings as this one. So, there's no point adding any fake context deadlines for this because that's not needed anymore.

pkg/notifications/notifications.go
yhabteab: See https://github.com/Icinga/icingadb/pull/1130#discussion_r3472527496

pkg/notifications/notifications.go
yhabteab: Just removed that comment entirely.

pkg/icingadb/runtime_updates.go
yhabteab: > Thus, my concern is, that a failing `upsertFn` - being most likely `notifications.Client.Submit` - might result in a lost notification.

I was thinking that since `notifications.Client.Submit` will always retry the request until it either succeeds or crashes fatally (thus, there won't be lost notifications), I didn't neeed to wait for it and can delete the message immediately. Even considering how the HA behavior will be changed in a follow-up PR, I think it still makes sense to keep this that way as every HA take over atomatically clears the runtime update streams before starting to actually process any events. Of course, just like the Icinga DB' initial config dump there will also be an initial Icinga Notifications state sync on daemon restart and HA take overs (will also be implemetend in a follow-up PR, as part of #1100, I guess), so nothing should get lost.

pkg/notifications/notifications.go
yhabteab: Added a note.

pkg/icingadb/runtime_updates.go
yhabteab: Fixed.

pkg/icingadb/runtime_updates.go
yhabteab: Removed the extra defer.

pkg/icingadb/runtime_updates.go
yhabteab: I didn't change the overall design of this but only the way how the messages are processed. For example, the upsertEntities channel was buffered for the `uspertFn` case as well, which is unnecessary, so I've changed it to be a non-buffered channel now. Making the channels unbuffered has implicitly resolved your concerns!

pkg/notifications/notifications.go
yhabteab: Fixed.

pkg/notifications/notifications.go
yhabteab: Fixed.

pkg/notifications/notifications.go
yhabteab: Changed.

yhabteab:
Addressed all your requested changes unless otherwise commented to the respective review comments.

pkg/notifications/notifications.go
oxzi: > I don't understand what you're trying to say with this!

I was sitting less than two meters away from you. You could have just asked me :)

---

I was referring to the removed `context.WithTimeout` in the method.

https://github.com/Icinga/icingadb/pull/1130/changes#diff-f8656d2cdeaef2a8c20cd230a394a023d121771870063579ee0a8b5d11343b85L37-L39

pkg/notifications/notifications.go
oxzi: Sounds fair. However, would you mind adding a one line note to those methods or just call a `panic("TODO: implement")` to ensure this won't be forgotten in the future. In this state, we are having fine looking code, which cannot work and might result in future headaches.

pkg/icingadb/runtime_updates.go
oxzi: Thanks for your considerations :)

pkg/notifications/notifications.go
yhabteab: These functions will now panic, but since `go vet` didn't allow me to place the panic on the top of the function (due to unreachable code errors), I've deferred the panic all of these function, so it looks kinda weird.

pkg/notifications/notifications.go
oxzi: > > I don't understand what you're trying to say with this!
> 
> I was sitting less than two meters away from you. You could have just asked me :)

After we just did this, I would close this conversation as the code makes sense. Thanks.

oxzi (approved):
⛵

PR #1150, opened 2026-07-14
notifications: use always `time.Now()` for event IDs

yhabteab:
Previously, the LastStateChange column was used to generate the event ID, but this is not reliable for events other than state changes, as the checkable can be ACKED, set in Downtime etc. while still being in the same problem state, thus the LastStateChange timestamp does not change. So, we now use the current timestamp in milliseconds to generate a unique event ID for each event, as everything sent to Icinga Notifications is considered a new event, even if the checkable is still in the same state.

I've just merged #1148 without actually testing it, I only noticed this again when I was testing the HA related in Icinga Notifications.

### DIFF
--- a/pkg/notifications/notifications.go
+++ b/pkg/notifications/notifications.go
@@ -285,7 +285,6 @@ func (client *Client) fetchIncidents(ctx context.Context) error {
 func (client *Client) buildCommonEvent(
 	ctx context.Context,
 	hostId, serviceId types.Binary,
-	eventTime types.UnixMilli,
 ) (*fetchableEvent, error) {
 	rel, err := client.fetchHostServiceData(ctx, hostId, serviceId)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/notifications/notifications.go
```diff
@@ -327,7 +326,7 @@ func (client *Client) buildCommonEvent(
 		}
 	}
 
-	objectId := objectName + fmt.Sprintf("!%d", eventTime.Time().UnixMilli())
+	objectId := objectName + fmt.Sprintf("!%d", time.Now().UnixMilli())
```
oxzi, line 329: Unless I am mistaken, this only works and does not result in duplicate events with different IDs in an HA scenario as since #1130 the `notifications.Client.Submit` method is only called by the HA-active node.

Once the following lines get altered, this might result in duplicate events, right?

https://github.com/Icinga/icingadb/blob/95e03e1d2fa853b5d505e6b1f4cb010a296e24a5/cmd/icingadb/main.go#L190-L194

Thus, I am fine for the moment, but please consider adding a bit warning here and at the commented out lines in the main.

yhabteab: > Once the following lines get altered, this might result in duplicate events, right?

No, there won't be any duplicates as we are never going to process state changes from the history streams anymore. Generally, Icinga DB won't ever send duplicate events regardless of the ID timestamp, that's it already sync all states at daemon startup and then it send all events consumed from the runtime updates as they come, and these events are always unique.

oxzi (approved):

Issue #1113, opened 2026-04-20
Support additional connection mechanisms to Icinga Notifications


nilmerg:
Icinga Notifications will add support for the following additional mechanisms:

**Authentication**

* Client certificates

**Protocol**

* HTTPs

**Transport**

* Unix Domain Sockets

Icinga DB must allow to use these and allow configuring them.

Issue #1108, opened 2026-04-08
Load Notifications configuration from the database


nilmerg:
Icinga DB Web will allow the user to configure the notification integration in the future. In order to do so, it will need to dump this configuration into the database so that Icinga Notifications can read and use it. Though, in case Icinga Notifications has been configured by local configuration, the particular settings need to be locked/flagged so that Icinga DB Web can show the user that changes need to be made locally only.

### Database

As mentioned in #1099 a new table `icingadb_config` is required. The same suggestion made there regarding its structure applies here, a key-value table. Though, for Icinga DB Web to know that particular settings are locked an additional column is required. `locked` using the usual `bool_enum` type seems to be good fit. Icinga DB should then insert/update rows in this table matching the local notification configuration and lock them.

### Reload Behavior

Icinga Notifications already detects changes to the sources configuration automatically at runtime so my suggestion for Icinga DB is to also detect changes to its notification configuration in the database automatically. The main reason for this isn't to detect changes but rather that the initial setup has been completed in Icinga DB Web. It's Icinga DB Web's responsibility to ensure that the configuration of Icinga Notifications succeeded as well, but Icinga DB still has to respect a given grace period before giving up in case the connection isn't accepted.

oxzi:
As the issue drafts it, the key-value pairs in the `icingadb_config` table lack clear ownership. Consider the following scenarios:

- Icinga DB Web configures a value. Afterward, Icinga DB starts up with a static configuration. Should Icinga DB now overwrite the database row with a `locked` flag?
- Icinga DB Web configured something. Icinga DB starts up with a static configuration and overwrites it in the database, including the `locked` flag. Then, the static configuration was removed. Icinga DB restarted and automatically reloaded its old configuration, still holding the `locked` flag in the database. How can this flag be removed?
- HA Setup: There are two Icinga DB nodes, both with a static configuration. If the configurations differ, which might be the case for the Icinga Notifications API endpoint, for example, which configuration would take precedence?[^-1]

After thinking about the implementation, I would argue that a simple table without ownership cannot work in an environment with multiple producers and producer-consumer hybrids.

So, what if we drop the entire "locked" logic and prevent Icinga DB from writing its configuration[^0] to the database? Web would be the only producer, and Icinga DB would load whatever Web wrote in the table but would prefer its static configuration.

[^-1]: This could even allow one Icinga DB node to re-configure the other one, if only one has certain keys set in its static configuration.
[...]

PR #1151, opened 2026-07-16
Probe notifications socket & monitor API health

### LINKED ISSUE #1141: Notifications configuration discovery and health monitoring
Even though Icinga Notifications will add high availability with v1.0, coupling more than one with Icinga DB will not be supported in the foreseeable future. But with the notification component being a very integral part of monitoring itself, we have to ensure it works seamlessly in our recommended HA architecture.

## HA Responsibility

In case the connection to Icinga Notifications cannot be established, Icinga DB must pass on responsibility to another instance.

## Health Monitoring

The database table `icingadb_instance` is already used by Icinga DB Web to identify the `responsible` instance and verifies its `heartbeat`. What I'd like to see is a flag in the same table that allows Web to verify a working notification component.

## Configuration Discovery

One aspect of our recommendation is how Icinga DB connects to Icinga Notifications: by a unix domain socket.
Our packages will set this up by default and so it's always the same path. This brought me to the idea that Icinga DB may also indicate the availability of said socket path inside `icingadb_instance` so that Web can ensure the connection will work. Since the process needs access to the socket so see it, it's username should also be reported by this and written to the database so that the UI has a guarantee that both settings are valid values.

## Goals

This should allow Web to do the following:

* Indicate in the *Health* section whether the notification component is active
* Provide a simple toggle to enable/disable the notification component (as otherwise it will require the user to enter an address and a username)

yhabteab:
This probes the Icinga Notifications listener socket availability once on startup and persists the result along with the Icinga DB service user to the `icingadb_instance` table. Also, when the Notifications component is indeed enabled, then it will monitor the Icinga Notifications API availability and persists the result into that very same table as well. If Icinga Notifications isn't available for some reason, then the current Icinga DB isntance will give up HA responsibilty after ~5m timeout, but if we don't have HA setup, then it will just retry the Icinga Notifications request endlessly and will never crash.

## TODO

- [x] Test it in HA setup (I've only tested this in a non-HA setup so far)
- [x] Agree on the newly introduced database column names (/cc @nilmerg)

## Testing

**Icinga DB 1:**

```bash
icingadb-1  | 2026-07-17T09:15:52.965+0200	INFO	high-availability	Another instance is active	{"instance_id": "c13c95aed1584ef49bd0df848c86c79c", "environment": "0162e462daea7c02249e1995a7fff030c776380f", "heartbeat": "2026-07-17T09:15:51.786+0200", "heartbeat_age": "1.179567057s"}
icingadb-1  | 2026-07-17T09:17:04.867+0200	INFO	icingadb	Taking over	{"reason": "no other instance is active"}
icingadb-1  | 2026-07-17T09:17:04.868+0200	INFO	icingadb	Starting config sync
icingadb-1  | 2026-07-17T09:17:04.868+0200	INFO	icingadb	Starting overdue sync
icingadb-1  | 2026-07-17T09:17:04.871+0200	INFO	icingadb	Starting initial state sync
icingadb-1  | 2026-07-17T09:17:05.054+0200	INFO	config-sync	Updating 2 items of type host state
icingadb-1  | 2026-07-17T09:17:05.082+0200	INFO	config-sync	Updating 24 items of type service state
icingadb-1  | 2026-07-17T09:17:05.757+0200	INFO	config-sync	Finished config sync in 889.91269ms
icingadb-1  | 2026-07-17T09:17:05.757+0200	INFO	icingadb	Starting config runtime updates sync
icingadb-1  | 2026-07-17T09:17:07.648+0200	INFO	notifications	Fetching 2 entities of type HostState from Redis for submission to Icinga Notifications
icingadb-1  | 2026-07-17T09:17:07.648+0200	INFO	notifications	Fetching 24 entities of type ServiceState from Redis for submission to Icinga Notifications
icingadb-1  | 2026-07-17T09:17:07.668+0200	INFO	icingadb	Starting state runtime updates sync
icingadb-1  | 2026-07-17T09:17:07.668+0200	INFO	icingadb	Starting history retention
icingadb-1  | 2026-07-17T09:17:07.668+0200	INFO	config-sync	Finished initial state sync in 2.800652177s
```

**Icinga DB 2:** It gave up HA responsibility after exactly 5 minutes the first time Icinga Notifications was identified as unhealthy.

```bash
icingadb-master2-1  | 2026-07-17T09:12:02.032+0200	ERROR	notifications	Cannot submit event to Icinga Notifications	{"event": "icinga-master2.devlab.com!disk /", "attempt": 1, "elapsed": "46.095872ms", "error": "cannot POST HTTP request to process event: cannot do HTTP request: Post \"http://notifications-master2:5680/process-event\": dial tcp: lookup notifications-master2 on 127.0.0.11:53: no such host", "errorVerbose": "Post \"http://notifications-master2:5680/process-event\": dial tcp: lookup notifications-master2 on 127.0.0.11:53: no such host\ncannot do HTTP request\ngithub.com/icinga/icinga-go-library/notifications/source.(*Client).doRequest\n\t/go/src/github.com/Icinga/icinga-go-library/notifications/source/client.go:340\ngithub.com/icinga/icinga-go-library/notifications/source.(*Client).ProcessEvent\n\t/go/src/github.com/Icinga/icinga-go-library/notifications/source/client.go:111\ngithub.com/icinga/icingadb/pkg/notifications.(*Client).Submit.func2\n\t/go/src/github.com/Icinga/icingadb/pkg/notifications/notifications.go:605\ngithub.com/icinga/icinga-go-library/retry.WithBackoff\n\t/go/src/github.com/Icinga/icinga-go-library/retry/retry.go:65\ngithub.com/icinga/icingadb/pkg/notifications.(*Client).Submit\n\t/go/src/github.com/Icinga/icingadb/pkg/notifications/notifications.go:593\ngithub.com/icinga/icingadb/pkg/icingadb.(*RuntimeUpdates).Sync.func2\n\t/go/src/github.com/Icinga/icingadb/pkg/icingadb/runtime_updates.go:216\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\t/go/pkg/mod/golang.org/x/sync@v0.22.0/errgroup/errgroup.go:93\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_arm64.s:1447\ncannot POST HTTP request to process event\ngithub.com/icinga/icinga-go-library/notifications/source.(*Client).ProcessEvent\n\t/go/src/github.com/Icinga/icinga-go-library/notifications/source/client.go:124\ngithub.com/icinga/icingadb/pkg/notifications.(*Client).Submit.func2\n\t/go/src/github.com/Icinga/icingadb/pkg/notifications/notifications.go:605\ngithub.com/icinga/icinga-go-library/retry.WithBackoff\n\t/go/src/github.com/Icinga/icinga-go-library/retry/retry.go:65\ngithub.com/icinga/icingadb/pkg/notifications.(*Client).Submit\n\t/go/src/github.com/Icinga/icingadb/pkg/notifications/notifications.go:593\ngithub.com/icinga/icingadb/pkg/icingadb.(*RuntimeUpdates).Sync.func2\n\t/go/src/github.com/Icinga/icingadb/pkg/icingadb/runtime_updates.go:216\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\t/go/pkg/mod/golang.org/x/sync@v0.22.0/errgroup/errgroup.go:93\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_arm64.s:1447"}
icingadb-master2-1  | 2026-07-17T09:12:12.783+0200	INFO	history-sync	Synced 10 state history items
icingadb-master2-1  | 2026-07-17T09:12:15.920+0200	INFO	runtime-updates	Upserted 7 ServiceState items
icingadb-master2-1  | 2026-07-17T09:12:32.783+0200	INFO	history-sync	Synced 2 state history items
icingadb-master2-1  | 2026-07-17T09:12:35.920+0200	INFO	runtime-updates	Upserted 3 ServiceState items
icingadb-master2-1  | 2026-07-17T09:12:52.783+0200	INFO	history-sync	Synced 1 state history items
icingadb-master2-1  | 2026-07-17T09:15:52.777+0200	INFO	high-availability	Continuing being the active instance	{"instance_id": "c13c95aed1584ef49bd0df848c86c79c", "environment": "0162e462daea7c02249e1995a7fff030c776380f"}
icingadb-master2-1  | 2026-07-17T09:17:04.670+0200	ERROR	high-availability	Icinga Notifications component has been unhealthy for a long time	{"unhealthy_since": "2026-07-17T09:12:02.032+0200"}
icingadb-master2-1  | 2026-07-17T09:17:04.672+0200	WARN	icingadb	Handing over	{"reason": "notifications component has been unhealthy for a long time"}
```

resolves #1141
refs https://github.com/Icinga/icingadb-web/issues/1356

### DIFF
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -181,11 +181,12 @@ ICINGADB_RETENTION_OPTIONS=comment:356
 
 ## Notifications Configuration
 
-!!! warning
+!!! tip
 
-    The Icinga Notifications integration is a feature preview that you can try out.
-    However, incompatible changes may happen, so be sure to check the changelog for future updates.
-    Please do not use it in production.
+    If the `ICINGA_NOTIFICATIONS_LISTENER_SOCKET` environment variable is set, Icinga DB will automatically detect the
+    Icinga Notifications API Unix Domain Socket availability and persist the result in the database. This allows
+    Icinga DB Web to provide a one-click configure button for Icinga Notifications, which will automatically configure
+    this component, and create the corresponding Icinga Notifications source configuration on your behalf.
 
 Icinga DB can act as an event source for [Icinga Notifications](https://icinga.com/docs/icinga-notifications/).
 If configured, Icinga DB will submit events to the Icinga Notifications API.
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -18,6 +18,10 @@ import (
 	icingaredisv1 "github.com/icinga/icingadb/pkg/icingaredis/v1"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
+	"golang.org/x/sys/unix"
+	"os"
+	"os/user"
+	"path/filepath"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -32,6 +36,25 @@ type haState struct {
 	responsibleTsMilli int64
 	responsible        bool
 	otherResponsible   bool
+}
+
+// NotificationsState holds the current state of the Icinga Notifications component.
+type NotificationsState struct {
+	unhealthySince time.Time // The time when the Icinga Notifications component was first detected as unhealthy.
+	healthy        types.Bool
+
+	// The discovered Icinga Notifications listener socket path, if any.
+	socketPath types.String
+
+	// lastProblemLoggedAt is the last time a problem with the Icinga Notifications component was logged.
+	// This is used to suppress recurring log messages.
+	lastProblemLoggedAt time.Time
+}
+
+// IsValidUnhealthy returns true if the healthy state is valid and set to false, indicating that the
+// Icinga Notifications component is unhealthy.
+func (ns *NotificationsState) IsValidUnhealthy() bool {
+	return ns.healthy.Valid && !ns.healthy.Bool
 }
 
 // HA provides high availability and indicates whether a Takeover or Handover must be made.
@@ -52,6 +75,13 @@ type HA struct {
 	errOnce       sync.Once
 	errMu         sync.Mutex
 	err           error
+
+	// notificationsHeartbeat is a channel that signals the status of the Icinga Notifications component.
+	//
+	// This will only receive events when the Icinga Notifications component is enabled, and its true value indicates
+	// that the component is healthy and able to process notifications, while a false value indicates that it is not.
+	notificationsHeartbeatCh chan bool
+	notifications            NotificationsState
 }
 
 // NewHA returns a new HA and starts the controller loop.
@@ -70,6 +100,8 @@ func NewHA(ctx context.Context, db *database.DB, heartbeat *icingaredis.Heartbea
 		handover:   make(chan string),
 		takeover:   make(chan string),
 		done:       make(chan struct{}),
+
+		notificationsHeartbeatCh: make(chan bool),
 	}
 
 	ha.state.Store(&haState{})
@@ -122,6 +154,12 @@ func (h *HA) Takeover() chan string {
 	return h.takeover
 }
 
+// NotificationsHeartbeat returns a channel to send the status of Icinga Notifications component into.
+//
+// The channel is used to signal the health of the Icinga Notifications component, where a true value indicates
+// that the component is healthy and able to process notifications, while a false value indicates that it is not.
+func (h *HA) NotificationsHeartbeat() chan<- bool { return h.notificationsHeartbeatCh }
+
 // State returns the status quo.
 func (h *HA) State() (responsibleTsMilli int64, responsible, otherResponsible bool) {
 	state := h.state.Load()
@@ -173,6 +211,12 @@ func (h *HA) controller() {
 
 					continue
 				}
+
+				// Returns true when a handover has been triggered, and we should skip the rest of the loop.
+				if h.verifyNotificationsState() {
+					continue
+				}
+
 				s, err := m.Stats().IcingaStatus()
 				if err != nil {
 					h.abort(err)
@@ -237,6 +281,16 @@ func (h *HA) controller() {
 				h.signalHandover("lost heartbeat")
 				h.realizeLostHeartbeat()
 			}
+
+		case healthy := <-h.notificationsHeartbeatCh:
+			h.logger.Debugw("Received notifications heartbeat", zap.Bool("healthy", healthy))
+			h.notifications.healthy = types.MakeBool(healthy)
+			if healthy {
+				h.notifications.unhealthySince = time.Time{}
+			} else if h.notifications.unhealthySince.IsZero() {
+				h.notifications.unhealthySince = time.Now()
+			}
+
 		case <-h.heartbeat.Done():
 			if err := h.heartbeat.Err(); err != nil {
 				h.abort(err)
@@ -365,6 +419,9 @@ func (h *HA) realize(
 				Icinga2FlapDetectionEnabled:       s.FlapDetectionEnabled,
 				Icinga2PerformanceDataEnabled:     s.PerformanceDataEnabled,
 				IcingadbVersion:                   internal.Version.Version,
+				IcingadbServiceUser:               ServiceUser(),
+				NotificationsHealthy:              h.notifications.healthy,
+				NotificationsDiscoveredSocketPath: h.probeNotificationsSocketPath(),
 			}
 
 			stmt, _ := h.db.BuildUpsertStmt(i)
@@ -480,10 +537,21 @@ func (h *HA) realize(
 	return nil
 }
 
+// instanceCount returns the number of instances in the current environment.
+func (h *HA) instanceCount() int64 {
+	query := h.db.Rebind("SELECT COUNT(*) FROM icingadb_instance WHERE environment_id = ?")
+	var count int64
+	if err := h.db.GetContext(h.ctx, &count, query, h.environment.Id); err != nil {
+		h.logger.Warnw("Cannot count instances", zap.Error(database.CantPerformQuery(err, query)))
+		return 0
+	}
+	return count
+}
+
 // realizeLostHeartbeat updates "responsible = n" for this HA into the database.
 func (h *HA) realizeLostHeartbeat() {
-	stmt := h.db.Rebind("UPDATE icingadb_instance SET responsible = ? WHERE id = ?")
-	if _, err := h.db.ExecContext(h.ctx, stmt, "n", h.instanceId); err != nil && !utils.IsContextCanceled(err) {
+	stmt := h.db.Rebind("UPDATE icingadb_instance SET responsible = ?, notifications_healthy = ? WHERE id = ?")
+	if _, err := h.db.ExecContext(h.ctx, stmt, "n", h.notifications.healthy, h.instanceId); err != nil && !utils.IsContextCanceled(err) {
 		h.logger.Warnw("Can't update instance", zap.Error(database.CantPerformQuery(err, stmt)))
 	}
 }
@@ -557,5 +625,82 @@ func (h *HA) signalTakeover(reason string) {
 		case <-h.ctx.Done():
 			// Noop
 		}
+	}
+}
+
+// verifyNotificationsState checks the state of the Icinga Notifications component and triggers a handover if necessary.
+//
+// It returns true if a handover was triggered, false otherwise.
+func (h *HA) verifyNotificationsState() bool {
+	now := time.Now()
+	if h.notifications.IsValidUnhealthy() && now.After(h.notifications.unhealthySince.Add(retry.DefaultTimeout)) {
+		lastLogged := h.notifications.lastProblemLoggedAt
+		if lastLogged.IsZero() || now.After(lastLogged.Add(retry.DefaultTimeout)) {
+			h.logger.Errorw("Icinga Notifications component has been unhealthy for a long time",
+				zap.Time("unhealthy_since", h.notifications.unhealthySince))
+		}
+		h.notifications.lastProblemLoggedAt = now
+
+		// If we're indeed running in a multi-instance environment, we should hand over responsibility
+		// to another instance, otherwise there's no other instance to take over, so we let this retry
+		// its connection attempts to the Icinga Notifications API.
+		if h.instanceCount() > 1 {
+			h.signalHandover("notifications component has been unhealthy for a long time")
+			h.realizeLostHeartbeat()
+
+			// Reset the notifications state, so that we have a fresh start when this instance regains responsibility.
+			h.notifications = NotificationsState{}
+
+			return true
+		}
+	} else {
+		h.notifications.lastProblemLoggedAt = time.Time{}
+	}
+
+	return false
+}
+
+// probeNotificationsSocketPath checks the availability of the Icinga Notifications listener socket and returns
+// its path if available.
+func (h *HA) probeNotificationsSocketPath() (socketPath types.String) {
+	if h.notifications.socketPath.Valid && h.notifications.socketPath.String != "" {
+		return h.notifications.socketPath
+	}
+	defer func() { h.notifications.socketPath = socketPath }()
+
+	path := os.Getenv("ICINGA_NOTIFICATIONS_LISTENER_SOCKET")
+	if path == "" {
+		h.logger.Debugw("Cannot probe Icinga Notifications listener socket availability: ICINGA_NOTIFICATIONS_LISTENER_SOCKET is not set")
+		return types.MakeString("", types.TransformEmptyStringToNull)
+	}
+
+	path = filepath.Clean(path)
+	info, err := os.Stat(path)
+	if err != nil {
+		h.logger.Warnw("Failed to stat Icinga Notifications listener socket", zap.String("path", path), zap.Error(err))
+		return types.MakeString("", types.TransformEmptyStringToNull)
+	}
+
+	if info.Mode()&os.ModeSocket == 0 {
+		h.logger.Warnw("Icinga Notifications listener socket path is not a socket", zap.String("path", path))
+		return types.MakeString("", types.TransformEmptyStringToNull)
+	}
+
+	if err := unix.Access(path, unix.R_OK|unix.W_OK); err != nil {
+		h.logger.Warnw("Icinga DB user has no permissions to access Icinga Notifications listener socket",
+			zap.String("path", path), zap.Error(err))
+		return types.MakeString("", types.TransformEmptyStringToNull)
+	}
+
+	h.logger.Debugw("Successfully probed Icinga Notifications listener socket availability", zap.String("path", path))
+	return types.MakeString(path)
+}
+
+// ServiceUser returns the name the service user the process is running as.
+func ServiceUser() string {
+	if currentUser, err := user.Current(); err != nil {
+		return "unknown"
+	} else {
+		return currentUser.Username
 	}
 }
--- a/pkg/icingadb/v1/icingadb_instance.go
+++ b/pkg/icingadb/v1/icingadb_instance.go
@@ -19,4 +19,7 @@ type IcingadbInstance struct {
 	Icinga2FlapDetectionEnabled       types.Bool      `json:"icinga2_flap_detection_enabled"`
 	Icinga2PerformanceDataEnabled     types.Bool      `json:"icinga2_performance_data_enabled"`
 	IcingadbVersion                   string          `json:"-"`
+	IcingadbServiceUser               string          `json:"-"`
+	NotificationsHealthy              types.Bool      `json:"-"`
+	NotificationsDiscoveredSocketPath types.String    `json:"-"`
 }
--- a/pkg/notifications/notifications.go
+++ b/pkg/notifications/notifications.go
@@ -77,6 +77,9 @@ type Client struct {
 	// incidentsByObjId is a map of object IDs to incidents populated by the first call to ApplyDelta.
 	incidentsByObjId map[string]source.Incident
 	incidentsMu      sync.Mutex
+
+	// heartbeatOutCh is a channel used to send heartbeat signals to the HA controller.
+	heartbeatOutCh chan<- bool
 }
 
 // NewNotificationsClient creates a new Client connected to an existing database and logger.
@@ -85,6 +88,7 @@ func NewNotificationsClient(
 	rc *redis.Client,
 	logger *logging.Logger,
 	cfg source.Config,
+	heartbeatOutCh chan<- bool,
 ) (*Client, error) {
 	notificationsClient, err := source.NewClient(cfg, "Icinga DB "+internal.Version.Version)
 	if err != nil {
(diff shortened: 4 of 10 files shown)


yhabteab:
Please have a look at the database schema changes.

nilmerg (changes requested):
Since #1113 isn't done yet, I'm not sure what type for `notifications_socket_path_discovered` is suitable. The most portable choice I think is text, as if the socket path isn't required for the configuration it is at least usable to present it to the user. Given that, I'd also suggest a different name: `notifications_discovered_socket_path`

yhabteab:
> Since #1113 isn't done yet, I'm not sure what type for `notifications_socket_path_discovered` is suitable. The most portable choice I think is text, as if the socket path isn't required for the configuration it is at least usable to present it to the user. Given that, I'd also suggest a different name: `notifications_discovered_socket_path`

I still didn't change that yet since Alvar is already working on the referenced issue and it might get merged before this, but also I still don't know what am I supposed to insert into that suggested text column (based on the suggested name, I guess the discovered path?). However, we already ruled out storing the path in the DB yesterday in our discussion, so I'm not sure why you changed your mind on this now.

Anyway, apart from that thing, everything else is ready-to-be-reviewed. I've already tested it in a real HA setup (see PR descs) and is working fine.

pkg/icingadb/v1/icingadb_instance.go
```diff
@@ -19,4 +19,7 @@ type IcingadbInstance struct {
 	Icinga2FlapDetectionEnabled       types.Bool      `json:"icinga2_flap_detection_enabled"`
 	Icinga2PerformanceDataEnabled     types.Bool      `json:"icinga2_performance_data_enabled"`
 	IcingadbVersion                   string          `json:"-"`
+	IcingadbServiceUser               string          `json:"-"`
```
oxzi, line 22: Pardon, but I don't really understand why we need this field? On a package install, this should be `icingadb` for all instances.

The referenced issue mentions the following:

> Since the (_Icinga DB_) process needs access to the socket so (_sic_) see it, it's (_sic_) username should also be reported by this and written to the database so that the UI has a guarantee that both settings are valid values.

Okay, but what should the UI do with this information? I would argue that the daemon user groups are more relevant, as a group permission on the socket will likely grant access. But even then, I am currently failing to understand the big "why".

pkg/icingadb/ha.go
```diff
@@ -480,10 +542,21 @@ func (h *HA) realize(
 	return nil
 }
 
+// instanceCount returns the number of instances in the current environment.
+func (h *HA) instanceCount() int64 {
+	query := h.db.Rebind("SELECT COUNT(*) FROM icingadb_instance WHERE environment_id = ?")
```
oxzi, line 542: This query would also return old instances not yet pruned via `HA.removeOldInstances`. So, there might be a duration where between a stopped or crashed Icinga DB and the next pruning, which happens ever 65 seconds.

Next to dead instances, not responsible active instances are counted as well. Considering how this method is used in `HA.verifyNotificationsState`, I am expecting a bailout in the `HA.controller` main for loop. Unless I am mistaken, this cannot recover when having this state initially, i.e., after a prior crash, as removing of old instances happens later in the same loop.

pkg/icingadb/ha.go
```diff
@@ -237,6 +280,22 @@ func (h *HA) controller() {
 				h.signalHandover("lost heartbeat")
 				h.realizeLostHeartbeat()
 			}
+
+		case healthy, ok := <-h.notificationsHeartbeat:
+			if !ok {
+				h.logger.Infof("Notifications heartbeat channel closed")
```
oxzi:
```suggestion
				h.logger.Info("Notifications heartbeat channel closed")
```

pkg/icingadb/ha.go
```diff
@@ -559,3 +632,79 @@ func (h *HA) signalTakeover(reason string) {
...
+// probeNotificationsSocketPathAvailability checks if the socket path is available, and we've permissions to access it.
+//
+// It returns a types.Bool indicating the availability of the socket path.
+func (h *HA) probeNotificationsSocketPathAvailability() (available types.Bool) {
+	if h.notifications.socketAvailable.Valid {
+		return h.notifications.socketAvailable
+	}
+	defer func() { h.notifications.socketAvailable = available }()
+
+	path := os.Getenv("ICINGA_NOTIFICATIONS_LISTENER_SOCKET")
```
oxzi, line 671: This should need documentation.

pkg/icingadb/ha.go
```diff
@@ -559,3 +632,79 @@ func (h *HA) signalTakeover(reason string) {
...
+		h.logger.Warnw("Icinga Notifications listener socket path is not a socket", zap.String("path", path))
+		return types.MakeBool(false)
+	}
+
+	file, err := os.OpenFile(path, os.O_RDWR, 0)
+	if err != nil {
+		h.logger.Warnw("Failed to open Icinga Notifications listener socket", zap.String("path", path), zap.Error(err))
+		return types.MakeBool(false)
+	}
+	_ = file.Close()
```
oxzi: I don't think we actually need to `open(2)`/`connect(2)` here, `access(2)` is enough. https://pkg.go.dev/golang.org/x/sys/unix#Access

```suggestion
	err := unix.Access(path, unix.R_OK|unix.W_OK)
	if err != nil {
		h.logger.Warnw("Failed to open Icinga Notifications listener socket", zap.String("path", path), zap.Error(err))
		return types.MakeBool(false)
	}
```

pkg/notifications/notifications.go
```diff
@@ -222,7 +223,10 @@ func (client *Client) ApplyDelta(ctx context.Context, delta *icingadb.Delta) err
 
 		attrs := source.ModifiableIncidentAttrs{Close: types.MakeBool(true)}
 		if err := client.notificationsClient.ModifyIncidents(ctx, attrs, filter); err != nil {
-			return errors.Wrap(err, "cannot close incidents for deleted objects")
+			client.logger.Errorw("Failed to bulk close obsolete incidents for deleted objects",
+				zap.String("entity", delta.Subject.Name()),
+				zap.Int("count", len(filter)),
+				zap.String("error", err.Error()))
```
oxzi, line 229:
```suggestion
				zap.Error(err))
```

Two more similar string errors follow below.

schema/mysql/schema.sql
```diff
@@ -550,6 +550,9 @@ CREATE TABLE icingadb_instance (
   icinga2_performance_data_enabled enum('n', 'y') NOT NULL,
 
   icingadb_version varchar(255) NOT NULL,
+  icingadb_service_user varchar(255) NOT NULL, -- useradd(8) on Debian 13 states "Usernames may only be up to 256 characters long."
```
oxzi: Based on <https://systemd.io/USER_NAMES/>, this answer is both right and wrong 🙃 

I have just tried to verify this. It seems, the "up to" is exclusive of the upper bound here. Thus, `varchar(255)` matches "up to 256".

```
root@45252ef835c5:/# useradd $(printf %256s | tr ' ' 'A')
useradd: invalid user name 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
```

So, I would keep the database type, but maybe change the comment, as it is very specific.

oxzi (commented):
Thanks, I have performed a first review, mainly focusing on the HA part. In general, all of this looks valid.

pkg/icingadb/v1/icingadb_instance.go
yhabteab: > But even then, I am currently failing to understand the big "why".

😄 ! Honestly, I don't know! My bad that I didn't question that either. When Icinga DB is using unix socket to connect to Icinga Notifications, then Icinga Notifications will get the service user of Icinga DB on its own, so I don't really know what this is needed for.

/cc @nilmerg

pkg/notifications/notifications.go
yhabteab: I changed those on purpose, since I don't want to include the ugly stack traces that aren't relevant for anything in the logs.

schema/mysql/schema.sql
yhabteab: This might not even be relevant anymore (https://github.com/Icinga/icingadb/pull/1151#discussion_r3601686977).

pkg/icingadb/ha.go
yhabteab: > This query would also return old instances not yet pruned via `HA.removeOldInstances`. So, there might be a duration where between a stopped or crashed Icinga DB and the next pruning, which happens ever 65 seconds.

AFIK, no it should not return dead instances. The old instances pruner runs 65 seconds after the first received Icinga 2 heartbeat and never again, i.e., it's only a one time thing. This query, on the other hand, is used only after a min timeout of 5 minutes, at which point the old instances pruner should have already run.

> Next to dead instances, not responsible active instances are counted as well. Considering how this method is used in `HA.verifyNotificationsState`, I am expecting a bailout in the `HA.controller` main for loop. Unless I am mistaken, this cannot recover when having this state initially, i.e., after a prior crash, as removing of old instances happens later in the same loop.

I don't understand what you're trying to say here. How do you identify _not responsible active instances_? As I explained above, old instances will get removed after 65s, and this method is only called after 5 minutes (if Icinga Notifications is unhealthy from the start on).

I'm not saying that this is a perfect solution, but it's good enough for my use case, but if you have a better idea on how this query should look like, please let me know because I don't have a better idea at the moment.

pkg/icingadb/v1/icingadb_instance.go
nilmerg: Folks, how does Notifications authenticate sources? And what is responsible to configure this in case of Icinga DB Web? See https://github.com/Icinga/icingadb-web/issues/1356 for details.

yhabteab: Arch, I was really confused by this. Of course, Icinga DB Web needs this to create the corresponding Icinga Notifications source.

yhabteab:
> Since #1113 isn't done yet, I'm not sure what type for `notifications_socket_path_discovered` is suitable.

That issue is now closed, and the decision on which transport to use is deduced from the url scheme. So, what's your final say on the type for `notifications_socket_path_discovered`?

pkg/icingadb/v1/icingadb_instance.go
oxzi: I considered that option, but as I stated initially, at least for every package install - which is the only really supported setup, iirc - this is always `icingadb`. When people deviate from this road, they are free to do so, but then they will be on their own.

We are going to need assumptions about the system anyway. With this PR, `ICINGA_NOTIFICATIONS_LISTENER_SOCKET` is set externally - most likely by the packaging in the systemd unit file - to connect to the Notifications API via a Unix Domain Socket. If this package-configured option is considered valid, why not also accept that the username will always be `icingadb`, which is set in the same packaging repo.

Nevertheless, maybe these two issues should get linked?

yhabteab: > When people deviate from this road, they are free to do so, but then they will be on their own.

Why? Just because they're using a different user to run Icinga DB as?

> If this package-configured option is considered valid, why not also accept that the username will always be `icingadb`, which is set in the same packaging repo.

That's an environment variable and can be set by users to point somewhere else too und it would still work because it's still unix socket. What's the problem you're seeing in persisting this in the ephemeral table that I don't currently see?

oxzi: I have no technical objections to the feature per se, I am simply questioning its use. And maybe my answer was a bit hard in tone, as a sarcastic response to an honest question was given by @nilmerg. Never mind though, go for it. All is fine from my side.

pkg/icingadb/ha.go
```diff
@@ -18,6 +18,10 @@ import (
 	icingaredisv1 "github.com/icinga/icingadb/pkg/icingaredis/v1"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
+	"golang.org/x/sys/unix"
```
oxzi, line 21: By implicitly importing `golang.org/x/sys/unix`, it moved up from the transitive `require` block to the direct `require` block in `go.mod`. `go mod tidy` ftw.

https://github.com/Icinga/icingadb/actions/runs/29582819542/job/87892394122?pr=1151

doc/03-Configuration.md
```diff
@@ -181,11 +181,12 @@ ICINGADB_RETENTION_OPTIONS=comment:356
...
-!!! warning
+!!! tip
 
-    The Icinga Notifications integration is a feature preview that you can try out.
-    However, incompatible changes may happen, so be sure to check the changelog for future updates.
-    Please do not use it in production.
+    If the `ICINGA_NOTIFICATIONS_LISTENER_SOCKET` environment variable is set, Icinga DB will automatically detect the
+    Icinga Notifications API socket availability and persit the result in the database. This allows Icinga DB Web to
+    provide a one-click configure button for Icinga Notifications, which will automatically configure this component,
+    and create the corresponding Icinga Notifications source configuration on your behalf.
```
oxzi: Thanks, but consider `s/socket/Unix Domain Socket/` as a socket could also be a TCP socket. At the moment, a file path to a unix domain socket is expected, which also differs from the `unix://` scheme used in `url` below.

pkg/icingadb/ha.go
```diff
@@ -559,3 +633,77 @@ func (h *HA) signalTakeover(reason string) {
...
+		return types.MakeBool(false)
+	}
+
+	if info.Mode()&os.ModeSocket == 0 {
+		h.logger.Warnw("Icinga Notifications listener socket path is not a socket", zap.String("path", path))
+		return types.MakeBool(false)
+	}
+
+	if err := unix.Access(path, unix.R_OK|unix.W_OK); err != nil {
+		h.logger.Warnw("Failed to open Icinga Notifications listener socket", zap.String("path", path), zap.Error(err))
```
oxzi:
```suggestion
		h.logger.Warnw("Icinga DB user has no permissions to open Icinga Notifications listener socket", zap.String("path", path), zap.Error(err))
```

pkg/icingadb/ha.go
```diff
@@ -559,3 +633,77 @@ func (h *HA) signalTakeover(reason string) {
...
+}
+
+// probeNotificationsSocketPathAvailability checks if the socket path is available, and we've permissions to access it.
+//
+// It returns a types.Bool indicating the availability of the socket path.
+func (h *HA) probeNotificationsSocketPathAvailability() (available types.Bool) {
+	if h.notifications.socketAvailable.Valid {
+		return h.notifications.socketAvailable
+	}
+	defer func() { h.notifications.socketAvailable = available }()
```
oxzi: This cache would result in a one-time warning if something is configured incorrectly, but it will never retry once the issue was fixed. Thus, either raise the warnings to error or consider retrying error states. Even as this method call is within the HA realization, I don't think anything here should block. So maybe the whole state caching would be dropped.

pkg/icingadb/ha.go
```diff
@@ -237,6 +281,22 @@ func (h *HA) controller() {
 				h.signalHandover("lost heartbeat")
 				h.realizeLostHeartbeat()
 			}
+
+		case healthy, ok := <-h.notificationsHeartbeatCh:
+			if !ok {
+				h.logger.Info("Notifications heartbeat channel closed")
+				h.notifications.healthy = types.MakeBool(false, types.TransformZeroBoolToNull)
+				continue
```
oxzi: If `h.notificationsHeartbeatCh` would be closed, this case fires. The `continue` jumps back to the `for` loop and the `select`, where the same case would directly fire again, and so on. This would result in both a storm of info log events as well as a 100% CPU utilization.

After briefly checking `notifications.Client`, I found no evidence of the channel being closed - might have missed it, though. So, if a closed channel is an unexpected situation, we could error out here instead.

pkg/icingadb/ha.go
oxzi: Okay, back then I missed the `now.After(h.notifications.unhealthySince.Add(retry.DefaultTimeout))` safeguard. After jumping multiple times through the HA controller, it seems all valid.

pkg/icingadb/ha.go
yhabteab: No, currently it's not closed but since we plan to enable and disable the [Notifications component at runtime](https://github.com/Icinga/icingadb/issues/1108), I thought, we might need this channel to close. However, after thinking a bit about it, we don't even need to close it at all, so I will remove that check.

pkg/icingadb/ha.go
yhabteab: > Even as this method call is within the HA realization, I don't think anything here should block. So maybe the whole state caching would be dropped.

I've now changed this to cache only the sucess case instead of making always two superfluous sys calls with each heartbeat realization.

yhabteab:
> > Since #1113 isn't done yet, I'm not sure what type for `notifications_socket_path_discovered` is suitable.
> 
> That issue is now closed, and the decision on which transport to use is deduced from the url scheme. So, what's your final say on the type for `notifications_socket_path_discovered`?

Ping @nilmerg.

doc/03-Configuration.md
```diff
@@ -181,11 +181,12 @@ ICINGADB_RETENTION_OPTIONS=comment:356
...
 ## Notifications Configuration
 
-!!! warning
+!!! tip
 
-    The Icinga Notifications integration is a feature preview that you can try out.
-    However, incompatible changes may happen, so be sure to check the changelog for future updates.
-    Please do not use it in production.
+    If the `ICINGA_NOTIFICATIONS_LISTENER_SOCKET` environment variable is set, Icinga DB will automatically detect the
+    Icinga Notifications API Unix Domain Socket availability and persit the result in the database. This allows
```
oxzi:
```suggestion
    Icinga Notifications API Unix Domain Socket availability and persist the result in the database. This allows
```

oxzi (commented):
Otherwise, LGTM.

yhabteab:
> > > Since #1113 isn't done yet, I'm not sure what type for `notifications_socket_path_discovered` is suitable.
> > 
> > 
> > That issue is now closed, and the decision on which transport to use is deduced from the url scheme. So, what's your final say on the type for `notifications_socket_path_discovered`?
> 
> Ping @nilmerg.

As just discussed offline, I've renamed it to `notifications_discovered_socket_path` as requested.

oxzi (approved):
LGTM. @nilmerg, it's up to you.

yhabteab:
Just moved some changes from the last to the first commit but didn't change anything code related, thus your approval wasn't dismissed.

nilmerg (approved):

PR #1154, opened 2026-07-20
notifications: replace `GetIncidents` with `YieldIncidents`

yhabteab:
Makes use of the newly introduced `YieldIncidents` function in IGL.


pkg/notifications/notifications.go
oxzi: Returning here could result in a not drained `incidentsCh` and a leaking Goroutine. Consider deriving a context with a deferred cancel function.

yhabteab:
## Blocked by
- https://github.com/Icinga/icingadb/pull/1151

pkg/notifications/notifications.go
yhabteab: It should be good now.

oxzi (approved):

PR #1156, opened 2026-07-21
Sync host & service check outputs to Notifications periodically

### LINKED ISSUE #1140: Continuously synchronize open Icinga Notification Incidents with updated check outputs
Icinga 2 is not transmitting a runtime update in case only the plugin output of a checkable changes. Such an update is only available in redis as long as no hard state change occurs. Icinga DB Web accesses redis directly to visualize the most recent outputs in the UI.

Icinga Notifications does not have direct access like this and will never have it, so it's Icinga DB's responsiblity that transmitted notifications contain the most recent output as well. In order for this to work, it needs to poll redis every 5 minutes for updated outputs.

This should be coupled with the `/incidents` API endpoint provided by Icinga Notifications:

1. Fetch open incidents via API
2. Fetch matching outputs via redis (1-2 queries, for hosts and services)
3. Push updated incident messages via API

With #1139 the first fetch can receive only incidents for which it is responsible and the accompanying object tags should provide enough data to construct the redis key for each individual object to keep interactions with redis to a minimum.

yhabteab:
In order to always have the latest check output of a checkable, there is now a new goroutine that periodically (every 5 minutes) syncs the check output of all checkables that have an active incident state from the `icinga:{host,service}:state` Redis hashes to Icinga Notifications. This is necessary because a checkable's check can, for instance, be in `CRITICAL` state for a long time, while its output can change with each check result. And since that won't trigger any state change events, thus, not queued to the runtime updates stream, there's no way for Icinga Notifications to know about the updated output.

```bash
icingadb-1  | 2026-07-21T13:08:45.586+0200	DEBUG	notifications	Syncing check outputs for incidents	{"host_incidents": 0, "service_incidents": 10, "last_sync": "2026-07-21T13:03:45.586+0200"}
icingadb-1  | 2026-07-21T13:13:45.650+0200	DEBUG	notifications	Syncing check outputs for incidents {"host_incidents": 0, "service_incidents": 10, "last_sync": "2026-07-21T13:08:45.645+0200"}
icingadb-1  | 2026-07-21T13:18:45.719+0200	DEBUG	notifications	Syncing check outputs for incidents {"host_incidents": 0, "service_incidents": 10, "last_sync": "2026-07-21T13:13:45.718+0200"}
icingadb-1  | 2026-07-21T13:23:45.792+0200	DEBUG	notifications	Syncing check outputs for incidents	{"host_incidents": 0, "service_incidents": 10, "last_sync": "2026-07-21T13:18:45.792+0200"}
icingadb-1  | 2026-07-21T13:28:45.869+0200	DEBUG	notifications	Syncing check outputs for incidents	{"host_incidents": 0, "service_incidents": 10, "last_sync": "2026-07-21T13:23:45.869+0200"}
```

resolves #1140

### DIFF
--- a/pkg/notifications/notifications.go
+++ b/pkg/notifications/notifications.go
@@ -8,6 +8,7 @@ import (
 	"maps"
 	"reflect"
 	"slices"
+	"strings"
 	"sync"
 	"time"
 
@@ -143,7 +144,17 @@ func (client *Client) ApplyDelta(ctx context.Context, delta *icingadb.Delta) err
 		return nil
 	}
 
-	client.fetchIncidents(ctx)
+	func() {
+		client.incidentsMu.Lock()
+		defer client.incidentsMu.Unlock()
+
+		// ApplyDelta is called in parallel for each entity type, so we need to ensure
+		// that we only fetch the incidents once per environment.
+		if client.incidentsByObjId != nil {
+			return
+		}
+		client.incidentsByObjId = client.retrieveEnvironmentIncidents(ctx)
+	}()
 	if len(client.incidentsByObjId) == 0 {
 		return nil
 	}
@@ -177,8 +188,8 @@ func (client *Client) ApplyDelta(ctx context.Context, delta *icingadb.Delta) err
 					if same, err := HaveSameState(incident, entity); err != nil {
 						return err
 					} else if same {
-						// Should we send a message update here too or just wait for the timer to expire
-						// and sync the message accordingly? See https://github.com/Icinga/icingadb/issues/1140
+						// Same state, but not necessarily the same output/message, and since we have a separate
+						// worker for syncing check outputs (see Client.SyncCheckOutputs), skip it entirely here.
 						continue
 					}
 				}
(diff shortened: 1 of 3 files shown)


pkg/notifications/notifications.go
```diff
@@ -234,27 +245,116 @@ func (client *Client) ApplyDelta(ctx context.Context, delta *icingadb.Delta) err
...
+						return modify(&ss.State, servicesIncidents[ss.ServiceId.String()].ObjectTags)
+					}, slices.Collect(maps.Keys(servicesIncidents))...)
+					if err != nil {
+						client.logger.Errorw("Failed to sync service state check outputs", zap.Error(err))
+					}
+				})
+			}
+			wg.Wait()
+
+			lastSync = time.Now()
```
oxzi: One single state which cannot be processed properly by Icinga Notifications is enough to take down the stream, resulting all following states to be skipped. However, the `lastSync` advances regardless of any error, resulting in those potentially skipped states to never be processed.

https://github.com/Icinga/icingadb/blob/f02d816fbeabc45a049b12ca7e4169dda1eba25a/pkg/notifications/notifications.go#L280

https://github.com/Icinga/icingadb/blob/f02d816fbeabc45a049b12ca7e4169dda1eba25a/pkg/notifications/notifications.go#L315-L316

Would it be possible to set `lastSync` based on the last processed state? For example by using `s.LastUpdate`?

(And, just to say it once, we are comparing remote timestamps against ours. In an ideal world, these clocks would be identical..)

pkg/notifications/notifications.go
```diff
@@ -234,27 +245,116 @@ func (client *Client) ApplyDelta(ctx context.Context, delta *icingadb.Delta) err
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+
+		case <-ticker.C:
+			hostIncidents := make(map[string]source.Incident)
+			servicesIncidents := make(map[string]source.Incident)
```
oxzi:
```suggestion
			serviceIncidents := make(map[string]source.Incident)
```

pkg/notifications/notifications.go
oxzi: That's the very issue this PR addresses, right?

yhabteab: Yep. Will remove that.

pkg/notifications/notifications.go
yhabteab: > Would it be possible to set `lastSync` based on the last processed state? For example by using `s.LastUpdate`?

That wouldn't fix the problem either since I'm processing the entries in arbitrary order and not sorted by last_update (Redis hash ≠ Redis stream). The only other option I would consider is retrying the failing request endlessly until it succeeds, but that seems to be an overkill for such an insignficant task.

> (And, just to say it once, we are comparing remote timestamps against ours. In an ideal world, these clocks would be identical..)

If not, we would have other serious problems than skipped message updates.
https://github.com/Icinga/icingadb/blob/bfb430e6d9b02b87fe5eaa127ab22dfb1a7d82f6/pkg/icingadb/ha.go#L202-L210

pkg/notifications/notifications.go
yhabteab: Fixed.

pkg/notifications/notifications.go
yhabteab: Rephrased the comment.

pkg/notifications/notifications.go
yhabteab: I've wrapped the `client.ModifyIncidents` call in a `retry.WithBackoff` now.

pkg/notifications/notifications.go
```diff
@@ -234,27 +245,132 @@ func (client *Client) ApplyDelta(ctx context.Context, delta *icingadb.Delta) err
...
+							zap.String("error", err.Error()))
+					}
+				},
+				OnRetryableError: func(elapsed time.Duration, attempt uint64, err, lastErr error) {
+					client.sendHeartbeat(false)
+					if lastErr == nil || err.Error() != lastErr.Error() {
+						client.logger.Errorw("Failed to update incident status",
+							zap.String("object_type", types.Name(s)),
+							zap.Duration("elapsed", elapsed),
+							zap.String("error", lastErr.Error()))
```
oxzi: The if allows `lastErr == nil`, but for the "error" field, `lastErr.Error()` is called, dereferencing a potential nil pointer.

pkg/notifications/notifications.go
```diff
@@ -234,27 +245,132 @@ func (client *Client) ApplyDelta(ctx context.Context, delta *icingadb.Delta) err
...
+			if len(serviceIncidents) > 0 {
+				wg.Go(func() {
+					_ = streamRedisHashObjects(ctx, client.redisClient, "icinga:service:state", func(ss *v1.ServiceState, _ string) error {
+						return modify(&ss.State, serviceIncidents[ss.ServiceId.String()].ObjectTags)
+					}, slices.Collect(maps.Keys(serviceIncidents))...)
+				})
+			}
+			wg.Wait()
+
+			lastSync = time.Now()
```
oxzi: Sorry for being annoying, but I am still unsure about how`lastSync` works. Doesn't updating the timestamp after the finished `WaitGroup` allow missing state changes mid flight?

Consider the following:

|`t`||
|-|-|
|0s| Start, `ticker.C` fires. Every event with `s.LastUpdate = 0s >= 0s` is accepted. |
|23s| New event appears with `s.LastUpdate = 23s` |
|42s| `wg.Wait()` finishes and `lastSync = 42s` |
|342s| Next round, `ticker.C` fires. The prior event is ignored since `s.LastUpdate = 23s <= 42s`, falling into the initial return case of `modify`. |

pkg/notifications/notifications.go
yhabteab: Fixed.

pkg/notifications/notifications.go
yhabteab: You're right. As discussed offline, the `lastSync` corresponds to the `ticker.C` delivered time now.

oxzi (approved):

PR #1158, opened 2026-07-28
Notifications: Load Configuration from Database

### LINKED ISSUE #1108: Load Notifications configuration from the database

oxzi:
A new configuration path for the Icinga Notifications client based on the database was added. While manually configured fields from Icinga DB take precedence, allowed fields from the database can be used. By request, this feature is opt-out and can be disabled.

---

The docs state that "a MySQL (≥8.0), MariaDB (≥10.2.2), or PostgreSQL (≥9.6) database is required to run Icinga DB." So, let's get rid of the outdated ones.

---

I have removed the unsupported versions as they are, well, unsupported and that the new index has a size of 20 + 255*4 + 20 = 1060, which exceeds MariaDB's old limit.

Fixes #1108.

### DIFF
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -137,7 +137,12 @@ func run() int {
 		}
 		defer func() { _ = db.Close() }()
 		db.SetMaxOpenConns(1)
-		ha = icingadb.NewHA(ctx, db, heartbeat, logs.GetChildLogger("high-availability"))
+		ha = icingadb.NewHA(
+			ctx,
+			db,
+			heartbeat,
+			logs.GetChildLogger("high-availability"),
+			cmd.Config.Notifications.SynchronizeWithDatabase)
 
 		telemetryLogger := logs.GetChildLogger("telemetry")
 		telemetrySyncStats = telemetry.StartHeartbeat(ctx, rc, telemetryLogger, ha, heartbeat)
@@ -170,18 +175,23 @@ func run() int {
 	signal.Notify(sig, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
 
 	var notificationsSource *notifications.Client
-	if cfg := cmd.Config.Notifications; cfg.Url != "" {
-		logger.Info("Starting Icinga Notifications source")
-
+	if cfg := cmd.Config.Notifications; cfg.SynchronizeWithDatabase || cfg.Url != "" {
 		notificationsSource, err = notifications.NewNotificationsClient(
 			db,
 			rc,
 			logs.GetChildLogger("notifications"),
 			cfg,
-			ha.NotificationsHeartbeat())
+			ha)
 		if err != nil {
 			logger.Fatalw("Can't create Icinga Notifications client from config", zap.Error(err))
 		}
+
+		go func() {
+			err := notificationsSource.PersistLockedConfigOnce(ctx)
+			if err != nil && !utils.IsContextCanceled(err) {
+				logger.Fatalf("%+v", err)
+			}
+		}()
 	}
 
 	go func() {
@@ -207,6 +217,15 @@ func run() int {
 			case takeoverReason := <-ha.Takeover():
 				logger.WithOptions(logs.ForceLog()).Infow("Taking over", zap.String("reason", takeoverReason))
 
+				if notificationsSource != nil {
+					go func() {
+						err := notificationsSource.SynchronizeConfigWithDatabase(hactx)
+						if err != nil && !utils.IsContextCanceled(err) {
+							logger.Fatalf("%+v", err)
+						}
+					}()
+				}
+
 				go func() {
 					for hactx.Err() == nil {
 						synctx, cancelSynctx := context.WithCancel(ha.Environment().NewContext(hactx))
@@ -215,6 +234,10 @@ func run() int {
 						// Runtime updates must wait for initial synchronization to complete.
 						configInitSync := sync.WaitGroup{}
 						stateInitSync := &sync.WaitGroup{}
+
+						// Decided once per sync round, so that everything within agrees on whether Icinga Notifications
+						// is available. Any later change of the Client cancels synctx and starts a new round.
+						notificationsConfigured := notificationsSource != nil && notificationsSource.IsConfigured()
 
 						// Clear the runtime update streams before starting anything else (rather than after the sync),
 						// otherwise updates may be lost.
@@ -231,11 +254,19 @@ func run() int {
 						})
 
 						g.Go(func() error {
+							var notificationsReconfiguredCh <-chan struct{}
+							if notificationsSource != nil {
+								notificationsReconfiguredCh = notificationsSource.ReconfiguredCh()
+							}
+
 							select {
 							case <-dump.InProgress():
 								logger.Info("Icinga 2 started a new config dump, waiting for it to complete")
 								cancelSynctx()
-
+								return nil
+							case <-notificationsReconfiguredCh:
+								logger.Info("Icinga Notifications component has been reconfigured, restarting sync")
+								cancelSynctx()
 								return nil
 							case <-synctx.Done():
 								return synctx.Err()
@@ -270,7 +301,7 @@ func run() int {
 								defer stateInitSync.Done()
 
 								var hook icingadb.DeltaHook
-								if notificationsSource != nil {
+								if notificationsConfigured {
 									hook = notificationsSource.ApplyDelta
 									defer func() {
 										if stateSyncWorkers.Add(-1) == 0 {
@@ -353,7 +384,7 @@ func run() int {
 							logger.Info("Starting state runtime updates sync")
 
 							runtimeUpdatesOpts := []icingadb.RUOption{icingadb.WithAllowParallel()}
-							if notificationsSource != nil {
+							if notificationsConfigured {
 								runtimeUpdatesOpts = append(runtimeUpdatesOpts, icingadb.WithRUUpsert(notificationsSource.Submit))
 							}
 
@@ -374,7 +405,7 @@ func run() int {
 							return ret.Start(synctx)
 						})
 
-						if notificationsSource != nil {
+						if notificationsConfigured {
 							g.Go(func() error {
 								stateInitSync.Wait()
 
--- a/config.example.yml
+++ b/config.example.yml
@@ -152,7 +152,7 @@ redis:
 	#   For example: https://example.com:5680
 	# - unix: HTTP connection over a Unix Domain Socket. Authentication is based on the operating system user.
 	#   For example: unix:///path/to/socket
-#  url: http://localhost:5680
+#  url:
 
   # Base URL of the Icinga Web 2 setup in front of this Icinga DB. It is used to build the absolute object URLs
   # transmitted with each event, which Icinga Notifications passes on to the notified contacts. This must be an
@@ -160,10 +160,10 @@ redis:
 #  icingaweb2_url: https://example.com/icingaweb2
 
   # Username to authenticate against the Icinga Notifications API.
-#  username: icingadb
+#  username:
 
   # Password for the defined user.
-#  password: insecureinsecure
+#  password:
 #  password_file: /run/secerets/icingadb_notifications_password
 
   # TLS configuration for https URLs. The ca is used to verify the server cert, while cert and key are the client certificate.
@@ -171,7 +171,10 @@ redis:
 #  cert: /path/to/client.crt
 #  key: /path/to/client.key
 
-  # Relations as a JSONPath to resolve and include in the events submitted to Icinga Notifications.
+  # Relations as an array of JSONPath to resolve and include in the events submitted to Icinga Notifications.
+  # For example '$.host.vars' and '$.services[*].vars'.
 #  default_relations:
-#    - '$.host.vars'
-#    - '$.services[*].vars'
+
+  # Synchronize the configuration with the database. This allows Icinga Notifications Web to configure this source.
+  # If set, all other fields can be left empty. However, if set, they take precedence.
+#  synchronize_with_database: true
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"github.com/creasty/defaults"
 	"github.com/icinga/icinga-go-library/database"
 	"github.com/icinga/icinga-go-library/logging"
@@ -9,6 +10,9 @@ import (
 	"github.com/icinga/icingadb/pkg/icingadb/history"
 	"github.com/pkg/errors"
 	"github.com/theory/jsonpath"
+	"reflect"
+	"slices"
+	"strings"
 	"time"
 )
 
@@ -17,11 +21,11 @@ const DefaultConfigPath = "/etc/icingadb/config.yml"
 
 // Config defines Icinga DB config.
 type Config struct {
-	Database      database.Config `yaml:"database" envPrefix:"DATABASE_"`
-	Redis         redis.Config    `yaml:"redis" envPrefix:"REDIS_"`
-	Logging       logging.Config  `yaml:"logging" envPrefix:"LOGGING_"`
-	Retention     RetentionConfig `yaml:"retention" envPrefix:"RETENTION_"`
-	Notifications source.Config   `yaml:"notifications" envPrefix:"NOTIFICATIONS_"`
+	Database      database.Config     `yaml:"database" envPrefix:"DATABASE_"`
+	Redis         redis.Config        `yaml:"redis" envPrefix:"REDIS_"`
+	Logging       logging.Config      `yaml:"logging" envPrefix:"LOGGING_"`
+	Retention     RetentionConfig     `yaml:"retention" envPrefix:"RETENTION_"`
+	Notifications NotificationsConfig `yaml:"notifications" envPrefix:"NOTIFICATIONS_"`
 }
 
 func (c *Config) SetDefaults() {
@@ -123,4 +127,65 @@ func (r *RetentionConfig) Validate() error {
 	}
 
 	return r.Options.Validate()
+}
+
+// NotificationsConfig wraps source.Config for autoconfiguration and env-like serialization.
+//
+// When NotificationsConfig.SynchronizeWithDatabase is set, everything is sourced from the database.
+type NotificationsConfig struct {
+	source.Config `yaml:",inline"`
+
+	SynchronizeWithDatabase bool `yaml:"synchronize_with_database" env:"SYNCHRONIZE_WITH_DATABASE" default:"true"`
+}
+
+// StaticConfig emits all set/changed/non-default configuration values with a key mimicking the env key.
+//
+// Pairs carrying sensitive information, such as passwords or private keys, are stored with a redacted value.
+func (c NotificationsConfig) StaticConfig() map[string]string {
+	out := make(map[string]string)
+	activeEncConfig(reflect.ValueOf(c.Config), "ICINGADB_NOTIFICATIONS_", out)
+	return out
+}
+
+// activeEncConfig emits all set config values into a map with a key mimicking the env key.
+//
+// This function is used in NotificationsConfig.StaticConfig and currently only implements what is required there.
+//
+// Fields with the "unset" env tag option are considered sensitive. Their value will be REDACTED.
+func activeEncConfig(v reflect.Value, prefix string, out map[string]string) {
+	t := v.Type()
+	for i := range t.NumField() {
+		fieldT := t.Field(i)
+		fieldV := v.Field(i)
+
+		if !fieldT.IsExported() {
+			continue
+		}
+
+		if fieldV.Kind() == reflect.Struct {
+			activeEncConfig(fieldV, prefix+fieldT.Tag.Get("envPrefix"), out)
+			continue
+		}
+
+		name, opts, _ := strings.Cut(fieldT.Tag.Get("env"), ",")
+		if name == "" || name == "-" {
+			continue
+		}
+
+		if defaults.CanUpdate(fieldV.Interface()) {
+			continue
+		}
+
+		if slices.Contains(strings.Split(opts, ","), "unset") {
+			out[prefix+name] = "REDACTED"
+		} else if fieldV.Kind() == reflect.Slice {
+			parts := make([]string, fieldV.Len())
+			for i := range parts {
+				parts[i] = fmt.Sprintf("%v", fieldV.Index(i).Interface())
+			}
+			out[prefix+name] = strings.Join(parts, ",")
+		} else {
+			out[prefix+name] = fmt.Sprintf("%v", fieldV.Interface())
+		}
+	}
 }
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -1,10 +1,12 @@
 package config
 
 import (
+	"github.com/caarlos0/env/v11"
 	"github.com/creasty/defaults"
 	"github.com/icinga/icinga-go-library/config"
 	"github.com/icinga/icinga-go-library/database"
 	"github.com/icinga/icinga-go-library/logging"
+	"github.com/icinga/icinga-go-library/notifications/source"
 	"github.com/icinga/icinga-go-library/redis"
 	"github.com/icinga/icinga-go-library/testutils"
 	"github.com/icinga/icingadb/pkg/icingadb/history"
@@ -183,6 +185,36 @@ retention:
 			},
 		},
 		{
+			Name: "Custom Notifications Config from YAML",
+			Data: testutils.ConfigTestData{
+				Yaml: yamlConfig + `
+notifications:
+  url: http://example.com/
+  username: icingadb
+  password: insecure
+`,
+			},
+			Expected: &Config{
+				Database: database.Config{
+					Host:     "192.0.2.1",
+					Database: "icingadb",
+					User:     "icingadb",
+					Password: "icingadb",
+				},
+				Redis: redis.Config{
+					Host: "2001:db8::1",
+				},
+				Notifications: NotificationsConfig{
+					Config: source.Config{
+						Url:      "http://example.com/",
+						Username: "icingadb",
+						Password: "insecure",
+					},
+					SynchronizeWithDatabase: true,
+				},
+			},
+		},
+		{
 			Name: "Unknown YAML field",
 			Data: testutils.ConfigTestData{
 				Yaml: `unknown: unknown`,
@@ -218,4 +250,30 @@ retention:
 			return actual, err
 		}))
 	}
+}
+
+func TestNotificationsConfig_StaticConfig(t *testing.T) {
+	cfg := NotificationsConfig{
+		Config: source.Config{
+			Url:              "https://example.com:5680",
+			Username:         "icingadb",
+			Password:         "icingadb",
+			TlsOptions:       config.TLS{Insecure: true},
+			DefaultRelations: []string{"$.foo", "$.bar"},
+		},
+	}
+
+	staticConf := cfg.StaticConfig()
+	require.Equal(t, map[string]string{
+		"ICINGADB_NOTIFICATIONS_URL":               "https://example.com:5680",
+		"ICINGADB_NOTIFICATIONS_USERNAME":          "icingadb",
+		"ICINGADB_NOTIFICATIONS_PASSWORD":          "REDACTED",
+		"ICINGADB_NOTIFICATIONS_INSECURE":          "true",
+		"ICINGADB_NOTIFICATIONS_DEFAULT_RELATIONS": "$.foo,$.bar",
+	}, staticConf)
+
+	var cmpCfg Config
+	require.NoError(t, env.ParseWithOptions(&cmpCfg, env.Options{Prefix: "ICINGADB_", Environment: staticConf}))
+	cmpCfg.Notifications.Config.Password = "icingadb" // Reset, as redacted in output
+	require.Equal(t, cfg, cmpCfg.Notifications)
 }
--- a/pkg/icingadb/ha.go
+++ b/pkg/icingadb/ha.go
@@ -59,33 +59,41 @@ func (ns *NotificationsState) IsValidUnhealthy() bool {
 
 // HA provides high availability and indicates whether a Takeover or Handover must be made.
 type HA struct {
-	state         atomic.Pointer[haState]
-	ctx           context.Context
-	cancelCtx     context.CancelFunc
-	instanceId    types.Binary
-	db            *database.DB
-	environmentMu sync.Mutex
-	environment   *v1.Environment
-	heartbeat     *icingaredis.Heartbeat
-	logger        *logging.Logger
-	responsible   bool
-	handover      chan string
-	takeover      chan string
-	done          chan struct{}
-	errOnce       sync.Once
-	errMu         sync.Mutex
-	err           error
+	state       atomic.Pointer[haState]
+	ctx         context.Context
+	cancelCtx   context.CancelFunc
+	instanceId  types.Binary
+	db          *database.DB
+	environment *v1.Environment
+	endpointId  types.Binary
+	heartbeat   *icingaredis.Heartbeat
+	logger      *logging.Logger
+	responsible bool
+	handover    chan string
+	takeover    chan string
+	done        chan struct{}
+	errOnce     sync.Once
+	err         error
+	mu          sync.Mutex // protects environment, endpointId, and err
 
 	// notificationsHeartbeat is a channel that signals the status of the Icinga Notifications component.
 	//
 	// This will only receive events when the Icinga Notifications component is enabled, and its true value indicates
 	// that the component is healthy and able to process notifications, while a false value indicates that it is not.
-	notificationsHeartbeatCh chan bool
-	notifications            NotificationsState
+	// An invalid types.Bool resets the state to unknown, e.g., after the component was stopped again.
+	notificationsHeartbeatCh             chan types.Bool
+	notifications                        NotificationsState
+	notificationsSynchronizeWithDatabase types.Bool
 }
 
 // NewHA returns a new HA and starts the controller loop.
-func NewHA(ctx context.Context, db *database.DB, heartbeat *icingaredis.Heartbeat, logger *logging.Logger) *HA {
+func NewHA(
+	ctx context.Context,
+	db *database.DB,
+	heartbeat *icingaredis.Heartbeat,
+	logger *logging.Logger,
+	notificationsSynchronizeWithDatabase bool,
+) *HA {
 	ctx, cancelCtx := context.WithCancel(ctx)
 
 	instanceId := uuid.New()
@@ -101,7 +109,8 @@ func NewHA(ctx context.Context, db *database.DB, heartbeat *icingaredis.Heartbea
 		takeover:   make(chan string),
 		done:       make(chan struct{}),
 
-		notificationsHeartbeatCh: make(chan bool),
+		notificationsHeartbeatCh:             make(chan types.Bool),
+		notificationsSynchronizeWithDatabase: types.MakeBool(notificationsSynchronizeWithDatabase),
 	}
 
 	ha.state.Store(&haState{})
@@ -130,16 +139,24 @@ func (h *HA) Done() <-chan struct{} {
 
 // Environment returns the current environment.
 func (h *HA) Environment() *v1.Environment {
-	h.environmentMu.Lock()
-	defer h.environmentMu.Unlock()
+	h.mu.Lock()
+	defer h.mu.Unlock()
 
 	return h.environment
 }
 
+// EndpointID returns the current endpoint ID.
+func (h *HA) EndpointID() types.Binary {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	return h.endpointId
+}
+
 // Err returns an error if Done has been closed and there is an error. Otherwise returns nil.
 func (h *HA) Err() error {
-	h.errMu.Lock()
-	defer h.errMu.Unlock()
+	h.mu.Lock()
+	defer h.mu.Unlock()
 
 	return h.err
 }
@@ -158,7 +175,8 @@ func (h *HA) Takeover() chan string {
 //
 // The channel is used to signal the health of the Icinga Notifications component, where a true value indicates
 // that the component is healthy and able to process notifications, while a false value indicates that it is not.
-func (h *HA) NotificationsHeartbeat() chan<- bool { return h.notificationsHeartbeatCh }
+// An invalid types.Bool marks the state as unknown, as it is before the component has been configured.
+func (h *HA) NotificationsHeartbeat() chan<- types.Bool { return h.notificationsHeartbeatCh }
 
 // State returns the status quo.
 func (h *HA) State() (responsibleTsMilli int64, responsible, otherResponsible bool) {
@@ -169,9 +187,9 @@ func (h *HA) State() (responsibleTsMilli int64, responsible, otherResponsible bo
 
 func (h *HA) abort(err error) {
 	h.errOnce.Do(func() {
-		h.errMu.Lock()
+		h.mu.Lock()
 		h.err = errors.Wrap(err, "HA aborted")
-		h.errMu.Unlock()
+		h.mu.Unlock()
 
 		h.cancelCtx()
 	})
@@ -222,6 +240,17 @@ func (h *HA) controller() {
 					h.abort(err)
 				}
 
+				h.mu.Lock()
+				if !bytes.Equal(h.endpointId, s.EndpointId) {
+					if h.endpointId.Valid() {
+						h.logger.Warnw("Endpoint ID changed unexpectedly",
+							zap.String("current", h.endpointId.String()),
+							zap.String("new", s.EndpointId.String()))
+					}
+					h.endpointId = s.EndpointId
+				}
+				h.mu.Unlock()
+
 				envId, err := m.EnvironmentID()
 				if err != nil {
 					h.abort(err)
@@ -234,14 +263,14 @@ func (h *HA) controller() {
 							zap.String("new", envId.String()))
 					}
 
-					h.environmentMu.Lock()
+					h.mu.Lock()
 					h.environment = &v1.Environment{
 						EntityWithoutChecksum: v1.EntityWithoutChecksum{IdMeta: v1.IdMeta{
 							Id: envId,
 						}},
 						Name: types.MakeString(envId.String()),
 					}
-					h.environmentMu.Unlock()
+					h.mu.Unlock()
 				}
 
 				select {
@@ -283,9 +312,9 @@ func (h *HA) controller() {
 			}
 
 		case healthy := <-h.notificationsHeartbeatCh:
-			h.logger.Debugw("Received notifications heartbeat", zap.Bool("healthy", healthy))
-			h.notifications.healthy = types.MakeBool(healthy)
-			if healthy {
+			h.logger.Debugw("Received notifications heartbeat", zap.Any("healthy", healthy))
+			h.notifications.healthy = healthy
+			if !healthy.Valid || healthy.Bool {
 				h.notifications.unhealthySince = time.Time{}
 			} else if h.notifications.unhealthySince.IsZero() {
 				h.notifications.unhealthySince = time.Now()
@@ -407,21 +436,22 @@ func (h *HA) realize(
 				EnvironmentMeta: v1.EnvironmentMeta{
 					EnvironmentId: envId,
 				},
-				Heartbeat:                         types.UnixMilli(time.UnixMilli(h.heartbeat.LastMessageTime())),
-				Responsible:                       types.Bool{Bool: takeover != "" || h.responsible, Valid: true},
-				EndpointId:                        s.EndpointId,
-				Icinga2Version:                    s.Version,
-				Icinga2StartTime:                  s.ProgramStart,
-				Icinga2NotificationsEnabled:       s.NotificationsEnabled,
-				Icinga2ActiveServiceChecksEnabled: s.ActiveServiceChecksEnabled,
-				Icinga2ActiveHostChecksEnabled:    s.ActiveHostChecksEnabled,
-				Icinga2EventHandlersEnabled:       s.EventHandlersEnabled,
-				Icinga2FlapDetectionEnabled:       s.FlapDetectionEnabled,
-				Icinga2PerformanceDataEnabled:     s.PerformanceDataEnabled,
-				IcingadbVersion:                   internal.Version.Version,
-				IcingadbServiceUser:               ServiceUser(),
-				NotificationsHealthy:              h.notifications.healthy,
-				NotificationsDiscoveredSocketPath: h.probeNotificationsSocketPath(),
+				Heartbeat:                            types.UnixMilli(time.UnixMilli(h.heartbeat.LastMessageTime())),
+				Responsible:                          types.Bool{Bool: takeover != "" || h.responsible, Valid: true},
+				EndpointId:                           s.EndpointId,
+				Icinga2Version:                       s.Version,
+				Icinga2StartTime:                     s.ProgramStart,
+				Icinga2NotificationsEnabled:          s.NotificationsEnabled,
+				Icinga2ActiveServiceChecksEnabled:    s.ActiveServiceChecksEnabled,
+				Icinga2ActiveHostChecksEnabled:       s.ActiveHostChecksEnabled,
+				Icinga2EventHandlersEnabled:          s.EventHandlersEnabled,
+				Icinga2FlapDetectionEnabled:          s.FlapDetectionEnabled,
+				Icinga2PerformanceDataEnabled:        s.PerformanceDataEnabled,
+				IcingadbVersion:                      internal.Version.Version,
+				IcingadbServiceUser:                  ServiceUser(),
+				NotificationsHealthy:                 h.notifications.healthy,
+				NotificationsDiscoveredSocketPath:    h.probeNotificationsSocketPath(),
+				NotificationsSynchronizeWithDatabase: h.notificationsSynchronizeWithDatabase,
 			}
 
 			stmt, _ := h.db.BuildUpsertStmt(i)
--- a/pkg/notifications/notifications.go
+++ b/pkg/notifications/notifications.go
@@ -10,8 +10,10 @@ import (
 	"slices"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
+	"github.com/caarlos0/env/v11"
 	"github.com/icinga/icinga-go-library/backoff"
 	"github.com/icinga/icinga-go-library/com"
 	"github.com/icinga/icinga-go-library/database"
@@ -26,6 +28,7 @@ import (
 	"github.com/icinga/icinga-go-library/types"
 	"github.com/icinga/icinga-go-library/utils"
 	"github.com/icinga/icingadb/internal"
+	"github.com/icinga/icingadb/internal/config"
 	"github.com/icinga/icingadb/pkg/common"
 	"github.com/icinga/icingadb/pkg/contracts"
 	"github.com/icinga/icingadb/pkg/icingadb"
@@ -34,9 +37,18 @@ import (
 	v1history "github.com/icinga/icingadb/pkg/icingadb/v1/history"
 	"github.com/icinga/icingadb/pkg/icingaredis"
 	"github.com/icinga/icingadb/pkg/icingaredis/telemetry"
+	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
+)
+
+var (
+	// errClientReset is used as a cancellation cause when replacing or removing the Notifications client.
+	errClientReset = errors.New("notifications client was reset")
+
+	// errNonVolatileNonHardState is returned when a non-hard state change is attempted to be submitted for a non-volatile checkable.
+	errNonVolatileNonHardState = errors.New("non-hard state change for non-volatile checkable")
 )
 
 // fetchableEvent wraps both event.Event and relations, allowing to enrich the Event based on Notifications feedback.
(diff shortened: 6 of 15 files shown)


oxzi:
Updated the PR. In a nutshell:
- Split the synchronization transaction into one initial static config dump and a continuous read.
- Only sync if HA active; activation after HA takeover and bound to the HA context.
- Redact sensitive information from static configuration when writing to the DB.
- Simplified the database schema, made endpoint ID `NOT NULL` - prior `UNIQUE` constraint did not work due to `NULL`.

I toyed with a `last_update` column in the database, but dropped it again. Similar to Notifications, this has the disadvantage that it is hard to detect absence after :spider_web: deleted rows again. As the `SELECT` query is cheap and the comparison as well, I kept it that way.

oxzi:
As discussed with @BastianLedererIcinga, this PR must also allow to unconfigure itself. So, I am going to alter the logic to allow stopping the Notifications Client if its config vanished, and not only restart it if a new and valid config is available.

---

Also consider switching from the enable option flag to the env variable.

internal/config/config.go
```diff
@@ -9,19 +10,86 @@ import (
...
 	"time"
 )
 
+// NotificationsConfig wraps source.Config for autoconfiguration and env-like serialization.
+//
+// When NotificationsConfig.SynchronizeWithDatabase is set, everything is sourced from the database.
+type NotificationsConfig struct {
+	source.Config `yaml:",inline"`
+
+	SynchronizeWithDatabase bool `yaml:"synchronize_with_database" env:"SYNCHRONIZE_WITH_DATABASE"`
```
yhabteab: As discussed numerous times about this offline with you, and also just discussed with @nilmerg and @BastianLedererIcinga, I'm commenting it here for the record. Honestly, I was initially going to comment on using the existing socket discovery mechanism for this instead but since the `default_relations` option has to be always configurable via the UI, that wouldn't work out well. So, this new option is indeed required but it **must** be set to `true` by default, and users can explicitly opt-out if they don't want to use this automation.

internal/config/config.go
```diff
@@ -9,19 +10,86 @@ import (
...
+}
+
+func (c NotificationsConfig) Equals(other NotificationsConfig) bool {
```
yhabteab:
```suggestion
func (c NotificationsConfig) Equal(other NotificationsConfig) bool {
```

internal/config/config.go
```diff
@@ -9,19 +10,86 @@ import (
...
+	activeEncConfig(reflect.ValueOf(c.Config), "ICINGADB_NOTIFICATIONS_", out)
+	return out
+}
+
+// activeEncConfig emits all set config values into a map with a key mimicking the env key.
+//
+// This function is used in NotificationsConfig.StaticConfig and currently only implements what is required there.
+//
+// Fields with the "unset" env tag option are considered sensitive. Their value will be REDACTED.
+func activeEncConfig(v reflect.Value, prefix string, out map[string]string) {
```
yhabteab: I would move the function (maybe even the newly introduced type) to the end of the file and not place it above all the existing config types.

internal/config/config.go
```diff
@@ -9,19 +10,86 @@ import (
...
+	t := v.Type()
+	for i := range t.NumField() {
+		fieldT := t.Field(i)
+		fieldV := v.Field(i)
```
yhabteab: You should skip unexported fields here.

internal/config/config.go
```diff
@@ -9,19 +10,86 @@ import (
...
+
+		if fieldV.Kind() == reflect.Pointer && !fieldV.IsNil() && fieldV.Elem().Kind() == reflect.Struct {
```
yhabteab:
```suggestion
		if fieldV.Kind() == reflect.Pointer && !fieldV.IsNil() {
```

It doesn't need an explict check on the struct type.

internal/config/config_test.go
```diff
@@ -182,6 +184,37 @@ retention:
...
+		{
+			Name: "Custom Notifications Config from YAML",
+			Data: testutils.ConfigTestData{
+				Yaml: yamlConfig + `
+notifications:
+  url: http://example.com/
+  username: icingadb
+  password: insecure
+  synchronize_with_database: true
+					`,
```
yhabteab:
```suggestion
`,
```

Otherwise, the test case crashes with following error locally (I'm not sure why it's working on GitHub).
```
$ go test -race -v './internal/config/...'
....
=== RUN   TestConfig/Custom_Notifications_Config_from_YAML
    testutils.go:41:
        	Error Trace:	/Users/yhabteab/Workspace/go/icinga-go-library/testutils/testutils.go:41
        	Error:      	Received unexpected error:
        	            	[16:1] found character '	' that cannot start any token
        	            	  13 |   username: icingadb
        	            	  14 |   password: insecure
        	            	  15 |   synchronize_with_database: true
        	            	> 16 |
        	            	       ^

        	            	can't parse YAML file /var/folders/4v/fmvd3nln7ybb1fmggygmyfwh0000gn/T/3227003167.yaml
        	            	github.com/icinga/icinga-go-library/config.FromYAMLFile
        	            		/Users/yhabteab/Workspace/go/icinga-go-library/config/config.go:99
        	            	github.com/icinga/icinga-go-library/config.Load
        	            		/Users/yhabteab/Workspace/go/icinga-go-library/config/config.go:267
        	            	github.com/icinga/icingadb/internal/config.TestConfig.func1.1
        	            		/Users/yhabteab/Workspace/go/icingadb/internal/config/config_test.go:239
        	            	github.com/icinga/icinga-go-library/testutils.WithYAMLFile
        	            		/Users/yhabteab/Workspace/go/icinga-go-library/testutils/testutils.go:97
        	            	github.com/icinga/icingadb/internal/config.TestConfig.func1
        	            		/Users/yhabteab/Workspace/go/icingadb/internal/config/config_test.go:238
        	            	github.com/icinga/icingadb/internal/config.TestConfig.TestCase[...].F.func3
        	            		/Users/yhabteab/Workspace/go/icinga-go-library/testutils/testutils.go:36
        	            	testing.tRunner
        	            		/opt/homebrew/Cellar/go/1.26.5/libexec/src/testing/testing.go:2036
        	            	runtime.goexit
        	            		/opt/homebrew/Cellar/go/1.26.5/libexec/src/runtime/asm_arm64.s:1447
        	            	github.com/icinga/icinga-go-library/config.Load
        	            		/Users/yhabteab/Workspace/go/icinga-go-library/config/config.go:278
        	            	github.com/icinga/icingadb/internal/config.TestConfig.func1.1
        	            		/Users/yhabteab/Workspace/go/icingadb/internal/config/config_test.go:239
        	            	github.com/icinga/icinga-go-library/testutils.WithYAMLFile
        	            		/Users/yhabteab/Workspace/go/icinga-go-library/testutils/testutils.go:97
        	            	github.com/icinga/icingadb/internal/config.TestConfig.func1
        	            		/Users/yhabteab/Workspace/go/icingadb/internal/config/config_test.go:238
        	            	github.com/icinga/icingadb/internal/config.TestConfig.TestCase[...].F.func3
        	            		/Users/yhabteab/Workspace/go/icinga-go-library/testutils/testutils.go:36
        	            	testing.tRunner
        	            		/opt/homebrew/Cellar/go/1.26.5/libexec/src/testing/testing.go:2036
        	            	runtime.goexit
        	            		/opt/homebrew/Cellar/go/1.26.5/libexec/src/runtime/asm_arm64.s:1447
        	Test:       	TestConfig/Custom_Notifications_Config_from_YAML
```

pkg/icingadb/ha.go
```diff
@@ -66,6 +66,8 @@ type HA struct {
 	db            *database.DB
 	environmentMu sync.Mutex
 	environment   *v1.Environment
+	endpointId    types.Binary
+	endpointIdMu  sync.Mutex
```
yhabteab: Does every single field of this type really need its own mutex? That's not necessarily your fault as there are already two mutexes that guard two different trivial fields, but I would merge them into a single one instead of adding yet another one here.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,290 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+	}
+}
+
+// SynchronizeLockedConfigToDatabase writes the static Notifications configuration to the database.
+//
+// This method must only be called once, as the static configuration is not expected to change. Multiple calls will
+// raise an error.
+//
+// If the Client config disabled SynchronizeConfigWithDatabase, the method immediately returns with a nil error.
+func (client *Client) SynchronizeLockedConfigToDatabase(ctx context.Context) error {
```
yhabteab: This function name should be something like `PersistLockedConfigOnce` or the like as it currently implies as if it runs continuously, which is not the case. That way, it wouldn't be confused with the other `SynchronizeConfigWithDatabase` function that runs continuously in the background.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,290 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+		if client.ha.Environment() == nil {
+			client.logger.Warn("Notifications Client finds no active HA Environment")
+			return
+		}
+
+		environmentId := client.ha.Environment().Meta().EnvironmentId
+		if !environmentId.Valid() {
+			client.logger.Warn("Notifications Client finds an unpopulated Environment ID")
+			return
+		}
```
yhabteab: At this point, this should never happen, and if it does I would just let it crash instead of spamming the logs with such boilerplate messages that the user cannot do anything about.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,290 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+
+		dbConfRows := []v1.IcingadbConfig{}
```
yhabteab:
```suggestion
		var dbConfRows []v1.IcingadbConfig
```

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,290 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+		err := client.db.SelectContext(ctx, &dbConfRows, selectStmt, environmentId, v1.UnconfiguredEndpointId())
+		if err != nil {
+			client.logger.Errorw("Notifications Client cannot fetch configuration from database", zap.Error(err))
```
yhabteab:
```suggestion
			client.logger.Errorw("Cannot fetch configuration from database", zap.Error(err))
```

That's the pattern we use for logging. Also, all the extra `[Icinga] Notifications Client` in the log messages are superfluous as the logger is already named `notifications`. If you absolutely need such context message, then please remove the `Client` suffix everywhere. So, please make all the logs consistent.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,290 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+			return
+		}
+
+		envConf := make(map[string]string)
+		for _, row := range dbConfRows {
+			if _, isLocked := lockedConf[row.EnvKey]; !isLocked {
+				envConf[row.EnvKey] = row.EnvValue
+			}
+		}
```
yhabteab: Why is this transformation needed? The query used to retrieve these tuples already contains a `locked = n` filter.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,290 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+			return
+		}
+
+		for _, row := range dbConfRows {
+			_, isLocked := lockedConf[row.EnvKey]
+
+			client.logger.Debugw("Fetched Icinga Notifications configuration option from the database",
+				zap.String("key", row.EnvKey),
+				zap.Bool("key_locked", isLocked))
+		}
```
yhabteab: Same here. Given this `String` impl in `icingadb_config.go`:
```go
// String returns a string representation of the IcingadbConfig.
func (c *IcingadbConfig) String() string {
	return c.EnvKey + "=" + c.EnvValue
}
```

You can just do (after making `dbConfRows` of type `[]*v1.IcingadbConfig`):

```suggestion
		client.logger.Debugw("Fetched Icinga Notifications configuration option from the database",
			zap.Stringers("config", dbConfRows))
```

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,290 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+
+		nc, err := source.NewClient(newConf.Config, "Icinga DB "+internal.Version.Version)
+		if err != nil {
+			client.logger.Errorw("New Notifications client cannot be created from configuration", zap.Error(err))
+			return
+		}
```
yhabteab: Replacing the client entirely with each config change is an overkill IMHO. Just by checking whether something has changed or not using `reflect` isn't good for this. You could even replace the `runtimeConfig` with each config load without having to check whether something has changed first and I would be fine with it, but for the actual client, you need to make an explicit check on the config options relevant for it. For example, the `default_relations` config might change really often since users will be allowed to change it over the UI and that would cause the otherwise fine working client to be replaced, which isn't ideal.

Also, at the moment, the only config options that are going to be changed by Icinga DB Web are `notifications.url` and `notifications.default_relations`, so the only thing relevant for client replacement would only be `notifications.url` but to account for future Icinga DB Web changes, I would check all relevant ones as suggested above.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,290 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+		client.markConfigured()
+
+		client.logger.Info("Synchronized new Icinga Notifications configuration from the database")
+	}
+
+	select {
+	case <-client.syncedLockedConfCh:
+	case <-ctx.Done():
+		return ctx.Err()
+	}
```
yhabteab, line 512: Currently, the `SynchronizeLockedConfigToDatabase` spins on the following check endlessly:

```go
			if !environmentId.Valid() || !endpointId.Valid() {
				...
```

If Icinga 2 never send a real endpoint ID, that function will never return, and thus `syncedLockedConfCh` will never be closed. However, even though, I don't yet understand the reasoning behind it, you've agreed with @BastianLedererIcinga to always populate the config with the `endpoint_id` column set to null bytes for nothing because this function will hang on this select until the HA context is canceled.

---

Just to clarify, two Icinga DB instances running in HA mode aren't supposed to share the same config generated by Icinga DB Web. For this reason, the Icinga DB instances probe and discover their respective Icinga Notifications listener socket and persist the result in the `icingadb_isntance` table each on their own. Just because one instance has successfully discovered the listener socket doesn't mean the other instance has done so as well. Also, the socket path might differ between the two instances, so it is not guaranteed that the same socket path will be valid for both instances. Therefore, I'm not sure why you think it's a good idea to let Icinga DB Web populate the `endpoint_id` column with null bytes unconditionally as that would defeat the whole purpose of having that column in `icingadb_config` table in the first place.

For this reason, I will describe shortly how I visualize this process to work in a way that makes sense to me. In the very optimal case, Icinga 2 always sends a valid endpoint ID, which is then persisted in the `icingadb_instance` table by each Icinga DB instance accordingly. Icinga DB Web then reads `icingadb_instance` and populates the `icingadb_config` table with the correct endpoint ID for each instance. Icinga DB (we) then need to ensure that we only load the config matching our own instance's endpoint ID. That's the very easy case and the least case I expect to just work.

However, in the case where Icinga 2 doesn't send a valid endpoint ID (in this case it's guaranteed that we aren't running in HA mode), Icinga DB Web might be allowed to populate the `icingadb_config` table with null bytes[^1] for the `endpoint_id` column. However, Icinga DB (we) must also persist the locked config in the `icingadb_config` table with null bytes for the `endpoint_id` column as well and not spin endlessly in the `SynchronizeLockedConfigToDatabase` function till Icinga 2 eventually sends a valid endpoint ID. This case on its own is easily implemented as well, the queries just need to be adjusted a bit.

The only case that I don't have a good solution for is when the endpoint ID changes at runtime (from NULL to a valid endpoint ID and vice-versa or from one valid endpoint ID to another). How do we handle that? I've two possible options in mind but if you all have a better idea, then please suggest it.

- The first option is to simply perform an update on the `icingadb_config` table to change the `endpoint_id` to the new value. If the old value was NULL, then we might not want to do the update in place but rather clone each tuple having the `endpoint_id` set to null bytes and set the new `endpoint_id` value for the cloned tuples, but doing it in place would be fine for me as well. That way, the user wouldn't have to re-save the config in Icinga DB Web to have the new `endpoint_id` value reflected in the `icingadb_config` table. This applies to both cases, endpoint ID -> NULL, NULL -> endpoint ID, and endpoint ID -> other endpoint ID. This would be the most convenient option for the user, but it would also require a bit more work to implement. 
- The second option is to simply purge the `icingadb_config` table, de-configure the Icinga Notifications component, and require the user to re-save the config in Icinga DB Web. Though, this would be rather inconvenient for the user and I don't think it would be a good idea to do it that way too.

[^1]: We could also just set it to `environment_id` instead of null bytes without that much of a difference (same storage size, same behavior).

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,290 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+
+	performSync()
+
+	ticker := time.NewTicker(15 * time.Second)
```
yhabteab, line 516: Missing `defer ticker.Stop()` here.

pkg/notifications/notifications.go
```diff
@@ -138,6 +385,11 @@ func (client *Client) ClearIncidents() {
 // errors encountered during the submission of events to the Icinga Notifications API and continue processing the
 // remaining events, but never returns an error for those.
 func (client *Client) ApplyDelta(ctx context.Context, delta *icingadb.Delta) error {
+	if !client.IsConfigured() {
+		client.logger.Warn("Cannot apply delta as the Notifications Client is not yet configured")
```
yhabteab: I don't think, we should warn the user about this. It should just be a debug log. Same goes for all the `!client.IsConfigured()` usages in this file.

pkg/notifications/notifications.go
```diff
@@ -706,10 +980,14 @@ func (client *Client) Submit(ctx context.Context, entity database.Entity) error
...
 
-	attributes := client.DefaultRelations
 	return retry.WithBackoff(
 		ctx,
 		func(ctx context.Context) (err error) {
+			client.runtimeMu.RLock()
+			attributes := client.runtimeConfig.DefaultRelations
+			nc := client.notificationsClient
+			client.runtimeMu.RUnlock()
+
```
yhabteab: Moving the `client.DefaultRelations` assignment within the retryable function is not a good idea. Suppose we have the following scenario:

We initialize `attributes` with `client.DefaultRelations` and then call `ev.completeAndUpdate` within the loop and succeeds. So, we proceed to submitting the event, but `client.ProcessEvent` fails with `source.ErrAttrsNegotiation` and re-assigns the `attributes` variable to the newly received missing attributes. Now, when we again call `ev.completeAndUpdate`, it might fail due to various reasons, so it exits the loop and returns an error. However, since almost all errors are retryable, we'll come back to the retryable function again, and this time `client.DefaultRelations` will be assigned to `attributes` again, which is not what we want. So, keep it where it was please.

cmd/icingadb/main.go
```diff
@@ -242,6 +261,19 @@ func run() int {
...
+								select {
+								case <-notificationsSource.Configured():
+									logger.Info("Icinga Notifications became configured, restarting sync")
+									cancelSynctx()
+									return nil
+								case <-synctx.Done():
+									return synctx.Err()
+								}
+							})
+						}
```
yhabteab: This could actually be combined into the existing `<-dump.InProgress()` groroutine and doesn't need its own one.

yhabteab (changes requested):

internal/config/config.go
oxzi: > So, this new option is indeed required but it **must** be set to `true` by default, and users can explicitly opt-out if they don't want to use this automation.

As I argued before, this would change the security model. This is not just a theoretical matter, especially regarding the Notifications Configuration. An adversary could exfiltrate confidential files by populating the database with `notifications.url` and `notifications.password_file`. This includes all files the Icinga DB daemon user has access to, including `/etc/passwd`.

IMO, an opt-in is always the safer choice - also regarding different privacy or consumer protection laws. Why do you think this must be enabled by default? An end user needs to change the Icinga DB configuration anyway, why not trust them to change this flag as well?

internal/config/config.go
oxzi: Fair. Since this method was only called at one place, I have inlined it instead.

internal/config/config.go
oxzi: Moved the whole block down under the `RetentionConfig`, where it fits as `RetentionConfig` is used as the `Retention` field in `Config` before the `Notifications` field.

internal/config/config.go
oxzi: Good point, added a check.

internal/config/config.go
oxzi: Changed.

internal/config/config_test.go
oxzi: This is weird, as the tests are working on my machine, `go version go1.26.5 linux/amd64`. Nevertheless, I have removed the redundant whitespaces.

pkg/icingadb/ha.go
oxzi: I briefly thought about this during the implementation, but didn't wanted to touch too much. However, I have just checked what each mutex is protecting - these are quick read or write access - and have merged them. Leaving the world a better place; or so.

pkg/notifications/notifications.go
oxzi: Fair, renamed. The prior name emerged during the prior changes, but did not real represent what this method does.

pkg/notifications/notifications.go
oxzi: At the moment: Yes, this should not happen. However, after the next three refactoring sessions, there might be a race between HA and Notifications. And as our declared goal was to reduce the amount of crashes, I don't want to introduce new potential panics.

pkg/notifications/notifications.go
oxzi: :shrug: :shrimp:

pkg/notifications/notifications.go
oxzi: Changed.

pkg/notifications/notifications.go
oxzi: `lockedConf` contains the locked configuration from this node. The query asks the database to exclude rows with `locked = 'n'`. But there is no constraint to ensure that a misconfigured (or malicious) client inserts something as unlocked to the database which this node has locked. So, this check ensures that nothing which this client considers locked gets overwritten.

internal/config/config.go
nilmerg: The goal, and the reason why it must be opt-out, is to remove as many as possible barriers to set up notifications for existing and new users, _following our recommended approach_: single system, HTTP over UDS.

Your argument regarding `password_file` is sound, though I'd argue that, since we are aware of this, why shouldn't the location of said file be constrained to a specific directory?

But in reality, it's rather `password` and `key` to worry about. In a locally configured environment, I wouldn't be surprised if end users complain about this being insecure. But back to why it needs to be opt-out to begin with:

When following our recommended approach, there's no `password`, `password_file`, `cert` or `key`. There's no requirement yet to have such in the database and so let's stop here and just don't insert them there. The only things required are: `url` and `default_relations`.

pkg/notifications/notifications.go
oxzi: The `c.EnvValue` might contain sensitive information, including passwords. This would be a [CWE-532: Insertion of Sensitive Information into Log File](https://cwe.mitre.org/data/definitions/532.html). The `key_locked` field was added to allow debugging when some field was not applied regardless of its presence in the database.

pkg/notifications/notifications.go
oxzi: Sorry, but I cannot completely follow what is especially bugging you here.

Do you consider the `reflect.DeepEqual` calls to be too expensive? At the moment, this is executed every 15 seconds. To put this into perspective, I have just benchmarked this. The most expensive case would be a complete comparison, which is our "hot path", as in most cases the config was not updated. This takes round about 500ns on my machine. We are talking about 5e-7 seconds here, compared with 15 seconds.

<details><summary>Benchmark Details</summary>
<p>

```diff
diff --git a/internal/config/config_test.go b/internal/config/config_test.go
index 564eb350..12d2a78d 100644
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -1,6 +1,10 @@
 package config

 import (
+       "os"
+       "reflect"
+       "testing"
+
        "github.com/caarlos0/env/v11"
        "github.com/creasty/defaults"
        "github.com/icinga/icinga-go-library/config"
@@ -12,8 +16,6 @@ import (
        "github.com/icinga/icingadb/pkg/icingadb/history"
        "github.com/stretchr/testify/require"
        "go.uber.org/zap/zapcore"
-       "os"
-       "testing"
 )

 // testFlags is a struct that implements the Flags interface.
@@ -193,7 +195,7 @@ notifications:
   username: icingadb
   password: insecure
   synchronize_with_database: true
-                                       `,
+`,
                        },
                        Expected: &Config{
                                Database: database.Config{
@@ -278,3 +280,50 @@ func TestNotificationsConfig_StaticConfig(t *testing.T) {
        cmpCfg.Notifications.Config.Password = "icingadb" // Reset, as redacted in output
        require.Equal(t, cfg, cmpCfg.Notifications)
 }
+
+func BenchmarkNotificationsConfigEquals(b *testing.B) {
+       baseConf := NotificationsConfig{
+               Config: source.Config{
+                       Url:      "https://example.com",
+                       Username: "foo",
+                       Password: "bar",
+                       TlsOptions: config.TLS{
+                               TLSCommon: config.TLSCommon{
+                                       Enable: true,
+                                       Ca:     "/path/to/ca.crt",
+                               },
+                       },
+                       DefaultRelations: []string{"$.foo", "$.bar"},
+               },
+               SynchronizeWithDatabase: true,
+       }
+
+       b.Run("identical", func(b *testing.B) { // Hot path: unchanged/identical configuration
+               confA := baseConf
+               confB := baseConf
+
+               for b.Loop() {
+                       _ = reflect.DeepEqual(confA, confB)
+               }
+       })
+
+       b.Run("different", func(b *testing.B) { // Diff in field.
+               confA := baseConf
+               confB := baseConf
+               confB.Password = "insecure"
+
+               for b.Loop() {
+                       _ = reflect.DeepEqual(confA, confB)
+               }
+       })
+
+       b.Run("different deep", func(b *testing.B) { // Diff in deeper nested field.
+               confA := baseConf
+               confB := baseConf
+               confB.Config.TlsOptions.Ca = "/path/to/new/ca.crt"
+
+               for b.Loop() {
+                       _ = reflect.DeepEqual(confA, confB)
+               }
+       })
+}
```

```
$ go test -bench='.*NotificationsConfigEquals.*' ./internal/config/...
goos: linux
goarch: amd64
pkg: github.com/icinga/icingadb/internal/config
cpu: 13th Gen Intel(R) Core(TM) i7-1355U
BenchmarkNotificationsConfigEquals/identical-12                  2248630               523.7 ns/op
BenchmarkNotificationsConfigEquals/different-12                  4269543               287.5 ns/op
BenchmarkNotificationsConfigEquals/different_deep-12             2963762               397.2 ns/op
PASS
ok      github.com/icinga/icingadb/internal/config      3.592s
```

</p>
</details>

The config, on the other hand, is only being replaced when an actual change happens. How often might this be? Maybe a few times during the initial setups, usually with minutes in between.

While I second your argument about unused fields in web, I would rather not depend on implementation details over there. Otherwise, changes to web could break things.

Pardon, if I have misunderstood you, but for me, this seems like micro-optimizing a code branch which is not on the hot path.

yhabteab: > Pardon, if I have misunderstood you, but for me, this seems like micro-optimizing a code branch which is not on the hot path.

Yep. You actually missed the point I was trying to make. I don't actually care about the benchmarks of `reflect.DeepEqual` and other alternatives. I was trying to say that you're currently replacing the **entire source.Client** and thus its HTTP client too just because `reflect.DeepEqual` said something has changed in the config. But that something might be completely irrelevant for the `source.Client` itself like `default_relations` config option for example, and replacing the client just because the user has changed this option is, what I was referring to as an overkill.

pkg/notifications/notifications.go
oxzi: The null bytes for an empty `endpoint_id` are not related to this. They were used rather than a `NULL` value to allow a `UNIQUE` constraint (now simplified as a primary key) as `NULL` is not unique in SQL.

> Just to clarify, two Icinga DB instances running in HA mode aren't supposed to share the same config generated by Icinga DB Web. For this reason, the Icinga DB instances probe and discover their respective Icinga Notifications listener socket and persist the result in the `icingadb_isntance` table each on their own.

This information should be relevant for :web:, as :web: wanted to populate this with an empty value - at least back when we talked the last time. And, later, also for this PR, as the `SELECT` query needs to be adjusted.

> For this reason, I will describe shortly how I visualize this process to work in a way that makes sense to me. […]

In general, this all sounds reasonable to me. But, let's be realistic, how many scenarios are there without an Endpoint ID? Wouldn't it be easier to just assert that it exists?

pkg/notifications/notifications.go
oxzi: Not since Go 1.23, see the [API doc comment](https://pkg.go.dev/time#NewTicker). I have left it out on purpose.

pkg/notifications/notifications.go
oxzi: Changed.

pkg/notifications/notifications.go
oxzi: I have moved it there because otherwise a config change of the default relations during a `retry.WithBackoff` operation would get ignored. I think this is a trade off: What would you prefer: A potential round trip or not using the latest requested default relations?

pkg/notifications/notifications.go
yhabteab: Oh, I thought the Go 1.23 thing was only related to ticker.Reset, so didn't even look at the API doc and still use defer the ticker.Stop everytime I use a ticker.

cmd/icingadb/main.go
oxzi: Changed.

pkg/notifications/notifications.go
yhabteab: > They were used rather than a `NULL` value to allow a `UNIQUE` constraint (now simplified as a primary key) as `NULL` is not unique in SQL

But why not use, for example, the environment ID for that instead of null bytes?

> > Just to clarify, two Icinga DB instances running in HA mode aren't supposed to share the same config generated by Icinga DB Web. For this reason, the Icinga DB instances probe and discover their respective Icinga Notifications listener socket and persist the result in the `icingadb_isntance` table each on their own.
> 
> This information should be relevant for :web:, as :web: wanted to populate this with an empty value - at least back when we talked the last time. And, later, also for this PR, as the `SELECT` query needs to be adjusted.

I'm not sure I can follow here, but what exactly wanted the web colleagues to populate with empty value? My wall of text wasn't necessarily only about this specific diff context but the general idea how we want to proceed with this, and that's why I've also explicitly mentioned Bastian and maybe @nilmerg can say something about this too because that's a kinda not that obvious decision we need to make.

> In general, this all sounds reasonable to me. But, let's be realistic, how many scenarios are there without an Endpoint ID? Wouldn't it be easier to just assert that it exists?

I mean, the nullability of that column exists for a reason, and changing that requirement now would be not something I do just for this purpose.

pkg/notifications/notifications.go
yhabteab: > A potential round trip or not using the latest requested default relations?

I would opt for the latter because the default ones should be relevant only for the initial request but not for any subsequent ones because when Icinga Notifications misses something that are now in the `default_relations`, then Icinga Notifications should have sent as all columns it misses in the response. So, restarting that negotiation again makes little sense IMHO.

pkg/notifications/notifications.go
yhabteab: See https://go.dev/wiki/CodeReviewComments#declaring-empty-slices

internal/config/config_test.go
yhabteab: Idk 🤷.

pkg/notifications/notifications.go
yhabteab: > And as our declared goal was to reduce the amount of crashes, I don't want to introduce new potential panics.

But what's the user supposed to do with these warnings if such a case is expected to happen due to a race condition in the Icinga DB code?

pkg/notifications/notifications.go
yhabteab: > But there is no constraint to ensure that a misconfigured (or malicious) client inserts something as unlocked to the database which this node has locked.

There is no such a client. The only client of this is Icinga DB Web and I wouldn't call it _malicious_. If we can't even trust the result of our database queries what else should we trust then?

pkg/notifications/notifications.go
yhabteab: > The `c.EnvValue` might contain sensitive information, including passwords. This would be a [CWE-532: Insertion of Sensitive Information into Log File](https://cwe.mitre.org/data/definitions/532.html).

Ok, that's a valid point. But still I don't see a reason for going through each individual key-value pairs and produce separate log messages. So, I would either change the string impl to only include the keys or not log anything at all here.

internal/config/config.go
oxzi: > The goal, and the reason why it must be opt-out, is to remove as many as possible barriers to set up notifications for existing and new users, _following our recommended approach_: single system, HTTP over UDS.

Sorry, but I cannot follow this argument. Users already have to edit the configuration file: at least for the database. So, is setting one flag opting in a feature that changes the roles that much?

> Your argument regarding `password_file` is sound, though I'd argue that, since we are aware of this, why shouldn't the location of said file be constrained to a specific directory?

I cannot think of any "safe" path. Some users might consider something under `/etc/icingadb/`, others are using a password manager populating something under `/run`, `/var/run`, or a home directory.

> But in reality, it's rather `password` and `key` to worry about. In a locally configured environment, I wouldn't be surprised if end users complain about this being insecure. But back to why it needs to be opt-out to begin with:
> 
> When following our recommended approach, there's no `password`, `password_file`, `cert` or `key`. There's no requirement yet to have such in the database and so let's stop here and just don't insert them there. The only things required are: `url` and `default_relations`.

I think you have misunderstood me: With this change, the database is no longer just a data sink, but also a source for Icinga DB. So, anyone with access to the database can now instruct the Icinga DB daemon to exfiltrate files. Such a change with an update is CVE-worthy, IMO.

internal/config/config.go
oxzi: Second thought: This whole `if` can be dropped as there are no pointer fields in `source.Config`.

pkg/notifications/notifications.go
oxzi: Well, in case of some race, it will magically work at some later point and this is only logged once or twice, and nobody noticed it. Otherwise, the log message will get annoying.

pkg/notifications/notifications.go
oxzi: > The only client of this is Icinga DB Web and I wouldn't call it _malicious_.

:raised_eyebrow: 

Nevertheless, I would argue that my point is still valid. Without such a check, we do not actually lock the configuration, but are relying on the database that nobody wrote a row regarding a key we are holding locked.

Consider the following:


| environment_id | endpiont_id       | env_key | env_value | locked |
|----------------|-------------------|---------|-----------|--------|
| 0xAAAAAA       | 0x01 (this node)  | FOO     | bar       | y      |
| 0xAAAAAA       | 0x00 (null bytes) | FOO     | buz       | n      |

The first row was inserted by us, the second either by web, from an adventurous sysadmin (, or an adversary). Without this check, `FOO` would be "buz" regardless of the key being locked.

yhabteab: > The first row was inserted by us, the second either by web, from an adventurous sysadmin (, or an adversary). Without this check, `FOO` would be "buz" regardless of the key being locked.

Well, obviously, the query should already filter out everything that doesn't match our `0x01` endpoint ID.

pkg/notifications/notifications.go
oxzi: With the String implementation, I would loose the information which key was locked.

pkg/notifications/notifications.go
oxzi: This was more straight forward, thanks.

Unless I am missing something, there is only the `default_relations` field unused in the `source.Client`. So, we could skip this step if and only if `default_relations` is the only changing field. However, this would rely on implementation knowledge of `source.Client`, which does not use `Config.DefaultRelations`. This is not obvious from an API perspective, where the whole `source.Config` struct is passed to `source.NewClient`.

Because updates are rare, I would stay with my prior argument, and prefer a restarting Client once a while before introducing another check completely relying on implementation knowledge, which might get outdated.

pkg/notifications/notifications.go
oxzi: > > The first row was inserted by us, the second either by web, from an adventurous sysadmin (, or an adversary). Without this check, `FOO` would be "buz" regardless of the key being locked.
> 
> Well, obviously, the query should already filter out everything that doesn't match our `0x01` endpoint ID.

Then, this is highly related to <https://github.com/Icinga/icingadb/pull/1158#discussion_r3720839281>. So far, my assumption was that web was inserting an empty (first `NULL`, then null bytes) `endpoint_id`. With your comment, this cannot happen any longer, unless there is an empty `endpoint_id` due to a single-node setup.

While then the primary key (`environment_id`, `endpoint_id`, and `env_key`) would forbid this, we would again rely on the potentially updated database instead of our locally sourced values. Btw, we could get rid of this (without me whining any longer about an untrusted database) with excluding the locked `env_keys` in the `SELECT` query.

oxzi: I am going to write something together, as it seems we are not really in sync. Furthermore, I am going to change the implementation as I have outlined, also addressing https://github.com/Icinga/icingadb/pull/1158#discussion_r3757325022.

pkg/notifications/notifications.go
oxzi: You are right, changed it.

internal/config/config.go
nilmerg: The database you speak of must be the Icinga DB and of course this is a mandatory manual step. But for a user not familiar with the product, it is reasonable to expect this, since it is it's primary job. The new notification integration is not that obvious let alone this exotic flag. That's what I described as barrier prior.

I did understand it and my proposal was to limit what's put into the *sink* so that no secrets get disclosed by accident. You can't reasonably argue that `url` and `default_relations` are secrets, do you?

pkg/notifications/notifications.go
oxzi: Implemented this approach.

pkg/notifications/notifications.go
oxzi: Implemented this together with <https://github.com/Icinga/icingadb/pull/1158#discussion_r3758354336>.

pkg/notifications/notifications.go
oxzi: After a discussion with @yhabteab and @nilmerg, we have decided to go with @yhabteab's first suggestion: If there is a non empty endpoint ID, Icinga DB's Notifications Client tracks this ID and `UPDATES` the rows if it changes. If there is no endpoint ID, use the null bytes, `v1.UnconfiguredEndpointId()`.

internal/config/config.go
oxzi: I restricted the synchronization to those two fields and enabled the sync by default.

pkg/notifications/notifications.go
oxzi: Implemented this change as described.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,344 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+}
+
+// endpointId returns this node's endpoint ID or the null endpoint ID, if unconfigured.
+func (client *Client) endpointId() types.Binary {
+	endpointId := client.ha.EndpointID()
+	if endpointId.Valid() {
+		return endpointId
+	}
+
+	return v1.UnconfiguredEndpointId()
```
yhabteab, line 249: This function is also used for the initial static config sync and assumes that the previously used endpoint ID is still the same (after Icinga DB is restarted). AFIAK, `client.ha.EndpointID()` returns just the endpoint ID received at runtime and doesn't pre-load the previous one from the `icingadb_instance` table, causing the cleanup ops done in `PersistLockedConfigOnce` to be a no-op.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,344 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+		if err != nil {
+			client.logger.Errorw("New Icinga Notifications configuration from the database cannot be parsed", zap.Error(err))
+			return
+		}
+
+		err = newConf.Validate()
+		if err != nil {
+			client.logger.Errorw("New Icinga Notifications configuration from the database is invalid", zap.Error(err))
+			return
+		}
```
yhabteab: Shouldn't all this be inside the `len(envKeys) > 0` check too? You just need to declare `newConfig` above that if check.

pkg/notifications/notifications.go
yhabteab: Fine. But I'm still not convinced by this, just for the record.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,344 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+
+	select {
+	case <-client.persistedLockedConfigCh:
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+
+	performSync()
+
+	ticker := time.NewTicker(15 * time.Second)
```
yhabteab, line 516: Since the argument for https://github.com/Icinga/icingadb/pull/1158/changes#r3757737767 was that these config shouldn't change that often, why not use here a bit bigger interval then? Like 30s or even 1m?

pkg/notifications/notifications.go
```diff
@@ -276,10 +586,14 @@ func (client *Client) SyncCheckOutputs(ctx context.Context) error {
 			filter["service"] = nil
 		}
 
+		client.runtimeMu.RLock()
+		nc := client.notificationsClient
+		client.runtimeMu.RUnlock()
+
```
yhabteab: You need to move this to within the retryable function just like you did for the `client.Submit` method.

yhabteab (changes requested):

config.example.yml
```diff
@@ -170,3 +170,7 @@ redis:
 #  default_relations:
 #    - '$.host.vars'
 #    - '$.services[*].vars'
+
+  # Synchronize the configuration with the database. This allows Icinga Notifications Web to configure this source.
+  # If set, all other fields can be left empty. However, if set, they take precedence.
+#  synchronize_with_database: true
```
nilmerg, line 180: Other sections indicate their default with

> `# Defaults to <value>`

Please do so here as well.

nilmerg (changes requested):
Quickly asked @BastianLedererIcinga what Icinga DB Web is supposed to do in case `synchronize_with_database: false` and it turned out no-one thought about this yet. It makes absolutely no sense to offer to configure the notifications source while it has no effect.

So my spontaneous idea is to simply synchronize at the very least `synchronize_with_database` with `icingadb_config`. Or drop the flag altogether, as it makes less sense now anyway with what's synchronized now.

pkg/notifications/notifications.go
oxzi: The retryable function in `PersistsLockedConfigOnce` starts by checking if HA has synced and is populated.

https://github.com/Icinga/icingadb/blob/51254b388de51e4f985659beefe246edd6dd5aeb/pkg/notifications/notifications.go#L206-L208

The HA code populates the environment ID and endpoint ID one after another, fortunately starting with the endpoint ID. So, `PersistsLockedConfigOnce` should not work with `v1.UnconfiguredEndpointId()` as an invalid fallback.

https://github.com/Icinga/icingadb/blob/51254b388de51e4f985659beefe246edd6dd5aeb/pkg/icingadb/ha.go#L233-L264

pkg/notifications/notifications.go
oxzi: All of this would mostly be noops with empty `envKeys`. Nevertheless, I have moved it in the `if len(envKeys) > 0` above.

pkg/notifications/notifications.go
oxzi: No real reason whatsoever, just an arbitrary number that seemed to fit. From an user's perspective, the change should be immediately after they click something in web. In the end, it's a tradeoff and I don't wanted to give users the bad experience of having to wait too long for a change to be applied, as otherwise they might get confused and reconfigure something before the prior config was even applied.

config.example.yml
oxzi: The "defaults" hint is not always present in the config and carries no additional information, imo. The line you marked already says its `true` by default - and so does the docs.

oxzi:
Pushed an updated version, trying to address both @yhabteab's and @nilmerg's comments. Furthermore, some locking issues were addressed.

---

> Quickly asked @BastianLedererIcinga what Icinga DB Web is supposed to do in case `synchronize_with_database: false` and it turned out no-one thought about this yet. It makes absolutely no sense to offer to configure the notifications source while it has no effect.
> 
> So my spontaneous idea is to simply synchronize at the very least `synchronize_with_database` with `icingadb_config`. Or drop the flag altogether, as it makes less sense now anyway with what's synchronized now.

`icingadb_instances.notifications_synchronize_with_database` was added, mirroring the configured `notifications.synchronize_with_database` value.

config.example.yml
nilmerg: Then this begs the question to me whether it should be added to those sections as well. To me, it seems that only the `notifications:` section differs and does not document it's defaults consequently. For example, is `http://localhost:5680` the default for `url`? I expect it to not to. But the same expectation is not met with `synchronize_with_database: true`, which is my entire point. The documentation might tell me this, but that's a different file and thus a different context.

pkg/notifications/notifications.go
yhabteab: > The HA code populates the environment ID and endpoint ID one after another, fortunately starting with the endpoint ID.

And that's exactly my point. That HA code doesn't pre-load previously used endpoint ID from the database but only populates it what it read from Redis. However, suppose we have the following situation:

1. Icinga DB is running fine and uses **X** as its endpoint ID.
2. Icinga DB is stopped for whatever reason for lets say 5 minutes
3. Now, within that 5m timespan the user renames its endpoint name and restarts Icinga 2 service
4. So, Icinga 2 re-dumps its configuration and sends **Y** as endpoint ID now.
5. Lastly, Icinga DB is started again and uses **Y** as its previous and current endpoint ID making the cleanup statement in `persistLockedConfig` useless because there's no tuples with enpoint ID **Y** stored in the database.

oxzi: You have a valid point, thanks for clarifying your argument.

However, I am unsure how to properly address this. When an Icinga DB node starts with a changed endpoint ID, it could see that the `icingadb_instance` table does not contain its ID. However, it cannot be certain that some other entry is its prior ID. There is no transition log. Furthermore, the HA retention will clean up this table later on.

In general, I would argue that the endpoint ID does not change regularly, or even at all. Introducing another cross reference between the HA code - which I would see as both crucial and fragile - for a probabilistic cleanup is not worth the potential bugs. I have play with some ideas and did not liked any of those. So, I would keep it this way and just consider that a changing Endpoint ID requires some manual cleanup. This is nothing which goes unnoticed, as it is displayed in web.

config.example.yml
oxzi: > To me, it seems that only the `notifications:` section differs and does not document it's defaults consequently.

Valid point. I guess, it was at some point in the past (for a development version). I am going to unify this with the rest of the example config.

pkg/notifications/notifications.go
yhabteab: I don't understand which cross-reference you are referring to. There is quite a trivial to implement solution to this problem. The cleanup statement can be adjusted to left join (anti-join) on the `icingadb_instance` table and remove any entries that do not have a corresponding entry in the `icingadb_instance` table. No cross-reference or probabilistic cleanup is needed.

> This is nothing which goes unnoticed, as it is displayed in Web.

How would Web notice this? And even if it does somehow, it's not allowed to remove anything from that table, especially not locked entries.

schema/pgsql/upgrades/add-icingadb_config.sql
```diff
@@ -0,0 +1,21 @@
+ALTER TABLE icingadb_instance ADD COLUMN notifications_synchronize_with_database boolenum NOT NULL DEFAULT 'n';
+ALTER TABLE icingadb_instance ALTER COLUMN notifications_synchronize_with_database DROP DEFAULT;
```
yhabteab: Please do not drop the default. See all existing boolenum columns for example. This is required because in MySQL even though the columns are marked as `NOT NULL`, MySQL will just use the first value of that enum if you don't provide a value for it and won't complain about missing value for such columns. So, to make that consistent, all the boolenum columns in PostgreSQL have the default value. So, add the default value in the normal schema file as well.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,379 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+			if err != nil {
+				client.logger.Errorw("New Icinga Notifications configuration from the database is invalid", zap.Error(err))
+				return
+			}
+		}
+
+		if newConf.Url == "" {
+			client.logger.Debug("Postponing Notifications config synchronization step as no URL is configured")
+			return
+		}
```
yhabteab: If I'm not mistaken, Icinga DB Web will unset the `notifications.url` in the database if the user wants to just disable the component. However, even if you're detecting and ignoring this here, but are ignoring the case that the client might still be configured here. So, you've to de-configure it as well, so that it no longer tries make any requests to Icinga Notifications anymore.

pkg/notifications/notifications.go
yhabteab: They have to wait for it to be applied anyway since this isn't applied instantly but Ok.

yhabteab (changes requested):

pkg/notifications/notifications.go
oxzi: Added a cleanup query using an anti-join like query.

schema/pgsql/upgrades/add-icingadb_config.sql
oxzi: Changed.

pkg/notifications/notifications.go
oxzi: This was a bit more than expected, but should work now.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,433 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+			})
+		},
+		func(err error) bool {
+			if errors.Is(err, errNotYetPopulated) {
+				return true
+			}
+			return retry.Retryable(err)
+		},
+		backoff.DefaultBackoff,
+		retrySettings)
```
yhabteab, line 362: Can you please use your own settings here and ignore the `errNotYetPopulated` case? I don't think, it's worth logging that as a warning with a very ugly stack trace like this.

```bash
icingadb-1  | 2026-09-02T08:58:09.767+0200	WARN	database	Can't execute query. Retrying	{"error": "notifications client is not yet populated", "errorVerbose": "notifications client is not yet populated\ngithub.com/icinga/icingadb/pkg/notifications.(*Client).persistLockedConfig\n\t/go/src/github.com/Icinga/icingadb/pkg/notifications/notifications.go:284\ngithub.com/icinga/icingadb/pkg/notifications.(*Client).PersistLockedConfigOnce.func1\n\t/go/src/github.com/Icinga/icingadb/pkg/notifications/notifications.go:270\nsync.(*Once).doSlow\n\t/usr/local/go/src/sync/once.go:78\nsync.(*Once).Do\n\t/usr/local/go/src/sync/once.go:69\ngithub.com/icinga/icingadb/pkg/notifications.(*Client).PersistLockedConfigOnce\n\t/go/src/github.com/Icinga/icingadb/pkg/notifications/notifications.go:269\nmain.run.func5\n\t/go/src/github.com/Icinga/icingadb/cmd/icingadb/main.go:190\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_arm64.s:1039", "after": "1.292µs", "attempt": 1}
```

schema/pgsql/upgrades/add-icingadb_config.sql
```diff
@@ -0,0 +1,20 @@
+ALTER TABLE icingadb_instance ADD COLUMN notifications_synchronize_with_database boolenum NOT NULL DEFAULT 'n';
+
+CREATE TABLE icingadb_config (
+  environment_id bytea20 NOT NULL,
+  endpoint_id bytea20 NOT NULL DEFAULT '\x0000000000000000000000000000000000000000',
+
+  env_key varchar(255) NOT NULL,
+  env_value text NOT NULL,
+
+  locked boolenum NOT NULL,
```
yhabteab: Missing DEFAULT clause. Same goes to the `schema.sql` file.

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,433 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+	runtimeMu sync.RWMutex
+	// runtimeConfig is the currently loaded configuration.
+	runtimeConfig config.NotificationsConfig
+	// notificationsClient is the API client, configured via SynchronizeConfigWithDatabase.
+	notificationsClient *source.Client
+	// clientCtx is canceled when notificationsClient is replaced or removed; see Client.apiClient.
+	clientCtx    context.Context
+	clientCancel context.CancelFunc
+	// configuredCh is closed while notificationsClient is available; see Client.setClient.
+	configuredCh chan struct{}
```
yhabteab: Since I found this way too confusing and not that easy to track down what is going on, I've played around with the code and pushed a few changes (https://github.com/Icinga/icingadb/compare/notifications-config-from-database...yhabteab/notifications-config-from-db) to make this a bit more clear and easier to follow. Especially with the four ephemeral fields here guarded by the same mutex, it's not immediately clear how they depend on each other and how they are used. The commit explains the changes in more detail, but the main idea is to outsource these ephemeral fields into a separate struct, which can be used to atomically exchange the entire struct at once. Please have a look at the commit and let me know what you think. If you like the idea, you can cherry-pick that into your branch however you like.

yhabteab (changes requested):

pkg/notifications/notifications.go
```diff
@@ -638,6 +1068,13 @@ func (client *Client) buildAcknowledgementHistoryEvent(ctx context.Context, h *v
 // Note that this function is used as [icingadb.RUUpsertFunc] for the runtime updates pipeline, so its signature must
 // match the [icingadb.RUUpsertFunc] type.
 func (client *Client) Submit(ctx context.Context, entity database.Entity) error {
+	nc, clientCtx, stopClientCtx := client.apiClient(ctx)
+	if nc == nil {
+		client.logger.Debug("Cannot submit event as the Notifications Client is not yet configured")
```
yhabteab: I would also remove this log as it doesn't really say anything special that is not already known. Otherwise, this spams the debug log, making just useless as it will be logged for each and every processed runtime update event.

pkg/notifications/notifications.go
oxzi: Thanks, in general, this looked good. However, with `synchronize_with_database` defaulting to `true`, it broke a YAML/env config.

1. `NewNotificationsClient` populates `client.currentAPIClient` with the given `cfg`.
   https://github.com/Icinga/icingadb/blob/be933f87d56dbe83624809730ebe38460bb86c27/pkg/notifications/notifications.go#L164
2. `performSync` loads `client.initialConfig`, which is set to the same `cfg` in `NewNotificationsClient`.
   https://github.com/Icinga/icingadb/blob/be933f87d56dbe83624809730ebe38460bb86c27/pkg/notifications/notifications.go#L152-L157
   https://github.com/Icinga/icingadb/blob/be933f87d56dbe83624809730ebe38460bb86c27/pkg/notifications/notifications.go#L443
4. Those two identical values are being compared: no diff, no config.
   https://github.com/Icinga/icingadb/blob/be933f87d56dbe83624809730ebe38460bb86c27/pkg/notifications/notifications.go#L475-L477

Otherwise, looked good - applied, thanks :)

pkg/notifications/notifications.go
```diff
@@ -69,46 +72,433 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
+				envConf[row.EnvKey] = row.EnvValue
+			}
+
+			err = env.ParseWithOptions(&newConf, env.Options{Prefix: "ICINGADB_NOTIFICATIONS_", Environment: envConf})
+			if err != nil {
+				client.logger.Errorw("New Icinga Notifications configuration from the database cannot be parsed", zap.Error(err))
+				return
+			}
+
+			err = newConf.Validate()
```
oxzi, line 480: FFS, this is https://github.com/Icinga/icinga-go-library/pull/227 once again.

```
ERROR   notifications   New Icinga Notifications configuration from the database is invalid     {"error": "both password and password file are set", "errorVerbose": "both password and password file are set\ngithub.com/icinga/icinga-go-library/config.LoadPasswordFile\n\tgithub.com/icinga/icinga-go-library@v0.9.1-0.20260902082926-b0052fa488ce/config/config.go:390\ngithub.com/icinga/icinga-go-library/notifications/source.(*Config).Validate\n\tgithub.com/icinga/icinga-go-library@v0.9.1-0.20260902082926-b0052fa488ce/notifications/source/config.go:74\ngithub.com/icinga/icingadb/pkg/notifications.(*Client).SynchronizeConfigWithDatabase.func2\n\tgithub.com/icinga/icingadb/pkg/notifications/notifications.go:475\ngithub.com/icinga/icingadb/pkg/notifications.(*Client).SynchronizeConfigWithDatabase\n\tgithub.com/icinga/icingadb/pkg/notifications/notifications.go:509\nmain.run.func7\n\tgithub.com/icinga/icingadb/cmd/icingadb/main.go:222\nruntime.goexit\n\truntime/asm_amd64.s:1264"}
```

When configuring `password_file` via YAML/env, the `Password` field gets populated, and the re-verification fails.

pkg/notifications/notifications.go
```diff
@@ -64,51 +76,450 @@ func (ev *fetchableEvent) completeAndUpdate(ctx context.Context, attributes []st
...
-	return &Client{
-		Config: cfg,
+	prev := client.currentAPIClient.Swap(newSession)
+	if prev.cancel != nil {
+		prev.cancel(errClientReset)
+	}
 
-		db:     db,
-		logger: logger,
+	if client.reconfiguredCh != nil {
```
yhabteab: I think, it makes sense to restart the sync loop only we have a valid client, thus please add an additional `&& withClient` check as discussed offline.

nilmerg (dismissed):

yhabteab (approved):
LFTM now ..😮‍💨. Anything discovered later can be fixed in a follow-up PR.

oxzi:
~~Someone with enough permissions needs to clean up the list of required actions before this PR can be merged; see the updated list of MySQL jobs.~~ Nevermind, I have all the rights :smiling_imp:

Issue #1112, opened 2026-04-13
Correlate Icinga Notifications history with our own


nilmerg:
We want to show Icinga Notifications history entries alongside our own history entries. Icinga Notifications v1.0 will provide a new API endpoint [`GET /notifications?since=<timestamp>`](https://github.com/Icinga/icinga-notifications/issues/415) for this.

### Database

Since there's already a `notification_history` table, I'd go with `alert_history`.

Note that this isn't intended as 1:1 relation with `history`, but as `1:n` (`history.id` <-> `alert_history.history_id`) relation. The linked `history` row is supposed to be the event that is identified by `event_id`.

### Visualization Related Changes

Icinga DB Web needs to be able to join `alert_history` on demand. For this a new column `history.alert_count` is required. i.e. The total amount of `alert_history` rows linked to it. The goal is to be able to easily determine, based on a particular `history` row, if any related entries in `alert_history` exist. And because the `alert_history` relation is not always joined, the number of entries must be derived from this column as well.

julianbrost:
> The goal is to be able to quickly determine, based on a particular history entry, how many related entries in `alert_history` exist. i.e. The total amount of `alert_history` rows linked to it.

As https://github.com/Icinga/icingadb-web/issues/1359 says "The table `alert_history` is not joined at all here.", this basically means that a join is not sufficient to "quickly determine" that, i.e. this is asking for a `history.alert_count` column?

nilmerg:
Updated the description, I hope it paints a clearer picture now.

yhabteab:
Since there seems to be a misunderstanding regarding how Icinga Notifications generate the event queue ID to deduplicate events, I will try to clarify this and describe the problem we currently have with this event correlation issue.

Since we've added HA support to Icinga Notifications, the [respective issue for that PR](https://github.com/Icinga/icinga-notifications/issues/394) had a requirement to provide a source to optionally submit a unique event queue ID on its own. This was kinda implemented in the PR, but not necessarily in a way that issue was describing/expecting it. Even though sources can provide a unique event queue ID now, Icinga Notifications still generates its own event queue ID based on the entire event payload (including the source-provided event queue ID). Apparently, this is not what @nilmerg expected, and was surprised about this behavior. This is of course a valid point, because I should have clarified this with him before approving and merging the PR, but I didn't. So, I apologize for that.

[...]

PR #1164, opened 2026-08-13
Event ID gets generated from UUID.new()

Jan-Schuppik:
Previously the event ID was built from the object name plus the current timestamp, which is not stable across notifications for the same object. Now as workaround until https://github.com/Icinga/icingadb/issues/1112 is fixed, we create the eventID simply with uuid.new().

Depends on:
- https://github.com/Icinga/icinga-go-library/pull/229


pkg/notifications/notifications.go
```diff
@@ -462,11 +463,9 @@ func (client *Client) buildCommonEvent(
 		}
 	}
 
-	objectId := objectName + fmt.Sprintf("!%d", time.Now().UnixMilli())
-
 	return &fetchableEvent{
 		Event: &event.Event{
-			ID:   objectId,
+			ID:   types.MakeUUID(uuid.New()), // build a random uuid until we fix the event ID generation in icinga2
```
yhabteab: You can just use https://github.com/Icinga/icingadb/issues/1112 as a reference instead since Icinga DB is the one having problems reconstructing the event ID and not Icinga 2.

yhabteab (changes requested):
> which is not stable across notifications for the same object. Now as workaround we create the eventID simply with uuid.new().

That's not why this woraround is needed. This woraround is needed only till we fix this issue:
- https://github.com/Icinga/icingadb/issues/1112

yhabteab:
@cla-bot check.

yhabteab (approved):

PR #1171, opened 2026-08-27
notifications: log the latest error not the previous one

yhabteab:
This was an oversight from me. The `OnRetryableError` callback logs something only when the last and current error don't match but the log entry included always the previous error instead of the current one.


yhabteab:
Can't merge it without an approval, so please review :).

oxzi (approved):

PR #1175, opened 2026-09-04
notifications: Mitigate startup configuration race

oxzi:
Fix a race with the notifications client during Icinga DB startup which might result in an unconfigured notifications subsystem.

The notifications client is started in parallel with the Icinga 2 config dump in the main function, and the main HA loop. Within the loop, the variable assignment to `notificationsConfigured` might occur before the notifications client became ready, marking notifications as unconfigured.

While the notifications client signals its readiness state via `reconfiguredCh`, its receiver is currently blocked with processing the Icinga 2 config dump. The implementation would then fallback to a default case, dropping the signal.

By making `reconfiguredCh` a buffered channel with a capacity of one, the notifications client will get configured once upon the main HA loop restarts.


yhabteab (approved):
Noice.

PR #1178, opened 2026-09-08
go fix: Modernize Codebase

oxzi:
## go fix: Modernize Codebase
Start with bumping the IGL to its latest version, which now transitively requires Go 1.27.

The go fix command started using nested struct field selectors.

## GHA: Pin Go Version to stable
Because of unknown reasons, the "1.x" Go version resolved to the outdated Go version 1.26.7, not even the latest dot version for 1.26. So, switched to "stable" as used in the IGL.


yhabteab (approved):
LFTM.

PR #1184, opened 2026-09-15
icingadb-migrate: reject inverted IDO time range

### LINKED ISSUE #1157: icingadb-migrate accepts inverted ido.from and ido.to ranges and exits successfully
## Describe the bug

`icingadb-migrate` accepts an invalid configuration where `ido.from` is newer than `ido.to`.

The two timestamps are parsed without validating their order. The migration then converts them independently into ID boundaries and uses those boundaries in queries based on `BETWEEN`.

With an inverted range, point-in-time history types such as state changes can therefore produce an empty result set. The migration still continues and exits successfully instead of reporting an invalid configuration.

Example:

```yaml
ido:
  from: 1767222000 # 2026-01-01 00:00:00 CET
  to:   1665999130 # 2022-10-17 11:32:10 CEST
```

## To Reproduce

1. Prepare a valid upstream `icingadb-migrate` configuration with working IDO and Icinga DB connections and a valid Icinga environment ID.

2. Configure an inverted IDO time range:

   ```yaml
   icinga2:
     env: "<valid-environment-id>"

   ido:
     type: mysql
     host: localhost
     database: icinga
     user: icinga
     password: "<password>"
     from: 1767222000
     to: 1665999130

   icingadb:
     type: mysql
     host: localhost
     database: icingadb
     user: icingadb
     password: "<password>"
   ```

3. Run the upstream migration tool with a new cache directory:

   ```bash
   icingadb-migrate \
     -c /tmp/icingadb-migration.yml \
     -t /tmp/icingadb-migration-cache
   ```

4. Observe that the invalid range is accepted. The migration computes zero eligible rows for point-in-time history types, performs no effective migration for them, and exits with status code `0`.

   ```bash
   echo $?
   ```

   Observed result:

   ```text
   0
   ```

## Expected behavior

`icingadb-migrate` should reject the configuration before connecting to the databases or computing ID boundaries when:

```text
ido.from > ido.to
```

It should exit with a non-zero status and print a clear error, for example:

```text
invalid IDO migration time range: from (1767222000) must not be newer than to (1665999130)
```

No migration or progress update should be attempted for an invalid range.

## Your Environment

* Icinga DB version: `1.4.0`
* Icinga 2 version: `2.15.1`
* Operating System and version: Red Hat Enterprise Linux 8

## Additional context

This report concerns only the upstream validation of `ido.from` and `ido.to`.

The upstream configuration declares both fields as independent `int32` values:

https://github.com/Icinga/icingadb/blob/v1.4.0/cmd/icingadb-migrate/main.go#L43-L56

`parseConfig()` decodes the YAML and returns the configuration without checking the ordering of the two timestamps:

https://github.com/Icinga/icingadb/blob/v1.4.0/cmd/icingadb-migrate/main.go#L118-L139

`computeIdRange()` then converts `From` and `To` independently into primary-key boundaries:

https://github.com/Icinga/icingadb/blob/v1.4.0/cmd/icingadb-migrate/main.go#L232-L267

The resulting boundaries are later used in a query with:

```sql
WHERE <id column> BETWEEN <fromId> AND <toId>
```

When the boundaries are inverted, this can produce an empty result set without being treated as an error:

https://github.com/Icinga/icingadb/blob/v1.4.0/cmd/icingadb-migrate/main.go#L270-L345

The current `main` branch still follows the same validation and ID-range calculation flow:

https://github.com/Icinga/icingadb/blob/main/cmd/icingadb-migrate/main.go

The migration documentation correctly states that `from` and `to` define the input time range and recommends double-checking the cutoff timestamp. This issue is specifically about the CLI accepting a semantically invalid range after configuration parsing:

https://github.com/Icinga/icingadb/blob/main/doc/06-Migration.md#input-time-range

A validation immediately after parsing the configuration would prevent the silent no-op, for example:

```go
if c.IDO.From > c.IDO.To {
    _, _ = fmt.Fprintf(
        os.Stderr,
        "invalid IDO migration time range: from (%d) must not be newer than to (%d)\n",
        c.IDO.From,
        c.IDO.To,
    )
    return nil, 2
}
```

SameerVers3:
`icingadb-migrate` accepts configs where ido.from > ido.to. This
silently migrates zero rows and exits 0.

`parseConfig()` now rejects from > to after parsing, before connecting.

- [x] Inverted range rejected with exit code 2
- [x] Valid range still migrates
- [x] `go build` and `go vet` pass

Closes #1157


bobapple:
@cla-bot check

oxzi (approved):
The code change is identical to the suggestion from @DanielVdWP in #1157. Should work.

